PR #4, opened 2018-04-19
Cleanup VO translation

### LINKED ISSUE #1: Cleanup VO translation
Looking here: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/CMS.yaml

Some thoughts:
1.  Remove redundant contact entries on conversion.
2. Remove the contact IDs.  We only need the IDs for cases where we can filter via the web.
3. Remove information (such as the list of resources) that can be calculated in other places.  Since the resource YAML will include the membership info, we don't also need it here too.
4. Remove defaults entries.

This should drastically simplify the VO entries.

djw8605:
- Remove redundant contact entries
- Remove ContactID's in OASIS section
- Remove resource information

Fixes #1


djw8605:
Closing, as #8 does this.

PR #17, opened 2018-05-16
Downtime info

matyasselmeci:
2 years' worth of downtime info

### DIFF
--- a/topology/ANL/ANL ASC/ANLASC_downtime.yaml
+++ b/topology/ANL/ANL ASC/ANLASC_downtime.yaml
@@ -1,0 +1,107 @@
+- Class: SCHEDULED
+  Description: power outage
+  EndTime: May 9, 2016 21:00 PM UTC
+  ID: 1004882
+  ResourceName: ANL-ATLAS-GRIDFTP1
+  Services:
+  - GridFtp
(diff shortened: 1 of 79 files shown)


topology/ANL/ANL ASC/ANLASC_downtime.yaml
```diff
@@ -0,0 +1,107 @@
+- Class: SCHEDULED
+  Description: power outage
+  EndTime: May 9, 2016 21:00 PM UTC
+  ID: 1004882
+  ResourceName: ANL-ATLAS-GRIDFTP1
+  Services:
+  - GridFtp
+  Severity: Outage
+  StartTime: May 9, 2016 11:00 AM UTC
+- Class: SCHEDULED
```
brianhlin: I forget, is whitespace allowed in between entries in a yaml list? If we added an extra newline in between entries, it'd be much more readable

matyasselmeci: I think it's allowed but I can't get the generator to add them so we can only do that for new entries.

brianhlin: That'll have to do

PR #19, opened 2018-05-16
Clarify parts of the rg template and organize services

brianhlin:
We refer to services.yaml from the template so we should clean it up a
bit for users

I'm not sure what some of these services are, i.e. FTS and LFC

### DIFF
--- a/template-resourcegroup.yaml
+++ b/template-resourcegroup.yaml
@@ -9,6 +9,9 @@ SupportCenterName: <NAME>
 # Resources contains one or more resources in this
 # ResourceGroup. A resource provides one or more services
 Resources:
+  # Resource Name should be a short descriptor of the resource.
+  # e.g. the Center for High Throughput Computing's GlideinWMS Frontend is "CHTC-glidein2"
+  # Resource Names need to be unique across all resources in the OSG.
   <RESOURCE NAME>:
     # Active is true if the resource is fully set up, and
     # false otherwise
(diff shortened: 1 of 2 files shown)


template-resourcegroup.yaml
```diff
@@ -49,7 +49,7 @@ Resources:
     #   - <HOSTNAME2>
 
     # Services is one or more services provided by this resource;
-    # valid services are listed in topology/services.yaml
+    # valid services are listed in topology/services.yaml with the format <SERVICE NAME>: <ID>
```
matyasselmeci: I don't the user needs to know this.

template-resourcegroup.yaml
```diff
@@ -90,4 +90,5 @@ Resources:
     #   TapeCapacity: 0
 
   # Other resources if you have any...
-
+  #   <SERVICE NAME>:
```
matyasselmeci: No, the "other resources" comment refers to having additional _entire resources_ e.g. everything from line 11 to here. Sounds like that needs to be clarified here -- can you suggest anything?

template-resourcegroup.yaml
brianhlin: We need to clarify so that users aren't entering `CE: 1:` for their service

template-resourcegroup.yaml
brianhlin: Oops, that's my own fault for not reading closely enough. New commit incoming

PR #20, opened 2018-05-17
- remove obsolete services on UConn-OSG site [rtj]

rjones30:


rjones30:
Brian,

Ok, that's fine. They were already marked as disabled, I believe. Maybe not
the GUMS, but it should be. Anyway, I was just looking at other reference
sites and going by what they had defined. Feel free to delete the PR.

-Richard

bbockelm:
@brianhlin - at the danger of setting policy in a github PR...

I think `git` history is more flexible than what we have in the webapp.  We should encourage / allow folks to delete removed resources as they can be easily revived and rediscovered .

brianhlin:
Since the migration occurred and this repo is now the Source of Truth, I'm merging this.

PR #22, opened 2018-05-17
Filtering and contacts

matyasselmeci:
Implements filtering for rgsummary and rgdowntime; also includes filling in contact info for rgsummary but not yet vosummary.


matyasselmeci:
Also, unless we change the argument format in the webapp (to something like `site_sel[]=University%20of%20Nebraska`), it means that future sites, facilities, VOs, services, support centers, etc. will need IDs.

app.py
```diff
@@ -75,104 +79,156 @@ def voinfo():
...
-    global _rgsummary, _rgdowntime
-    if not _rgsummary:
-        _rgsummary, _rgdowntime = get_rgsummary_rgdowntime()
 
-    rgsummary = copy.deepcopy(_rgsummary)
-    rgs = rgsummary["ResourceSummary"]["ResourceGroup"]
-    args = flask.request.args
+
+
+def get_filters_from_args(args) -> Filters:
```
brianhlin: Whoa wait, are we using Python3 for this project?

matyasselmeci: Yes.

matyasselmeci:
The semantics of the filtering are somewhat different here than in the current OIM -- there, some of the filters overrode the others, e.g. it seemed like sc (support center) overrode facility, and all_resources overrode some things but not others, and if you gave inconsistent arguments (like specifying "active" without "active_value"), it would just give a blank XML.

I just AND all the filters together and give a 400 response for inconsistent arguments. Hopefully that behavior is more logical.

app.py
```diff
@@ -75,104 +79,156 @@ def voinfo():
...
+            if v == "all":
+                filters.past_days = -1
+            elif not v:
+                filters.past_days = 0
+            else:
+                filters.past_days = int(args["downtime_attrs_showpast"])
+        except ValueError:
+            raise InvalidArgumentsError("downtime_attrs_showpast must be an integer, \"\", or \"all\"")
+
+    # 2 ways to filter by service: either pass service_1=on, service_2=on, etc.
```
brianhlin: Should this comment be more generalized? The following code doesn't seem to just apply to services

matyasselmeci: yeah, let me fix that.

PR #23, opened 2018-05-17
Some documentation for downtime

brianhlin:


template-downtime.yaml
```diff
@@ -0,0 +1,35 @@
...
+  Description: <TEXT>
+  # Downtime severity in descending order:
+  # - Outage
+  # - Severe
+  # - Intermittent Outage
+  # - No Significant Outage Expected
+  Severity: <TEXT>
+  # Start date and time of the donwtime in the following format:
+  # <MONTH> <DAY>, <YEAR> <UTC TIME>
+  # e.g. Mar 28, 2017 22:00 PM UTC
```
matyasselmeci: Can you remove the AM/PM? It's already in 24-hour time. I know the existing data contains it, but I'd like the new data to suck less.

Also, it doesn't have to be in UTC time but they may need to provide the time zone so say something like
```
# <MONTH> <DAY>, <YEAR> <TIME> <TIME ZONE>
# e.g. Mar 28, 2017 17:00 CDT
```

template-downtime.yaml
```diff
@@ -0,0 +1,35 @@
...
+  StartTime: <DATE AND TIME>
+  # End date and time of the donwtime in the following format:
+  # <MONTH> <DAY>, <YEAR> <UTC TIME>
+  # e.g. Mar 30, 2017 22:00 PM UTC
```
matyasselmeci: here too

PR #25, opened 2018-05-21
More contacts info; now serves miscuser xml file too

matyasselmeci:


converters/download
```diff
@@ -3,40 +3,50 @@
...
+            euid = os.geteuid()
+            proxypath = "/tmp/x509up_u%d" % euid
+            if euid == 0:
+                cert = '/etc/grid-security/hostcert.pem'
+                key = '/etc/grid-security/hostkey.pem'
+            elif os.path.exists(proxypath):
+                cert = key = proxypath
+            else:
+                cert = os.path.expanduser("~/.globus/usercert.pem")
+                key = os.path.expanduser("~/.globus/userkey.pem")
```
brianhlin: Can we add a TODO to respect `X509_USER_CERT` and `X509_USER_KEY` env vars?

webapp/contacts_reader.py
```diff
@@ -0,0 +1,113 @@
...
+        return self.yaml_data["ContactInformation"].get("PrimaryPhone", None)
+
+    @property
+    def sms_address(self):
+        return self.yaml_data["ContactInformation"].get("SMSAddress", None)
+
+    @property
+    def dns(self):
+        dns = self.yaml_data["ContactInformation"].get("DNs", None)
+        return dns
```
brianhlin: This is a nitpick but I was at first confused why we had contact DNS information

converters/download
matyasselmeci: That's quick enough to fix. Updated.

webapp/contacts_reader.py
matyasselmeci: Yeah, them's the breaks for language naming conventions...

PR #28, opened 2018-05-21
Put site location data into SITE.yaml files

matyasselmeci:


topology/ANL/ANL ASC/SITE.yaml
```diff
@@ -1 +1,10 @@
+AddressLine1: Building 360
+AddressLine2: ANL
+Country: USA
+Description: Argonne National Lab Analysis Support Center
 ID: 10122
+Latitude: '41.705917'
+LongName: Argonne National Lab Analysis Support Center
+Longitude: '-87.975157'
```
djw8605: It doesn't actually matter, but odd that this float is a "string".  But again, doesn't really matter.  Just as long as it's consistent.

PR #31, opened 2018-05-23
Adding Map functionality

djw8605:
Works at the same URL as before.

Test instance: https://personal-topology.herokuapp.com/map/iframe

What do people think?


matyasselmeci:
Looks great in Chrome! I can't get it work in Firefox, but maybe I just need to mess with my plugins. @brianhlin, can you try it?

djw8605:
Works for me in firefox 60.0.1 (which is the newest)

brianhlin:
Works for me but all the resources appear gray. Is that intended?

djw8605:
What do you mean "appear gray"?  screenshot?

webapp/rg_reader.py
```diff
@@ -72,8 +72,9 @@ def get_topology(indir="topology", contacts_data=None):
         topology.add_facility(name, id_)
     for site_path in root.glob("*/*/SITE.yaml"):
         facility, name = site_path.parts[-3:-1]
-        id_ = anymarkup.parse_file(site_path)["ID"]  # TODO put other site attributes in topology?
-        topology.add_site(facility, name, id_)
+        site_info = anymarkup.parse_file(site_path)
+        id_ = site_info["ID"]  # TODO put other site attributes in topology?
```
matyasselmeci: You can remove this TODO since you just did it :)

webapp/topology.py
```diff
@@ -44,13 +44,16 @@ def get_tree(self) -> OrderedDict:
...
         self.name = name
         self.id = id
         self.facility = facility
+        self.other_data = site_info
+        if "ID" in self.other_data:
+            del self.other_data["ID"]
 
     def get_tree(self) -> OrderedDict:
-        return OrderedDict([("ID", self.id), ("Name", self.name)])
+        return OrderedDict([("ID", self.id), ("Name", self.name)] + list(self.other_data.items()))
```
matyasselmeci: You should sort `list(self.other_data.items())` because currently the order is not defined.

djw8605:
Added a new root level directory, `static`.  Seems to all work, without adding anything to `app.py`.

Test away at the test URL: https://personal-topology.herokuapp.com/map/iframe

PR #33, opened 2018-05-23
Add proper auth

djw8605:


app.py
```diff
@@ -241,6 +230,40 @@ def _get_contacts_data():
     return _contacts_data
 
 
+def _get_authorized():
+    """
+    Determine if the client is authorized
+
+    returns: True if authorized, False otherwise
+    """
+    if 'GRST_CRED_AURI_0' in request.environ:
```
bbockelm: We should iterate through all the `GRST_CRED_AURI_` headers instead... I'm not sure we're guaranteed that the DN itself is in `0`.

app.py
```diff
@@ -241,6 +230,40 @@ def _get_contacts_data():
...
+        # Gridsite already made sure it matches something in the CA distribution
+
+        # HTTP unquote the DN:
+        client_dn = urllib.parse.unquote_plus(request.environ['GRST_CRED_AURI_0'])
+
+        # Get list of authorized DNs
+        authorized_dns = _get_dns()
+
+        # Authorized dns should be a set, or dict, that supports the "in"
+        if client_dn[3:] in authorized_dns: # "dn:" is at the beginning of the DN
```
bbockelm: If we authorize based on DNs, we should verify the first three characters are `dn:`.

bbockelm:
Whoops.  Meant to "request changes" instead of approve.  Can you fix the two things I highlighted?

djw8605:
@bbockelm Changes made

PR #34, opened 2018-05-23
Generalize contacts capability in osg-topology

DrDaveD:
This PR replaces the list-resource-security-contacts command with list-resource-contacts and list-vo-contacts.  They both get these two additional options:
1. --type: type of contact, for example security.  Default is to list all unique contacts for a resource or vo, regardless of type.
2. --output: output mode, either email or full.  If email, it prints only unique email addresses.

I thought this PR was ready for merging, but as I write this now I am thinking that it's not quite what the security team is going to need.  I think there needs to be a way to select both VOs and resource security contacts at the same time and to merge the list of email addresses, removing duplicates.  So maybe there should be just a list-contacts command rather than two types, and an option to select which or both types are listed.  Or maybe list-contacts should be added to the existing two.

So for now I am submitting this PR for comment rather than for merging.  I probably won't be able to get back to it until Friday.

### DIFF
--- a/bin/osg-topology
+++ b/bin/osg-topology
@@ -196,24 +196,43 @@ def list_resource_contacts(args):
                                 contacts = contact_list_info
 
                 if name and contacts and (not args.name_filter or fnmatch.fnmatch(name, args.name_filter) or args.name_filter in name):
-                    keys = results.setdefault(name, set())
-                    for contact in contacts:
-                        key = None
-                        if 'Name' in contact and 'Email' in contact:
-                            key = "%s<%s>" % (contact['Name'], contact['Email'])
-                        elif 'Name' in contact:
-                            key = contact['Name']
-                        if key:
-                            keys.add(key)
+                    results[name] = contacts
 
-    sites = results.keys()
-    sites.sort()
-    for site in sites:
-        keys = list(results[site])
-        keys.sort()
-        print("- Resource: %s" % site)
-        for key in keys:
-             print(u"  * %s" % key)
+    if args.output_mode == "email":
+        sites = results.keys()
+        emails = set()
+        for site in sites:
+            for contact in results[site]:
+                if 'Email' in contact:
+                    emails.add(contact['Email'])
+        emails = list(emails)
+        emails.sort()
+        hadone = False
+        for email in emails:
+            if hadone:
+                sys.stdout.write(' ')
+            else:
+                hadone = True
+            sys.stdout.write(email)
+        sys.stdout.write('\n')
+    else:
+        sites = results.keys()
+        sites.sort()
+        for site in sites:
+            print("- Resource: %s" % site)
+            keys = set()
+            for contact in results[site]:
+                key = None
+                if 'Name' in contact and 'Email' in contact:
+                    key = "%s <%s>" % (contact['Name'], contact['Email'])
+                elif 'Name' in contact:
+                    key = contact['Name']
+                if key:
+                    keys.add(key)
+            keys = list(keys)
+            keys.sort()
+            for key in keys:
+                 print(u"  * %s" % key)
 
 
 def get_parser():
@@ -241,6 +260,9 @@ def get_parser():
     list_resource_contacts_parser.add_argument("--type",
         dest="contact_type", help="Filter on contact type "
         "(e.g. security, default all)")
+    list_resource_contacts_parser.add_argument("--output",
+        dest="output_mode", help="Select the output mode "
+        "(email or full; default full)")
     list_resource_contacts_parser.add_argument(dest="name_filter", action="store",
         nargs="?", help="Filter on the resource name for results.")
 
--- a/bin/osg-topology
+++ b/bin/osg-topology
@@ -103,7 +103,7 @@ def mangle_url(url, args, session = None):
     qs_dict = urlparse.parse_qs(url_tuple[3])
     qs_list = urlparse.parse_qsl(url_tuple[3])
 
-    if args.provides_service:
+    if hasattr(args, 'provides_service') and args.provides_service:
         if 'service' not in qs_dict:
             qs_list.append(("service", "on"))
         for service in args.provides_service.split(","):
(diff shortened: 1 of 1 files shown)


DrDaveD:
@bbockelm please review.  There doesn't seem to be a github button for me to request a particular reviewer.

bin/osg-topology
```diff
@@ -166,48 +166,148 @@ def list_security_contacts(args):
...
+    if response.status_code != requests.codes.ok:
+        print("MyOSG request failed (status %d): %s" % (response.status_code,
+              response.text[:2048]), file=sys.stderr)
+        return 1
+
+    root = ET.fromstring(response.content)
+    if root.tag != 'VOSummary':
+        print("MyOSG returned invalid XML with root tag %s" % root.tag,
+              file=sys.stderr)
+        return 1
```
brianhlin: Can we refactor grabbing and verifying the XML into its own function? The first bit is almost exactly the same between the VO and RG functions

bin/osg-topology
```diff
@@ -166,48 +166,148 @@ def list_security_contacts(args):
...
+                            continue
+                        for contact in child:
+                            contact_info = { 'ContactType' : contact_list_type }
+                            for contact_contents in contact:
+                                contact_info[contact_contents.tag] = \
+                                    contact_contents.text
+                            contact_list_info.append(contact_info)
+
+        if name and contact_list_info:
+            results[name] = contact_list_info
```
brianhlin: Also, once we're in a VO or Resource tag, getting the contact info seems to be exactly the same. Let's refactor that, too

bin/osg-topology
```diff
@@ -114,7 +114,7 @@ def mangle_url(url, args, session = None):
                     " names: %s" % (service, ", ".join(SERVICE_IDS)))
             qs_list.append(("service_sel[]", str(service_id)))
 
-    if args.owner_vo:
+    if hasattr(args, 'owner_vo') and args.owner_vo:
```
brianhlin: Same thing about `args.get`

bin/osg-topology
```diff
@@ -166,48 +166,148 @@ def list_security_contacts(args):
...
+
+    print_contacts(args, 'Resource', results)
+
+def list_vo_contacts(args):
+    """
+    List vo contacts for OSG.
+    """
+    base_url = "https://my.opensciencegrid.org/vosummary/xml" \
+                "?summary_attrs_showcontact=on&all_vos=on" \
+                "&active_value=1&active=on&disable=on&disable_value=0"
```
brianhlin: Could you update the URL to:

```
base_url = "https://my.opensciencegrid.org/vosummary/xml" \
          "?active_value=1&active=on&disable=on&disable_value=0"
```

DrDaveD:
In what way is the listing by resources broken?

bin/osg-topology
```diff
@@ -223,17 +323,35 @@ def get_parser():
...
+    list_vo_contacts_parser.add_argument(dest="name_filter", action="store",
+        nargs="?", help="Shell expression filter on the VO name")
+
+    for parser in contactparsers:
+        parser.add_argument("--type",
+            dest="contact_type", help="Filter on contact type "
+            "(e.g. security, default all)")
+        parser.add_argument("--output",
+            dest="output_mode", help="Select the output mode "
+            "(email or full; default full)")
```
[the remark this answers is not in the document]
DrDaveD: Ok.  That slightly simplifies the --type but has no effect on --output (because it actually does full for anything other than email).

bin/osg-topology
DrDaveD: I was first going to do that but I decided it wasn't worth it because the common piece was so small.

brianhlin:
The base URL doesn't work because we turn on facilities filtering with `?facility=on`. If you drop the base URL into your browser, you'll see the 400 error

bin/osg-topology
DrDaveD: Ok, I refactored that too.

bin/osg-topology
```diff
@@ -223,17 +323,35 @@ def get_parser():
...
+        help='List OSG contacts by VO name.')
+    contactparsers.append(list_vo_contacts_parser)
+    list_vo_contacts_parser.set_defaults(which="list_vo_contacts")
+    list_vo_contacts_parser.add_argument(dest="name_filter", action="store",
+        nargs="?", help="Shell expression filter on the VO name")
+
+    for parser in contactparsers:
+        parser.add_argument("--type",
+            dest="contact_type", help="Filter on contact type "
+            "(e.g. security, default all)")
```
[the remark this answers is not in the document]
DrDaveD: There are too many types: administrative, miscellaneous, security, submitter seem the most obvious, but there's also registration authority, sponsors, VO manager, and Resource Report.  I'll put in the 4 most obvious ones.

DrDaveD:
Ok I think this is ready for merging.

bin/osg-topology
```diff
@@ -166,48 +193,140 @@ def list_security_contacts(args):
...
+    if args.output_mode == "email":
+        results = get_resource_contacts(args)
+        results.update(get_vo_contacts(args))
+        print_contacts(args, 'combined', results)
+    else:
+        list_resource_contacts(args)
+        list_vo_contacts(args)
+
+
+def print_contacts(args, type, results):
```
brianhlin: Can we rename `type` so we don't override the built-in?

DrDaveD: Done

bin/osg-topology
```diff
@@ -131,14 +131,30 @@ def mangle_url(url, args, session = None):
...
+
+    return contact_list_info
+
+
+def get_contacts(args, urltype, roottype):
+    """
+    Get one type of contacts for OSG.
+    """
+    base_url = "https://my.opensciencegrid.org/" + urltype + "summary/xml?" \
+               "&active=on&active_value=1&disable=on&disable_value=0"
```
brianhlin: It looks like we lost `gridtype=on&gridtype_1=on` for resources, waiting on an answer from @matyasselmeci to see if this is relevant.

DrDaveD: I see what those extra parameters do is to exclude entries for test resources.  Is that important?

DrDaveD: The [ticket code](https://github.com/opensciencegrid/ticket/blob/master/app/models/Address.php#L6) which sent security email does not exclude the test resources, so I don't think osg-topology should either.

matyasselmeci: AFAIK filtering by gridtype has no meaning for contacts, only for resources and resource downtimes.

brianhlin: Well we're filtering on resources then grabbing the contacts from the resultant XML. Including the testing resources seems fine to me, especially if it was the old behavior, so I'll merge this.

PR #36, opened 2018-05-24
Put the instructions on how to change the data at the top of README

DrDaveD:
Documentation is starting to refer users to this README.  Most of them will not want to modify the data themselves.  The instructions on what they need to do should be first, before the detailed info about the data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,23 +48,6 @@ as follows:
     `topology/University of Wisconsin/CHTC/CHTC.yaml`
 
 
-Accessing the Data
-------------------
-
-Data can be accessed in two formats: the YAML files can be accessed directly through the GitHub interface or from a
-clone of the GitHub repository.
-Alternatively, the data can be accessed in XML format at the following URLs:
-
-| The following data... | Can be accessed in XML format via URL...         |
-|-----------------------|--------------------------------------------------|
-| Project               | <https://my.opensciencegrid.org/miscproject/xml> |
-| Resource Downtime     | <https://my.opensciencegrid.org/rgdowntime/xml>  |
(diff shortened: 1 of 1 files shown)


DrDaveD:
I agree, I moved Modifying the Data to be between Structure of the Data and Accessing the Data.

brianhlin:
I was thinking more along the lines of keeping "Creating new data" and "Updating existing data" as sub-sections of "Modifying the data" and having that entire section below "Structure of the Data". The creating/updating data sections get a little confusing without the table of YAML file locations before them.

If you're concerned about being able to quickly find the relevant info for modifying the data, we can put a list of links at the top to each section that would act as a ToC.

brianhlin:
We need to provide some context to what information they need to provide when reaching out to help@. At the very least, they need to know where to find existing data or where new data should be created that is found in the `Modifying the data` section. I'm ok with the outline looking look like this:

```
Topology
    Modifying the Data
        Creating new data
        Updating existing data
    Structure of the Data
    Accessing the Data
```

Personally, I prefer what I said above with a table of contents:

```
Topology
    Structure of the Data
    Modifying the Data
        Creating new data
        Updating existing data
    Accessing the Data
```

I think we can trust our users to navigate around a short README, especially if we're linking them directly to specific sections from other parts of our documentation.

DrDaveD:
I don't think we can expect most grid users to pay any attention to the internal data layout.   

But whatever, this is not my responsibility, you do what you want to.  I was just trying to be helpful.

brianhlin:
If they're not paying attention to where to put files and for their content, what are they going to email us asking for? I'd rather not encourage requests where we have to play 20 questions to get the relevant data.

PR #39, opened 2018-05-25
Simplify GridType attribute in YAMLs

matyasselmeci:
Turn the GridType attribute into production:true/false since that's what it means anyway.


djw8605:
Did you update the template file as well?

-Derek

matyasselmeci:
Good call. Fixed.

Issue #30, opened 2018-05-22
Store all sources under src/ dir


brianhlin:
To improve usability for our end-users, we should move all the sources under a `src/` dir so that we have the following structure

```
projects/
src/
topology/
.gitignore
.travis.yml
virtual-organizations/
README.md
template-*.yaml
```

brianhlin:
Yes but as @bbockelm mentioned, we want to keep the option open for deployment via heroku. If we separate the source into a different repo, how does that work? @djw8605?

djw8605:
I'm not sure how that will work.  Even now, what we have won't work on heroku due to using git from the system to retrieve contact information.

Since we can't use git from the system to download the topology information, I'm not entirely sure how we would get it.

matyasselmeci:
Will Heroku check out Git submodules in the repo?

djw8605:
A quick google says no:

> Using submodules for builds on Heroku is only supported for builds triggered with git pushes. Builds created with the API do not resolve submodules. This is also the case for GitHub sync.

We use GitHub sync.  We do not `git push` directly to heroku.

brianhlin:
In the future, the contact info may go through COManage so that git dependency could very possibly go away.

matyasselmeci:
This is not my preferred solution, but it's better than nothing. Mind if I just do this?

brianhlin:
Go for it

matyasselmeci:
This has been pushed to production and works now.

PR #41, opened 2018-05-30
Move code to src/ subdir

matyasselmeci:
Fix for #30


matyasselmeci:
Tested this on my-itb.opensciencegrid.org; requires httpd config changes to WSGIDaemonProcess and WSGIScriptAlias

PR #43, opened 2018-05-31
Initial version of a new osg-notify tool.

bbockelm:
Rough draft of the new notification tool.

### DIFF
--- a/src/topology_utils.py
+++ b/src/topology_utils.py
@@ -1,0 +1,278 @@
+"""
+Various helper utilities necessary for clients of the topology
+service.
+"""
+
+from __future__ import print_function
+
+import os
+import sys
+import urllib
+import fnmatch
+import urlparse
+
+import xml.etree.ElementTree as ET
+
+import requests
+
+def get_auth_session(args):
+    """
+    Return a requests session ready for an XML query.
+    """
+    euid = os.geteuid()
+    if euid == 0:
+        cert = '/etc/grid-security/hostcert.pem'
+        key = '/etc/grid-security/hostkey.pem'
+    else:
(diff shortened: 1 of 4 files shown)


matyasselmeci:
Should have some sort of a `--dry-run` option that only prints the message to console.

matyasselmeci:
This should also be integrated with a tool that queries contacts from the osg-topology, such that we can do something like `osg-notify --type security --service ce ...` to a) send a message to security contacts for CEs and b) label it as coming from the Security Team.

bbockelm:
@matyasselmeci - now it works like this:

```
osg-notify --type test \
   --message /tmp/baz \
   --subject 'This is a test osg-notify email' \
   --dry-run \
   --oim-recipients resources \
   --oim-contact-type security \
   --from security
```

Thoughts?

src/topology_utils.py
```diff
@@ -0,0 +1,278 @@
...
+        vo_info = {}
+        for child_info in child_vo:
+            vo_info[child_info.tag] = child_info.text
+        if 'ID' in vo_info and 'Name' in vo_info:
+            vo_map[vo_info['Name'].lower()] = vo_info['ID']
+
+    return vo_map
+
+
+SERVICE_IDS = {'ce': 1,
```
matyasselmeci: You can get service IDs from `topology/services.yaml`

jteheran:
It looks good overall. 
We can start sending announcements right away!

src/topology_utils.py
bbockelm: Hm.  I'd prefer to keep this PR to refactoring.  I'll open a new issue for this idea.

PR #45, opened 2018-06-01
Auto-update contacts and topology info

matyasselmeci:
The Git repos are updated whenever data that has been stale for 15 minutes gets accessed. (This assumes that fetches are cheap.)

Tested this on topology-itb.opensciencegrid.org: set the cache lifetime to 5 min, merged #44, and made sure one of the deleted RG's was no longer in the output.

Requires `/etc/opt/topology/config-production.py`:
```python
GIT_SSH_KEY = <CENSORED>

TOPOLOGY_DATA_DIR = "/var/cache/topology/topology"
CONTACT_DATA_DIR = "/var/cache/topology/contact"
```
and `/etc/opt/topology/config-itb.py`:
```python
GIT_SSH_KEY = <CENSORED>

TOPOLOGY_DATA_DIR = "/var/cache/topology-itb/topology"
CONTACT_DATA_DIR = "/var/cache/topology-itb/contact"
```
`/var/cache/topology` and `/var/cache/topology-itb` must exist and be owned by `apache`.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -55,8 +55,12 @@ class GlobalData:
         if not self.config["NO_GIT"]:
             parent = os.path.dirname(self.config["TOPOLOGY_DATA_DIR"])
             os.makedirs(parent, mode=0o755, exist_ok=True)
-            git_clone_or_pull(self.config["TOPOLOGY_DATA_REPO"], self.config["TOPOLOGY_DATA_DIR"],
-                              self.config["TOPOLOGY_DATA_BRANCH"])
+            ok = git_clone_or_pull(self.config["TOPOLOGY_DATA_REPO"], self.config["TOPOLOGY_DATA_DIR"],
+                                   self.config["TOPOLOGY_DATA_BRANCH"])
+            if ok:
+                app.logger.debug("topology repo update ok")
+            else:
+                app.logger.warning("topology repo update failed")
         for d in [self.projects_dir, self.topology_dir, self.vos_dir]:
             if not os.path.exists(d):
                 raise FileNotFoundError(d)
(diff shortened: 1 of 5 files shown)


src/webapp/default_config.py
```diff
@@ -0,0 +1,13 @@
...
+
+TOPOLOGY_DATA_DIR = "/tmp/topology/topology"
+TOPOLOGY_DATA_REPO = "https://github.com/opensciencegrid/topology"
+TOPOLOGY_DATA_BRANCH = "master"
+
+CONTACT_DATA_DIR = "/tmp/topology/contact"
+CONTACT_DATA_REPO = "git@bitbucket.org:opensciencegrid/contact.git"
+CONTACT_DATA_BRANCH = "master"
+
+CACHE_LIFETIME = 60 * 15
```
bbockelm: Two thoughts:

1.  Can we use anywhere besides `/tmp`?  I hate named files in `/tmp`.
2.  I'm OK if we bump the cache lifetime to 30 minutes.... I'd prefer to focus on getting webhooks working smoothly.

matyasselmeci: 1. Both of these directories are supposed to get overridden in the config files on the hosts (the config files I put on the topology host write to `/var/cache/topology` and `/var/cache/topology-itb`), so there are just defaults if you don't have a config file, e.g. when testing on a local machine. I wanted something that a user could write to.

2. How about if we bump the cache lifetime once we have the webhooks implemented?

src/app.py
```diff
@@ -3,50 +3,143 @@
...
+            if not self.config["GIT_SSH_KEY"]:
+                raise RuntimeError("Contacts data requires an SSH key")
+            parent = os.path.dirname(self.config["CONTACT_DATA_DIR"])
+            os.makedirs(parent, mode=0o700, exist_ok=True)
+            ok = git_clone_or_pull(self.config["CONTACT_DATA_REPO"], self.config["CONTACT_DATA_DIR"],
+                                   self.config["CONTACT_DATA_BRANCH"], self.config["GIT_SSH_KEY"])
+            if ok:
+                app.logger.debug("contact repo update ok")
+            else:
+                app.logger.warning("contact repo update failed")
```
brianhlin: Can we get the failure message from git here?

src/webapp/common.py
```diff
@@ -148,3 +152,48 @@ def trim_space(s: str) -> str:
 
 def email_to_id(email: str) -> str:
     return hashlib.sha1(email.strip().lower().encode()).hexdigest()
+
+
+def run_git_cmd(cmd: List, dir=None, ssh_key=None) -> bool:
+    if ssh_key and not os.path.exists(ssh_key):
+        raise FileNotFoundError(ssh_key)
```
brianhlin: It doesn't look like we catch this exception anywhere so if the ssh key gets deleted accidentally, the app will stop

matyasselmeci: I've got another branch where I'm going through and changing most exceptions to log messages instead. I'll fix it in that branch (if I fix it here, there will be merge conflicts).

PR #49, opened 2018-06-04
Making the homepage a template

djw8605:
Making the homepage a template will make it easier in the future to edit.


djw8605:
I tested it on my laptop locally.  `render_template` is what we are using for the [map](https://github.com/opensciencegrid/topology/blob/59ec986559ccbdaee25d6057169d74372041053c/src/app.py#L64) as well, so it's not the first time we have used it.

PR #51, opened 2018-06-04
Don't crash on nonexistent data for downtimes

matyasselmeci:


matyasselmeci:
No, I think that should be the job of the validation scripts instead of cluttering up the server logs.
We will want to add some hooks to the reader classes so they can pass data to the validators, but that's a separate issue.

matyasselmeci:
Yep, ID is not in any of the other templates.

PR #53, opened 2018-06-04
Remove default value for "Active" in topology yamls

matyasselmeci:
Did the yaml-to-xml conversion with rg_reader.py before and after the removal; results were identical.


matyasselmeci:
Done and done.

PR #55, opened 2018-06-05
Crash less

matyasselmeci:
Log most errors instead of crashing w/ exceptions.

Also catch errors w/ the ssh key at startup time.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -34,9 +34,10 @@ class CachedData:
     def should_update(self):
         return self.force_update or (not self.data or time.time() - self.timestamp > self.cache_lifetime)
 
-    def update(self, data):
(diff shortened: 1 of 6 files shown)


matyasselmeci:
Tested on topology-itb.opensciencegrid.org

src/app.py
```diff
@@ -61,14 +77,15 @@ def _update_topology_repo(self):
                 app.logger.debug("topology repo update ok")
             else:
                 app.logger.warning("topology repo update failed")
```
brianhlin: Can we include the git output in case of failure?

matyasselmeci: `run_git_cmd` already logs it.

PR #59, opened 2018-06-08
Add detailed support center info

matyasselmeci:
The webapp also needed modification to handle the new format. Tested on topology-itb.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -150,6 +150,11 @@ app.config.from_pyfile("config.py", silent=True)
 if "TOPOLOGY_CONFIG" in os.environ:
     app.config.from_envvar("TOPOLOGY_CONFIG", silent=False)
 _verify_config(app.config)
+if "AUTH" in app.config:
+    if app.debug:
+        default_authorized = app.config["AUTH"]
+    else:
+        print("ignoring AUTH option in non-debugging mode", file=sys.stderr)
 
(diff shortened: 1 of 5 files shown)


matyasselmeci:
After merging, **immediately update** `/opt/topology` on `topology.opensciencegrid.org` and restart httpd, or things will break.

src/app.py
```diff
@@ -150,6 +150,11 @@ def get_projects(self) -> Dict:
 if "TOPOLOGY_CONFIG" in os.environ:
     app.config.from_envvar("TOPOLOGY_CONFIG", silent=False)
 _verify_config(app.config)
+if "AUTH" in app.config:
+    if app.debug:
+        default_authorized = app.config["AUTH"]
+    else:
+        print("ignoring AUTH option in non-debugging mode", file=sys.stderr)
```
brianhlin: Should we use `app.logger` here instead of `print`?

matyasselmeci: Logging hasn't been initialized at that point in the code, so that wouldn't work.

brianhlin: So does this get logged in the journal (assuming it's run as a systemd service)?

matyasselmeci: Yeah, it ends up in the httpd logs.

brianhlin:
@matyasselmeci let's get this merged and done today

PR #61, opened 2018-06-12
Add timezone to downtime timestamps

brianhlin:
Noticed by AGIS and Fermi. They also noticed that we dropped the am/pm field but that's redundant so we're not going to bring it back.


brianhlin:
Labeling this `do-not-merge` so @matyasselmeci can speak to the auto-update behavior wrt to code updates

brianhlin:
Also, we need to announce this format change to our users before making it

brianhlin:
Thinking about this further, we don't need to change the datetime format of the XML. If we want to clean up the underlying YAML that's fine but we should be resistant to change in the XML.

PR #67, opened 2018-06-15
Disable brown-cms and update contacts for brown-cms-new (helpdesk #6949)

matyasselmeci:


brianhlin:
I'm confused. The last comment on the ticket seems to imply that we're going to be using the `brown-cms` resource group but you've disabled it here.

matyasselmeci:
You're right, the ticket has moved forward since I made these commits. I'll make another PR when things become decided.

PR #78, opened 2018-06-27
Make the downtime time parsing code public & remove dateparser dependency

matyasselmeci:
Parse the date using the standard library because dateparser's behavior is not predictable enough. Prefer numeric time zones (e.g. -0500) because parsing text time zones like "CDT" is too complicated in Python. (SOFTWARE-3313)

### DIFF
--- a/src/webapp/topology.py
+++ b/src/webapp/topology.py
@@ -2,11 +2,8 @@ from collections import OrderedDict
 from datetime import datetime, timedelta, timezone
 from enum import Enum
 from logging import getLogger
-import re
 import urllib.parse
 from typing import Dict, List
-
-import dateparser
 
 from .common import MaybeOrderedDict, RGDOWNTIME_SCHEMA_URL, RGSUMMARY_SCHEMA_URL, Filters,\
     is_null, expand_attr_list_single, expand_attr_list, ensure_list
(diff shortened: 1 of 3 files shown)


template-downtime.yaml
```diff
@@ -14,11 +14,11 @@
   Severity: <TEXT>
   # Start date and time of the donwtime in the following format:
   # <MONTH> <DAY>, <YEAR> <TIME> <TIME ZONE>
-  # e.g. Mar 28, 2017 22:00 UTC
+  # e.g. Mar 28, 2017 22:00 -0500
```
edquist: Just to note here: you are changing the example timezone from UTC to (presumably) CDT.  Not sure which is more common for sysadmins, eg, based on server log timestamps.

src/webapp/topology.py
```diff
@@ -367,28 +364,39 @@ def _expand_downtime(self, service_filter=None) -> MaybeOrderedDict:
...
+
+        time = None
+        for fmt in fmts:
+            try:
+                time = datetime.strptime(time_str, fmt)
+            except ValueError:
+                pass
+            if time:
+                if not time.tzinfo:
+                    time = time.replace(tzinfo=timezone.utc)
```
edquist: Sadly `datetime.timezone` appears to be a python3 thing, or in any case it doesn't work on my rhel6 workstation.

Ah well.

matyasselmeci:
The way tzinfo is handled in Python is ridiculous and having `timezone.utc` exist saves me a lot of pain and unnecessary code. Since the target Python version is 3.6, I'm not going to do that, especially since there are other syntax differences (like the type annotations) that will prevent the function from running on Python 2 anyway. Sorry, Carl.

PR #80, opened 2018-06-27
Change downtime time format to preferred (w/ numeric time offsets)

matyasselmeci:


brianhlin:
Implementing https://opensciencegrid.atlassian.net/browse/SOFTWARE-3304 would be nice to make sure we don't bork the XML

matyasselmeci:
True. I verified by hand that the output format hasn't changed, though.

PR #85, opened 2018-06-28
Allow missing or empty downtime ID

matyasselmeci:


src/schema/rgdowntime.xsd
```diff
@@ -5,7 +5,7 @@
       <xsd:element minOccurs="0" maxOccurs="unbounded" name="Downtime">
         <xsd:complexType>
           <xsd:sequence>
-            <xsd:element name="ID" type="xsd:int" />
+            <xsd:element minOccurs="0" name="ID"/>
```
brianhlin: If ID is set, should we still make sure that it's an int?

matyasselmeci: No because it might be present but empty -- as is currently the case.

matyasselmeci: Actually maybe I should remove the minOccurs="0"

brianhlin: Arghhh. Does the `rg_reader` always set an ID?

matyasselmeci: Yep. Turns out our clients don't like missing fields, even if the fields our empty.

brianhlin: When you say "even if the fields are empty", do you mean that our consumers are ok with the field being empty?

brianhlin:
We're waiting on some answers from AGIS/SAM developers about their usage of downtime IDs

PR #86, opened 2018-06-28
script to validate xml downtimes (SOFTWARE-3304)

edquist:
First cut, still needs integration with test_verify_schema.sh script


brianhlin:
What if `topology.Downtime.parsetime` changes the format on us?

edquist:
> What if `topology.Downtime.parsetime` changes the format on us?

I thought the point was that that function would be the standard for whatever we want to accept.  Is that not the case?  Do we strictly want a fixed format for the xml?  If so i'll do that instead.

brianhlin:
Yea, we want to make sure the time format in the XML doesn't change

edquist:
Ok, it now does a `time.strptime` against a fixed `%b %d, %Y %H:%M %p %Z` time format

PR #87, opened 2018-06-29
Prompt users to verify sending email without dry-run

brianhlin:
To help avoid accidentally sending mail. We could also detect how many `@opensciencegrid.org` recipients there are to catch how many OSG mailing lists are included.

### DIFF
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -157,14 +157,18 @@ def main():
     if args.dryrun:
         print(msg)
     else:
-        while True:
-            verify_send = raw_input("Really send mail to {0} recipients? (y/N)".format(len(recipients)))
+        for _ in range(1, 4):
+            try:
+                verify_send = raw_input("Really send mail to {0} recipients? (y/N)".format(len(recipients)))
+            except EOFError:
+                verify_send = 'n'
+
             if verify_send.lower() == 'y':
                 session = smtplib.SMTP('localhost')
                 session.sendmail("help@opensciencegrid.org", recipients, msg)
                 session.quit()
                 break
-            elif verify_send.lower() == 'n' or verify_send.lower() == '':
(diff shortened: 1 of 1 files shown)


bin/osg-notify
```diff
@@ -157,10 +157,19 @@ def main():
     if args.dryrun:
         print(msg)
     else:
-        session = smtplib.SMTP('localhost')
-        session.sendmail("help@opensciencegrid.org", recipients, msg)
-        session.quit()
-
+        while True:
+            verify_send = raw_input("Really send mail to {0} recipients? (y/N)".format(len(recipients)))
```
matyasselmeci: You should catch EOFError here and treat it as a "no".

bin/osg-notify
```diff
@@ -157,10 +157,19 @@ def main():
...
+            if verify_send.lower() == 'y':
+                session = smtplib.SMTP('localhost')
+                session.sendmail("help@opensciencegrid.org", recipients, msg)
+                session.quit()
+                break
+            elif verify_send.lower() == 'n' or verify_send.lower() == '':
```
matyasselmeci: I recommend using `elif verify_send.lower() in ["n", ""]` instead to avoid the unnecessary second function call.

bin/osg-notify
```diff
@@ -157,10 +157,19 @@ def main():
     if args.dryrun:
         print(msg)
     else:
-        session = smtplib.SMTP('localhost')
-        session.sendmail("help@opensciencegrid.org", recipients, msg)
-        session.quit()
-
+        while True:
```
matyasselmeci: It would be a good idea to break after, say, 3 unrecognized answers.

brianhlin:
Overriding unrelated CI failures

PR #88, opened 2018-06-29
Add downtime ID

matyasselmeci:


brianhlin:
Just doing this so PRs stop failing as we wait on AGIS/SAM devs to let us know if the downtime ID values matter to them

PR #93, opened 2018-07-02
Add note on how to register downtime (SOFTWARE-3302)

matyasselmeci:


matyasselmeci:
Better?

README.md
```diff
@@ -75,6 +76,41 @@ methods:
 For definitions for the various fields, consult the corresponding template file for the type of data you are updating.
 
 
+### How to Register Downtime ###
+
+Downtime is a period of time for which one or more services you provide are unavailable.
+You should register downtime if one of these is true:
+
+-  your site is part of the WLCG
+-  your CE is one of the services affected
```
brianhlin: @djw8605 @bbockelm any other scenarios here? Or should we just tell users to always register downtimes?

README.md
```diff
@@ -75,6 +76,41 @@ methods:
...
+
+1. The file that contains downtime information about resources you own is named
```
brianhlin: This should start with something like "To register a downtime, identify the file..."

https://opensciencegrid.org/technology/documentation/writing-documentation/#tips-for-writing-procedural-instructions

README.md
```diff
@@ -75,6 +76,42 @@ methods:
...
+-  your site is part of the WLCG
+-  your CE is one of the services affected
+
+1. Find the file that should contain downtime information about resources you own.
+   It is named `topology/<FACILITY>/<SITE>/<RESOURCE GROUP NAME>_downtime.yaml`.
+
+   For example, `topology/University of Wisconsin/GLOW/GLOW.yaml` has the corresponding downtime file
+   `topology/University of Wisconsin/GLOW/GLOW_downtime.yaml`.
+
+   If the downtime file does not exist, you will have to create it.
```
brianhlin: For procedural instructions, we can be demanding of our users :). "Create it" instead of "you will have to create it".

README.md
```diff
@@ -75,6 +76,42 @@ methods:
...
+
+   To find out what resource group a host is in, you can grep for the FQDN of the host with something like:
```
brianhlin: Similarly here with "you can", and instead of "grep", just say "search"

README.md
```diff
@@ -75,6 +76,42 @@ methods:
...
+
+   `grep -F "<FQDN>" topology/*/*/*.yaml | grep -Fv _downtime.yaml | grep -Fv SITE.yaml`
+
+   If the grep returns nothing, then you do not have that host registered in the topology data
```
brianhlin: - "If the above command"
- "then the host is not registered in the OSG topology data"

README.md
```diff
@@ -75,6 +76,42 @@ methods:
...
+   and you don't need to register downtime for it.
+
+2. Add the contents of `template-downtime.yaml` to the end of the downtime file in the path above.
+
+3. Follow the instructions in the comments to fill out the necessary fields.
+
+   **Note:** Make sure there is no additional indentation in the YAML data you add;
+   the first line (with the `-`) must be completely unindented.
```
brianhlin: Maybe we should generalize this a bit more and say that the formatting/indentation shouldn't be changed?

matyasselmeci: I don't think they'll change formatting manually; I think unwanted indentation comes from a copy/paste error e.g. if your text editor does auto-indent.

brianhlin: Outside of the first line for each entry, there needs to be indentation, right? And just saying "the first line" may make it a little confusing if users are adding multiple entries.

matyasselmeci: I'm not sure how to word all that in order to not be confusing. Actually, Carl's script might catch it since the YAML parser will barf on mixing dict and list syntax: ```
foo: bar
narf: poit
- troz
```

matyasselmeci:
Mind if I squash this before merging?

PR #95, opened 2018-07-02
Drop MembershipServicesURL (SOFTWARE-3286)

matyasselmeci:


brianhlin:
I don't think we should be dropping *every* VOM Admin URL, just the ones that were retired. CMS and ATLAS still use the CERN VOMS Admin server for example.

matyasselmeci:
Does that mean SOFTWARE-3286 is going to remain open until we're done with SOFTWARE-3182? Maybe the tickets should be merged.

matyasselmeci:
Better?

PR #99, opened 2018-07-10
Modified pull request to add ID

a2bobball:
Added ID issued by Matyas Selmeci


matyasselmeci:
Hi Bob,
You added the ID line to the wrong downtime entry -- you should put it after the "- Class" line.
You don't need to create a new pull request, just add another commit to the branch this pull request is from.

PR #101, opened 2018-07-10
Use downtime ID of 0 if it's not present

matyasselmeci:
Nobody looks at the ID but we want to make the XML schema happy


matyasselmeci:
I created this PR against the `itb` branch so we can test it against the itb instance.

PR #106, opened 2018-07-12
Create GPGrid_downtime.yaml

tiradani:
A downtime file does not exist for GPGRID.  We are having a downtime July 20 - July 21, 2018.


brianhlin:
Hey @tiradani, this looks good but the CI failed due to missing `ID` fields. We're currently working on the best way to assign IDs so you can ignore the failures and we'll make sure this gets merged

PR #107, opened 2018-07-13
Add next_downtime_id script

matyasselmeci:
This prints the next available ID for downtime.

### DIFF
--- a/bin/next_downtime_id
+++ b/bin/next_downtime_id
@@ -3,6 +3,7 @@ set -o errexit
 set -o pipefail
 cd "$(dirname "$0")"
 datadir=$(git rev-parse --show-toplevel)/topology
-lastid=$(grep -Eh '^[[:space:]]+ID:[[:space:]]*[[:digit:]]+' "$datadir"/*/*/*_downtime.yaml | cut -d: -f2 | awk '{print $1}' | sort -n | tail -n 1)
+lastid=$( grep -Eh '^ +ID: *[0-9]+' "$datadir"/*/*/*_downtime.yaml | 
+          cut -d: -f2 | awk '{print $1}' | sort -n | tail -n 1 )
 echo $(( lastid + 1 ))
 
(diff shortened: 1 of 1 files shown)


bin/next_downtime_id
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+cd "$(dirname "$0")"
+datadir=$(git rev-parse --show-toplevel)/topology
```
edquist: Minor: `datadir` could just as well be `../topology` rather than calling `git`; or appended to the `cd` line.

bin/next_downtime_id
```diff
@@ -0,0 +1,8 @@
...
+lastid=$(grep -Eh '^[[:space:]]+ID:[[:space:]]*[[:digit:]]+' "$datadir"/*/*/*_downtime.yaml | cut -d: -f2 | awk '{print $1}' | sort -n | tail -n 1)
```
edquist: 147 characters wide in a shell script kinda grinds my gears ...

Do you really like it all on a single line?

Also, I'm a bit skeptical about the named character classes (`[:space:]` and `[:digit:]`), since they take the locale into account.  Does this mean we want to allow weird unicode space and digit characters?

Anyway if you like it how you have it, that's fine.  Otherwise I might suggest something like

```
lastid=$( grep -Eh '^ +ID: *[0-9]+' "$datadir"/*/*/*_downtime.yaml |
          awk '{print $2}' | sort -n | tail -n 1 )
```

bin/next_downtime_id
matyasselmeci: The git command needs to be run from the right dir so the cd is necessary.

bin/next_downtime_id
matyasselmeci: Yes, that looks better.

bin/next_downtime_id
edquist: No, i meant, either
```
cd "$(dirname "$0")"
datadir=../topology
```
or just
```
cd "$(dirname "$0")"/../topology
```
(and omit `$dirname` in the glob)

PR #109, opened 2018-07-14
New site and project

rynge:
This is a registration for a new project (lost in the transition from GOC) and a brand new site. I need somebody to take a pass on the site as the template has a bunch of IDs and other stuff I do not understand.


rynge:
@matyasselmeci Thanks! How do these entries look now?

rynge:
The checks got further, but now fails on the facility config.

matyasselmeci:
I made another PR against your fork with an ID for the facility.

PR #103, opened 2018-07-11
drop inactive/disabled resources (SOFTWARE-3332)

edquist:
drop resources marked as inactve or disabled, which are not referenced
in any downtime files nor in gracc records from the last year.

the list of resource names dropped here is:

```
KISTI-NSDC-CE03
KISTI-NSDC-GUMS
UVA-sunfire
uva-cms-se
OSG_LIGO_MIT
OSG_LIGO_GUMS
UARK_ACE
UTA_DPCC
UTA_SWT2_TEST1
Engagement_VOMS
RENCI-Hatteras
RENCI-Blueridge
RENCI-Blueberry
Engage-Submit3-Frontend
RENCI-Engagement
TTU-TINGE
Tusker
Firefly
Firefly_SE
Firefly-3
Firefly-2
T2_Nebraska_Hadoop
T2_Nebraska_Hadoop3
GPN-HUSKER
T2_Nebraska_Hadoop2
T2_Nebraska_Storage
US-Nebraska Disabled
T2_Nebraska_OSG
red-gateway3-ITB
VT_OSG
Hampton_ATLAS_CE
Davis_CMS
OSG_US_ASU_OCTILLO
UNI-OSGEDU
DEPRECATE MSU LAT
DEPRECATED UM LAT BW
SBGrid-Harvard-East
SBGrid-Harvard
SBGrid-Harvard-Exp
WQCG-Harvard-OSG
NYSGRID-ALBANY-ATHENA
Bern ce01
Bern ce02
UTD-HEP
UTD-HEP2
CancerComputer_Miron_CE
HMS-Orchestra-Dev
GCVO_UIS_CE
Clemson-IT
Clemson-ciTeam
Clemson-Birdnest
GCVO_PUJ_CE_TEST
GCVO_PUJ_CE
gpnjayhawk
GPN_KSU_CLUSTER
Vanderbilt_Dev_SE
Vanderbilt_Perfsonar_BW
Vanderbilt_Perfsonar_Latency
NYSGRID-CUNY-GRID
NOvA_StashCache_Server
ASGC_OSG
GCVO_UAM_CE
OCI-NSF
isuhep
T3_US_OSU
OSG_US_OSU_COWBOY
OSG_TEST_CE
GCVO_UAC_CE
GRASE-NU-CARTMAN
T3_US_FIT
Purdue-Hadoop-SE2
Purdue-Steele
Purdue-RCAC
Purdue-CMS-Test
T2_Purdue_Perfsonar_Bandwidth
[...]

PR #110, opened 2018-07-15
Updated topology for perfSONAR at UFL

ShawnMcKee:
I made two changes:
1) Added the new perfSONAR instance Bockjoo requested (IPv6 testing)
2) Fixed the topology to have one entry per perfSONAR host
Shawn


brianhlin:
@matyasselmeci are spaces allowed in resource names?

matyasselmeci:
We have at least one of those from existing data, so yes.

matyasselmeci:
Shawn, these resources are marked as inactive ("Active: false") and disabled ("Disable: true"). Is that correct, or do you want to change that?

ShawnMcKee:
Sure, I was just copying the original here but they shouldn't be marked
that way.  They should be "Active: true" and "Disable: false".

Do you need another pull request?

Thanks,

Shawn

matyasselmeci:
I don't need another pull request; you can just add a commit to this one.

brianhlin:
In the support ticket, it sounds like the new (?) perfsonar host isn't installed yet so `Active: False` may actually be what we want here. @ShawnMcKee does that sound correct? If so, then I think we just need to set `Disable: False` as well (@matyasselmeci I vaguely remember making `Disable: False` the default so we don't have to explicitly set it anymore)

matyasselmeci:
Active: true and Disable: false are the defaults. We can set Active: false for now, but we'll need to remember to 1) update it when the new hosts are ready and 2) not have the new entries get wiped in our great purge #103

ShawnMcKee:
Hi Brian,

This sounds right to me.  Bockjoo noted that the new host would be
installed "soon".

Shawn

brianhlin:
@ShawnMcKee could you remove the `Disable: True` lines and set `Active: True` for any of the hosts that are currently up? Once that's done, we can get this merged into master. Then we can have Bockjoo flip the `Active` value when the new host is up

PR #111, opened 2018-07-16
Adding July 20-21 downtime

DAMason:
Attempting to add the July 20-21 downtime for FNAL T1.  Did not do anything with the ID records in each of these -- was not clear where those numbers come from.


DAMason:
OK -- have added the new ID's

PR #112, opened 2018-07-16
Tell people to run next_downtime_id

matyasselmeci:


matyasselmeci:
Sites, facilities, RGs are easy to get the next ID for, but I can't think of a good way to get the next ID for a Resource using basic shell tools. @edquist or I might have to hack together a primitive Python YAML parser so a user can run the script without installing anything.

matyasselmeci:
Or write it in Perl and use YAML::Tiny which is a single file and can be embedded into the script.

edquist:
Discussed a bit with @matyasselmeci -

Seems "fine" (or, at least sufficient) to check for something like `^ +ID: +[0-9]{1,9} *$` - the main thing we are trying to avoid are the Contact IDs, which look like `^ +ID: [0-9a-f]{40}$`

edquist:
Or `^ +ID: +[0-9]{1,6} *$`, if @matyasselmeci really wants to make sure we never exceed 999,999 historical resources.

PR #113, opened 2018-07-17
next_id scripts (SOFTWARE-3346)

edquist:


matyasselmeci:
jeez Carl is Perl not good enough for you?

brianhlin:
Seems to me like Carl *hates* Perl.

edquist:
Fwiw, i did in fact try to write one of these in perl, to be a good sport at least, but fact is it's just way more verbose (and, arguably, ugly) than the line-and-a-half of grep+awk

PR #114, opened 2018-07-18
Fix sign error for past days downtime filter

brianhlin:
Before this change, specifying any past_days >= 0 resulted in no past
downtimes ever being displayed and only future downtimes whose
end times were more than past_days in the future.

Before: https://my.opensciencegrid.org/rgdowntime/xml?downtime_attrs_showpast=39
After: https://my-itb.opensciencegrid.org/rgdowntime/xml?downtime_attrs_showpast=39

### DIFF
--- a/src/webapp/topology.py
+++ b/src/webapp/topology.py
@@ -303,6 +303,10 @@ class Downtime(object):
 
     @property
     def end_age(self) -> timedelta:
+        """Returns a timedelta instance for the end time of a downtime relative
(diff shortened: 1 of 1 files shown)


edquist:
I waited 2 seconds too long to click "Request changes", but we should really fix `end_age` instead, to return `current_time - self.end_time`.  The "age" of `end_time` should be positive if `end_time` is in the past and (i guess) negative if `end_time` is in the future.

PR #115, opened 2018-07-18
Time goes forward

edquist:


brianhlin:
I've got a similar commit on the topology-itb and it doesn't seem to be working for whatever reason:
https://my-itb.opensciencegrid.org/rgdowntime/xml?downtime_attrs_showpast=39

src/webapp/topology.py
```diff
@@ -303,12 +303,9 @@ def timeframe(self) -> Timeframe:
 
     @property
     def end_age(self) -> timedelta:
-        """Returns a timedelta instance for the end time of a downtime relative
-        to the current time. Past end times yield a negative timedelta and
-        future end times yield a positive timedelta.
-        """
+        """Return timedelta elapsed since end_time"""
```
brianhlin: Could we convey that future end times will result in a "negative" timedelta?

src/webapp/topology.py
```diff
@@ -324,9 +321,8 @@ def get_tree(self, filters: Filters = None) -> MaybeOrderedDict:
             return
         # unlike the other filters, if past_days is not specified, _no_ past downtime is shown
         if filters.past_days >= 0:
-            # Filter out all downtimes that are older than 'past_days' so we swap its sign to account for end_age being
-            # negative for past end times. Always show downtimes with end_ages in the future.
-            if self.end_age.total_seconds() // 86400 < -filters.past_days:
+            # Filter out downtimes older than 'past_days'
+            if self.end_age.total_seconds() // 86400 > filters.past_days:
```
brianhlin: I think it's worth keeping the comment about showing downtimes with `end_age` in the future to remind ourselves that this block doesn't just affect past downtimes.

matyasselmeci: The block _only_ affects past downtimes. All present and future downtimes are shown.

brianhlin: Ok, nevermind then!

edquist:
I've added two more commits:

- change the behavior not to truncate `end_age` - previously `past_days = 1` wouldn't filter anything until 2 days, ie 48 hours, but up to 47 hours was not filtered.  New behavior is 24h + 1s gets filtered.

- added comments based on @brianhlin's feedback.  In principle i feel like the language "Filter out downtimes older than 'past_days'" itself suggests that future downtimes are not subject to this filter, and (i agree with @matyasselmeci), the block _does_ just effect past downtimes.  However, if I understand Brian's intention, he wanted a comment "to remind ourselves" that future downtimes are a thing too, and that they do not get filtered out.  So whatever, i'm fine with that too.

PR #117, opened 2018-07-19
Update EndTime of permanent outages

matyasselmeci:


brianhlin:
I'd prefer removing them outright, let's see what Alessandro says

brianhlin:
@edquist can we handle the exceptions mat's seeing to get a more helpful failure message?

matyasselmeci:
@edquist , this was due to an empty downtime file. You should check if the parsed YAML data is None and if so, add an error saying "$filename was not parseable" or something.

edquist:
I've made #134 to make better errors in the future for empty yaml files.

PR #132, opened 2018-07-27
Drop resources allocated to disabled VOs (SOFTWARE-3335)

edquist:


edquist:
Only interesting bit to mention here is the `USP-LCCA` resources all had `OSGEDU: 80` (as opposed to 100), but it was still the only VO listed in those cases.

brianhlin:
@edquist 
1. Have you reached out to the admins of USP-LCCA?
1. Have you verified that the resources you've removed have not had GRACC records in the past year?
1. Can we remove [the ignored list of VOs](https://github.com/opensciencegrid/topology/blob/master/src/tests/verify_resources.py#L34-L36)?

edquist:
> 1. Have you reached out to the admins of USP-LCCA?

Looks like i did not.  I've just emailed them today.

> 2. Have you verified that the resources you've removed have not had GRACC records in the past year?

Yes, I don't see any GRACC records from the last year referencing USP-LCCA.

> 3. Can we remove [the ignored list of VOs](https://github.com/opensciencegrid/topology/blob/master/src/tests/verify_resources.py#L34-L36)?

After the current changes for this PR, no more resources refer to any VOs in that ignored list in their VOOwnership; so for the purposes of the verify_resources script, yes.

brianhlin:
> Yes, I don't see any GRACC records from the last year referencing USP-LCCA.

What about the other resources?

edquist:
Ah, right - yeah those don't appear in my GRACC list for the last year either.

brianhlin:
Alright, sort out the conflicts (@matyasselmeci  re-added `Active: true` to the resources without it) and remove the ignored VO list from `verify_resources.py`

PR #131, opened 2018-07-27
Add ACCRE VO

PerilousApricot:
Add ACCRE as a VO. ACCRE is Vanderbilt's research computing center

PR #133, opened 2018-07-27
Clarify OASISRepoURLs field

brianhlin:
https://github.com/opensciencegrid/topology/pull/131#issuecomment-408492682 @DrDaveD


template-virtual-organization.yaml
```diff
@@ -81,7 +81,7 @@ SupportURL: http://www.usatlas.bnl.gov/index.shtml
 #   #     - ...
 #   #   ...
 #
-####### OASISRepoURLs (optional) are one or more URLs for OASIS content owned by this VO
+####### OASISRepoURLs (optional) are one or more URLs for OASIS repositories owned by this VO (https://opensciencegrid.org/docs/data/external-oasis-repos/)
```
DrDaveD: I'm not sure it's good to have such a long line.  I don't see precedent for it, and I some editors (such as vim) do not by default split lines on word boundaries.

matyasselmeci: Fair enough. We tend to use longer lines in our code, but we should be stricter in our line lengths for user-editable data.

brianhlin: PRs encouraged :). Splitting the URL onto a separate line is fine with me

PR #134, opened 2018-07-27
give error messages for empty yaml files

edquist:
As requested in #117

SOFTWARE-3313 related.


src/tests/verify_downtimes.py
```diff
@@ -22,6 +22,13 @@ def validate_downtime_file(dt_fname):
     rg_yaml = yaml.safe_load(open(rg_fname))
 
     errors = []
+    if dt_yaml is None:
+        errors.append("Downtime file is empty: %s" % dt_fname)
```
matyasselmeci: It might not be empty, just unparseable

PR #142, opened 2018-08-01
decomissioning Purdue-Conte

marabgol:
my first commit, I am not sure if it works with my gmail 
Majid


brianhlin:
@marabgol hi, due to some intricacies of GRACC accounting, we cannot disable resources (which is the same as deletion) that have uploaded Gratia records in the [past year](https://gracc.opensciencegrid.org/dashboard/db/site-summary?orgId=1&from=now-1y&to=now&var-interval=$__auto_interval&var-site=All&var-type=Batch&var-Filter=OIM_Resource%7C%3D%7CPurdue-Conte). The proper way to do this for these types of resources is by setting `Active: false` in the resource. We hope to document this and improve the templates tomorrow through  [this JIRA ticket](https://opensciencegrid.atlassian.net/browse/SOFTWARE-3371).

We appreciate the pull request but I'll close this and separately deactivate your resource!

PR #143, opened 2018-08-01
Set old SL6 CE as Disabled.  Update total storage in the SE

a2bobball:
Disable AGLT2_SL6, and update our total storage size.

### DIFF
--- a/topology/University of Michigan/AGLT2/AGLT2.yaml
+++ b/topology/University of Michigan/AGLT2/AGLT2.yaml
@@ -174,7 +174,7 @@ Resources:
           Name: Bob Ball
     Description: New SL6 CE at AGLT2
     Active: false
-    Disable: true
+    Disable: false
     FQDN: gate04.aglt2.org
     ID: 547
     Services:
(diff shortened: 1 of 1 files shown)


brianhlin:
@a2bobball Could you revert the change to the `Disable` field? It's the same as deleting a resource and we can't remove resources that have GRACC records in the past year.

To indicate that a resource is retired, we set `Active: false`, which I've done for AGLT2_SL6 already.

a2bobball:
Note that the Active switch is not documented in the template-resourcegroup.yaml file.

brianhlin:
There are some other resources with `Active: false` at other sites. We'll fixup the template file, thanks for the heads up!

brianhlin:
Also, we hope to address this hole in our documentation tomorrow. You can track progress via this [JIRA ticket](https://opensciencegrid.atlassian.net/browse/SOFTWARE-3371) or expect to see detailed documentation by sometime next week [here](https://opensciencegrid.org/docs/common/registration/)

PR #149, opened 2018-08-01
Add IDs to templates

matyasselmeci:


brianhlin:
This is ok for now but from our discussion yesterday, I think that all of these are candidates for ID-less entries. Where can we find the topo web app access logs to see what kind of queries that people are making to see if they're filtering based on these IDs?

brianhlin:
> And if so, their scripts may not be able to handle non-numeric values.

Since the plan is to eliminate the ID field for future registered entries the issue may be the ability to handle entries without an ID field entirely.

Since we enforce site/facility/VO name uniqueness and will be enforcing resource FQDN uniqueness, we just need to add the ability to filter based on the above attributes. Then we can ask our consumers to update their scripts to filter that way instead.

matyasselmeci:
Sure. We can remove the ID fields once we are sure that our consumers have updated their scripts and no longer need them.

PR #150, opened 2018-08-01
Log errors instead of crashing when adding bad RGs to topology

matyasselmeci:


matyasselmeci:
I tested this by commenting out SupportCenter and GroupID in one of the RG YAML files and running `rg_reader.py`. Note that the CI catches a missing SupportCenter but does not currently catch a missing GroupID -- that's part of [SOFTWARE-3287](https://opensciencegrid.atlassian.net/browse/SOFTWARE-3287), assigned to @edquist.

brianhlin:
So in the CI, does your message get printed to stdout?

matyasselmeci:
stderr I believe

PR #151, opened 2018-08-02
Remove extraneous contacts

brianhlin:


matyasselmeci:
Are you sure we don't care about Resource Report Contact? I didn't mark it as optional when making the template...

PR #152, opened 2018-08-02
Fix location of email-hash script

matyasselmeci:


matyasselmeci:
Yeah, the VO template also needs to be updated.

matyasselmeci:
How does this look?

PR #155, opened 2018-08-06
updating ASU's profile, adding an ASU VO, removing Jay etching, adding the new cluster

562uned:
updating ASU's profile, adding an ASU VO, removing Jay etching, adding the new cluster


matyasselmeci:
I updated the data in PR #165 so I'm going to close this.

PR #166, opened 2018-08-15
Downtime web form

matyasselmeci:
This is a web form for generating YAML blocks to be pasted into a downtime YAML file. It's running on topology-itb: https://topology-itb.opensciencegrid.org/

PR #167, opened 2018-08-15
Fix cache bug that may load broken data & do some reorganizing

matyasselmeci:
This fixes a bug in the webapp where data is loaded from the disk even if the underlying git update fails.

There's also some reorganization because app.py was getting large.


matyasselmeci:
I tried this out on topology-itb and it's also part of #166.

PR #168, opened 2018-08-16
Retire Lehigh_coral

brianhlin:
- Hosts don't appear to have DNS records anymore
- Contacts are listed as "alumni": http://coral.ise.lehigh.edu/people/
- Ashutosh's emails bounce


matyasselmeci:
These are marked as active. Have you checked that they haven't reported into GRACC?

brianhlin:
Just checked, couldn't find any sign of 'em

PR #170, opened 2018-08-17
Add test downtime generated by webform

matyasselmeci:
This is a piece of downtime generated by the webform to test if our consumers have problems... consuming the data.


matyasselmeci:
wow i'm dumb

PR #172, opened 2018-08-21
Added aragon.cyverse.org submit node

rynge:
This is a new submit host flocking to flock.opensciencegrid.org. Before accepting this PR, we probably need to setup a facility definition, but I don't know how to do that.


matyasselmeci:
Hi Mats,

Some things changed in our template resourcegroup so I made a few fixups to the Aragon.yaml file and also added a FACILITY.yaml and a SITE.yaml file with the IDs filled in, and submitted them as a PR against your repo. Could you edit the new Cyverse/Cyverse/SITE.yaml file and fill in the info?

Thanks,
-Mat

rynge:
I have updated SITE.yaml. It is weird that it wants what looks like a full address, but not the city.

matyasselmeci:
My bad, I left that out of the template. Could you add the city as well?
```
City: <name of city>
```

rynge:
City added

matyasselmeci:
Thanks! I made one more PR to add the GroupID and ID fields to the resource and resource group.

PR #176, opened 2018-08-23
add network checks for osg-notify emails (SOFTWARE-3389)

edquist:
The `bin/network-hostname-ip-check` symlink can be used as a stand-alone script to check the network preconditions for sending an email.

The conditions are:
 - the fqdn of the hostname maps to an IPv4 address that is not in a local or private range
 - the fqdn of the reverse lookup for that IP address matches the original hostname fqdn

The `src/net_name_addr_utils.py` script can also be used as a library; and `osg-notify` now imports it to use these network function checks.

### DIFF
--- a/bin/network-hostname-ip-check
+++ b/bin/network-hostname-ip-check
@@ -1,0 +1,1 @@
+../src/net_name_addr_utils.py
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -16,6 +16,7 @@ import email.mime.multipart
 import gnupg
 
 import topology_utils
+import net_name_addr_utils
 
 # Parts of this implementation are from the following StackOverflow answer:
 # https://stackoverflow.com/questions/10496902/pgp-signing-multipart-e-mails-with-python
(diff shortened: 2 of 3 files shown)


edquist:
I have tested this and it works as expected for
(1) a correctly configured network (does not complain)
(2) a host with hostname set to `localhost.localdomain` (fails as expected for non-public IP)
but not for
(3) a host that lies about its hostname (eg, `fermicloud024.fnal.gov` instead of `fermicloud023.fnal.gov`)

I believe this actually means that the logic in htcondor-ce's [condor_ce_host_network_check](https://github.com/opensciencegrid/htcondor-ce/blob/master/src/condor_ce_host_network_check) script has the same flaw, as I am computing the reverse hostname the same way.

See this sequence: [1](https://github.com/opensciencegrid/htcondor-ce/blob/v3.1.3/src/condor_ce_host_network_check#L369), [2](https://github.com/opensciencegrid/htcondor-ce/blob/v3.1.3/src/condor_ce_host_network_check#L386), [3](https://github.com/opensciencegrid/htcondor-ce/blob/v3.1.3/src/condor_ce_host_network_check#L420), [4](https://github.com/opensciencegrid/htcondor-ce/blob/v3.1.3/src/condor_ce_host_network_check#L432-L433), [5](https://github.com/opensciencegrid/htcondor-ce/blob/v3.1.3/src/condor_ce_host_network_check#L442-L445)

Basically, the problem is, if a host lies and says that its hostname is `wisc.edu` which resolves to `144.92.9.70`, and that IP reverse-resolves back to `wisc.edu`, the check passes.  Maybe this is OK if we don't care about hosts that lie about their hostnames.  But i suspect if we want to catch this, we will need to contact an external service (or one that we run) that will report our external public IP, and verify that this external IP matches the one that the hostname resolves to.  (This check is in addition to the current check, that hostname matches reverse hostname.)

PR #179, opened 2018-08-29
Downtime webform improvements

matyasselmeci:
These implement most of the wishlist items from our discussion except for the textbox with autocomplete for selecting the facility, which looks like it requires too much code to implement.

This is currently running on https://topology-itb.opensciencegrid.org/generate_downtime.

(This is a PR against the `itb` branch.)


brianhlin:
What's the point of the change facility/resource buttons? It looks like if I pick a new facility/resource in the dropdown, it automatically unselects services and resources as appropriate.

matyasselmeci:
That's in case people disable javascript.

brianhlin:
I tried using the form while disabling js and boy you can really get yourself into an annoying situation:

1. Select a facility, resource, and service
1. Change your mind and try to change the facility, you can't because all the other fields complain about being required

I vote that we don't bother supporting the no-js case. Anyone using a js-blocker will be well-accustomed with unblocking js to get a functional web experience.

matyasselmeci:
OK, how about now?

PR #180, opened 2018-08-31
CHTC glidein2 upcoming downtime

tjslauson:
I'm taking down glidein2, and there's nothing you can do to stop me! >:D


brianhlin:
Thanks for the form review!

PR #186, opened 2018-09-05
Downtime webform updates

matyasselmeci:
These address some of the concerns from our run-through.

### DIFF
--- a/src/templates/generate_downtime_form.html.j2
+++ b/src/templates/generate_downtime_form.html.j2
@@ -102,6 +102,10 @@
                 <pre>"{{ filename }}"</pre>
               </li>
             {% endif -%}
+            <li>
+              If you don't have a fork of the topology repo,
+              click the "Fork" button to create one.
+            </li>
           {% else -%}{# github #}
             <li>
               In your topology repo checkout, open the file named:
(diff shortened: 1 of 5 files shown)


src/templates/generate_downtime_form.html.j2
```diff
@@ -99,14 +102,32 @@
...
+            {%- endif %}
+          </li>
+          <li>
+            {% if github -%}
+              Specify a commit message in the "Commit changes" section.
+            {%- else -%}
+              Commit your changes to the repository.
+            {%- endif %}
+            The commit message should look something like<br>
+            <code>Add downtime for &lt;RESOURCE(S)&gt; due to &lt;REASON&gt;</code>
```
brianhlin: This form can only generate a single downtime so this should read `RESOURCE`

brianhlin:
@matyasselmeci are you going to address resource FQDNs in a separate PR?

matyasselmeci:
Ah, I _knew_ I was missing something. Nah, I'll do it in this one.

src/webapp/forms.py
```diff
@@ -29,7 +29,7 @@ class GenerateDowntimeForm(FlaskForm):
     start_time = TimeField("&nbsp;", [InputRequired()])
     end_date = DateField("End Date/Time (UTC)", [InputRequired()])
     end_time = TimeField("&nbsp;", [InputRequired()])
-    services = SelectMultipleField("Services (select one or more)", [InputRequired()], choices=[])
+    services = SelectMultipleField("Known OSG Services (select one or more)", [InputRequired()], choices=[])
```
brianhlin: Finally found a resource with multiple services, maybe we should change this to `(select one or more with ctrl-click)`?

matyasselmeci: I would've put that but it's not ctrl-click on macs.

PR #188, opened 2018-09-07
Add downtime for CHTC-glidein2 due to physical host migration

aaronmoate:


matyasselmeci:
Thanks for helping me test!

PR #190, opened 2018-09-07
Freeze requirements at their current minor versions

matyasselmeci:
Except for patched versions, upgrades should be explicit.


bbockelm:
Seems reasonable, but can you fixup the merge conflict that has developed?

matyasselmeci:
Done. Anything else?

PR #193, opened 2018-09-10
Update contacts and machine description

duncan-brown:
These nodes are no longer used by the LIGO Scientific Collaboration for OSG submission, so I have updated the description to remove mention of LIGO and removed Peter Couvares from the contacts.


matyasselmeci:
I made the fixup in another PR and merged them so this is done now.

PR #195, opened 2018-09-11
Removing Compute Element until activated

ahkumar:
Removed CE as it has been decommissioned until new server is setup. 
Also removed SRM resource as we transitioned to Gridftp only Storage Element.

### DIFF
--- a/topology/Southern Methodist University/SMU_HPC/SMU_HPC.yaml
+++ b/topology/Southern Methodist University/SMU_HPC/SMU_HPC.yaml
@@ -3,53 +3,6 @@ GroupDescription: Southern Methodist University HPC
 GroupID: 242
 Production: true
 Resources:
-  SMU_ManeFrame_CE:
-    Active: true
-    ContactLists:
-      Administrative Contact:
-        Primary:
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Hi Amit,

Could you also remove the downtime info associated with that CE? It's in the file `topology/Southern Methodist University/SMU_HPC/SMU_HPC_downtime.yaml`. Just remove all the entries where `ResourceName` is `SMU_ManeFrame_CE`.

ahkumar:
Just created a new pull request after updating downtime info, please let me know if I did it right?

matyasselmeci:
Hi Amit,
It would have been easier if you had put the new commit in this PR. Unfortunately, you deleted a little too much -- the files are in YAML format so the record actually begins when the line starts with a `-` character. I merged them after a bit of fixing up.

I'll try to come up with some way of making the record boundaries more obvious; you were not the only person that ran into this problem.

ahkumar:
Hi Mat, sorry about that I should have been careful as well. Please let me know if there is anything else I need to do from my end? Also where would I check that my resource is active on the OSG site?

PR #203, opened 2018-09-12
Add downtime webform

matyasselmeci:
This is a web form for generating YAML files that an admin can copy-and-paste into a downtime file to report downtime. It is running on topology-itb at https://topology-itb.opensciencegrid.org/generate_downtime.


brianhlin:
Marking this as `do-not-merge` because we will tell our consumers the day we plan on making this change and should not merge this until then.

matyasselmeci:
Dang, that was fast.

In the meantime, feel free to give it to someone else for review since you probably don't have the time.

src/templates/homepage.html.j2
```diff
@@ -0,0 +1,29 @@
+{% extends "base.html.j2" %}
+{% block title -%}
+OSG Topology Interface
+{%- endblock %}
+{% block content %}
+<div class=container>
```
djw8605: `container` should be in quotes.

matyasselmeci:
I don't know enough JavaScript to do that properly, and what I have now took long enough anyway. WTForms does a good job of validating and providing useful feedback already.

PR #204, opened 2018-09-13
Unscheduled CIT_CMS_T2 downtime for next 24h

juztas:
Networking issues with routes and also few links are dropping packets. Networking team will follow up on this tomorrow morning, so I announce this downtime for the next 24h. I will keep posted when it is done


juztas:
thanks @bbockelm

PR #207, opened 2018-09-13
Add NMSU hosted ce topology info

sthapa:
Can you double check the facility and group ids?  I believe they're correct but am not sure.


brianhlin:
Butting in here: don't we need a `SITE.yaml`? Should we have a CI test for that?

matyasselmeci:
Good catch! Yes, we do. And we should have a CI test for that too.

matyasselmeci:
Suchandra, could you create a `topology/New Mexico State University/High Performance Computing/SITE.yaml` file? There's a `template-SITE.yaml` file in the top of the repo.

PR #212, opened 2018-09-18
Add site yaml for NMSU

sthapa:


topology/New Mexico State University/High Performance Computing/SITE.yaml
```diff
@@ -0,0 +1,15 @@
+LongName: New Mexico State University Discover Cluster
+Description: Discovery cluster at NMSU
+
+ID: 12346
```
matyasselmeci: ...apparently putting 12345 in the template was dumb of me because it confused next_site_id. I'll fix the template; in the meantime, could you use 10239 instead?

sthapa: Updated

PR #213, opened 2018-09-18
Don't put a number in template-SITE.yaml because it confuses next_site_id

matyasselmeci:
If someone copies template-SITE.yaml into a site's SITE.yaml and runs next_site_id, next_site_id will see "12345" and go "that's the largest, next one must be 12346". Let's just put a placeholder in the template.


brianhlin:
Did you check the ID fields in the other templates?

matyasselmeci:
Yes.

PR #216, opened 2018-09-20
Human-readable contacts page

matyasselmeci:
This is basically a prettier, HTML version of the contacts XML data. No JavaScript in there but later on we may want to add some searches or filtering since right now people will have to use their browser's search.

This is running on https://topology-itb.opensciencegrid.org/contacts

### DIFF
--- a/src/templates/contacts.html.j2
+++ b/src/templates/contacts.html.j2
@@ -3,13 +3,15 @@
 OSG Contacts
 {%- endblock %}
 {% block content %}
-{% if not authorized %}
-<div class="alert alert-info" role="alert">
-  To see contact information, load your grid certificate into your browser.
-</div>
-{% endif %}
-<div class="alert alert-info" role="alert">
-  Profile information is provided by the user and may be out of date.
+<div class="container">
+  {% if not authorized %}
+  <div class="alert alert-info" role="alert">
+    To see contact information, load your grid certificate into your browser.
+  </div>
+  {% endif %}
+  <div class="alert alert-info" role="alert">
+    Profile information is provided by the user and may be out of date.
+  </div>
 </div>
 <div class="container">
 {%- for user in users %}
--- a/src/templates/contacts.html.j2
+++ b/src/templates/contacts.html.j2
@@ -13,7 +13,8 @@ OSG Contacts
 </div>
 <div class="container">
   {%- for user in users %}
-    <div class="row text-white bg-dark">
+  <div class="container">
+    <div class="row text-white bg-dark" style="margin-top:1em;">
       <div class="col-sm"><span class="font-weight-bold">{{ user.FullName }}</span></div>
       <div class="col-sm-auto text-right">
         <span class="font-weight-bold">ID:</span>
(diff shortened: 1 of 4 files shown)


src/templates/contacts.html.j2
```diff
@@ -0,0 +1,64 @@
...
+      <p class="col-sm-8">
+        {% if user.Profile -%}
+        {{ user.Profile|replace("\n", "<br>") }}
+        <br>
+        {%- else -%}
+        <span class="text-danger">(no profile)</span>
+        {%- endif %}
+      </p>
+    </div>
+    <div class="row">
```
djw8605: This `row` should probably go inside the if.

src/templates/contacts.html.j2
```diff
@@ -0,0 +1,64 @@
...
+      <div class="row">
+        <div class="col-sm-3"><span class="font-weight-bold">{{ lbl }}:</span></div>
+        <div class="col-sm">{{ user.ContactInformation[attr] }}</div>
+      </div>
+      {% endif %}
+      {% endfor %}
+    {%- endif %}
+  {% endfor -%}
+</div>
+{% endblock %}
```
brianhlin: No newline??

src/templates/contacts.html.j2
```diff
@@ -0,0 +1,64 @@
...
+<div class="alert alert-info" role="alert">
+  To see contact information, load your grid certificate into your browser.
+</div>
+{% endif %}
+<div class="alert alert-info" role="alert">
+  Profile information is provided by the user and may be out of date.
+</div>
+<div class="container">
+  {%- for user in users %}
+    <div class="row text-white bg-dark">
```
djw8605: I think there needs to be some separation between the users.  If you don't want to add a style sheet, just add the style attribute to this div, so it would look like:

    <div class="row text-white bg-dark" style="margin-top:1em;">

src/templates/contacts.html.j2
matyasselmeci: Geez you'd really think Git would have told me about that.

src/templates/contacts.html.j2
```diff
@@ -0,0 +1,64 @@
...
+{% if not authorized %}
+<div class="alert alert-info" role="alert">
+  To see contact information, load your grid certificate into your browser.
+</div>
+{% endif %}
+<div class="alert alert-info" role="alert">
+  Profile information is provided by the user and may be out of date.
+</div>
+<div class="container">
+  {%- for user in users %}
```
djw8605: I would prefer each user be in their own "div".  Then, you can do things like [bootstrap cards](https://getbootstrap.com/docs/4.1/components/card/) (kinda like contact cards)

brianhlin:
After this is merged:

- Follow up on [this ticket](https://support.opensciencegrid.org/a/tickets/7588) and link them to the new contact page so they know how to find user IDs in the future
- Replace the XML link in the [registration doc](https://opensciencegrid.org/docs/common/registration/) with this new page

src/templates/contacts.html.j2
```diff
@@ -0,0 +1,64 @@
...
+{% block title -%}
+OSG Contacts
+{%- endblock %}
+{% block content %}
+{% if not authorized %}
+<div class="alert alert-info" role="alert">
+  To see contact information, load your grid certificate into your browser.
+</div>
+{% endif %}
+<div class="alert alert-info" role="alert">
```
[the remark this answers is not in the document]
matyasselmeci: Both the alerts or just the second one? And if both, should both be in the same container, or separate?

matyasselmeci:
@djw8605 I made the changes. I messed around with cards for a couple of minutes but I couldn't get them to look nice; it kept putting the name and the ID on separate lines.

matyasselmeci:
Never mind, got cards working. Cards (left) vs non-cards (right); which do you prefer?
card vs nocard

brianhlin:
I vote cards

djw8605:
Cards, left!

src/templates/contacts.html.j2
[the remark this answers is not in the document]
djw8605: Both alerts.

matyasselmeci:
Cards it is. Anything else?

djw8605:
Looks good to me!

PR #217, opened 2018-09-20
Display placeholder for missing GH account & add who to contact for fixing info

matyasselmeci:
Hopefully the "GitHub: (none)" in people's contact info will serve as a hint to get people to send us their info?

### DIFF
--- a/src/templates/contacts.html.j2
+++ b/src/templates/contacts.html.j2
@@ -34,7 +34,7 @@ OSG Contacts
           {{ user.Profile|replace("\n", "<br>") }}
           <br>
           {%- else -%}
-          <span class="text-danger">(no profile)</span>
+          <span class="text-danger" data-toggle="tooltip" title="Mail help@opensciencegrid.org to update">(no profile)</span>
           {%- endif %}
         </p>
       </div>
--- a/src/templates/contacts.html.j2
+++ b/src/templates/contacts.html.j2
@@ -1,3 +1,4 @@
+{% set want_js = true %}
 {% extends "base.html.j2" %}
 {% block title -%}
 OSG Contacts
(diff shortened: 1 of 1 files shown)


src/templates/contacts.html.j2
```diff
@@ -36,12 +37,16 @@ OSG Contacts
...
       </div>
-      {% if user.GitHub -%}
       <div class="row">
         <div class="col-sm-3"><span class="font-weight-bold">GitHub:</span></div>
-        <div class="col-sm"><a href="https://github.com/{{user.GitHub}}">{{ user.GitHub }}</a></div>
+        <div class="col-sm">
+          {% if user.GitHub -%}
+          <a href="https://github.com/{{user.GitHub}}">{{ user.GitHub }}</a>
+          {%- else -%}
+          <span class="text-danger">(none)</span>
```
djw8605: Maybe make it explicit.  Like have the `(none)` text be:

    (none) Update github account name here

Or something more clever?  Bonus points if you make it a mailto:help@opensciencegrid.org with subject and some body already filled out.

matyasselmeci: That might be a little too glaring; we don't have GitHub info for 99% of our contacts.

matyasselmeci:
Sure, that works. I couldn't add a link to the tooltip though.

bbockelm:
I can confirm that if I'm not logged in, I don't see email addresses (but I can see them if I'm logged in).

How impossible is it to filter on people associated with an active resource, VO, or project?

matyasselmeci:
Gee, doing that sort of correlation would be a lot easier if the data was in a SQL database or something :stuck_out_tongue:
It'd probably take me a day or two overall; I'll have to modify the data structures to keep track of contacts "seen." Hopefully it won't add too much to the processing time.

djw8605:
I like it.  I say merge this now, and worry about filtering later.

matyasselmeci:
Yeah, that should be a separate pull request anyway.

PR #226, opened 2018-09-28
Update SWT2_CPB_downtime.yaml

ptmcguigan:
Tried this last night, but I think I was not modify the correct trunk.  Trying again.

### DIFF
--- a/topology/University of Texas Arlington/SWT2 ATLAS UTA/SWT2_CPB_TEST_downtime.yaml
+++ b/topology/University of Texas Arlington/SWT2 ATLAS UTA/SWT2_CPB_TEST_downtime.yaml
@@ -75,3 +75,27 @@
   - net.perfSONAR.Latency
 
 # ----------------------------------------------------------
+- Class: UNSCHEDULED
+  ID: 31883589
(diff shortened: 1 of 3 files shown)


bbockelm:
Hi @ptmcguigan - I just corrected a small formatting issue.

To double-check: this is listed as being at 0700 GMT (i.e., 0100 central).  Is that correct?

ptmcguigan:
HI,

Actually it is at 2am Sunday morning, which should be 07:00 GMT.

Patrick

bbockelm:
Looks like a few downtimes are in the wrong files:

- Downtime for UTA_SWT2_perfSONAR_latency is in SWT2_CPB but should be in UTA_SWT2
- perfSONAR_SWT2_CPB_testbed_bandwidth is in SWT2_CPB but should be in SWT2_CPB_TEST
- perfSONAR_SWT2_CPB_testbed_latency is in SWT2_CPB but should be in SWT2_CPB_TEST.

Can you fix these?  Or would you prefer I take a whack at it?

ptmcguigan:
Many thanks if you can rearrange them,

I am afraid that I would screw it up again.

Patrick

PR #233, opened 2018-10-12
verify projects

edquist:
in particular, the project Name has to match the filename.

this is dumb/redundant, but maybe we oughta have a check for it until
we remove the Name field.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,4 @@ script:
   - ./src/tests/verify_downtimes.py
   - ./src/tests/verify_unique_downtime_ids.sh
   - ./src/tests/verify_resources.py
+  - ./src/tests/verify_projects.py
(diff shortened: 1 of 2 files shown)


matyasselmeci:
I dunno. Changing `src/webapp/project_reader.py` to take the `Name` field from the filename instead of the contents is trivial, and then we can get rid of the `Name` field from the project yaml files.

edquist:
That'd be fine too - wasn't sure what all might have expected the project.Name field to be there, but if this is the only place for sure give it the boot.

matyasselmeci:
The data structure should look the same but the value of the Name field should be taken from the filename instead of the Name line in the YAML.

matyasselmeci:
I'm closing this since the proper solution is to get rid of the Name field.

PR #238, opened 2018-10-17
add github pull_request webhook handler to app.py (SOFTWARE-3325)

edquist:
This is a squash of edquist:wip/SOFTWARE-3325.downtime-automerge-webhook,
which contains the original history of this webhook.

I'll reference one commit here below from the many squashed, as we may
want to go back to using `pull/N/merge` / `merge_commit_sha`, if we only
want to allow auto-merging the initial PR (action="opened"), rather than
on every push to the head branch (action="synchronize").

---

commit 547d418020fcf4dca30b9fd9915b9dd172b7a9a1
Author: Carl Edquist
Date:   Tue Oct 16 18:38:19 2018 -0500

    use pull/N/head instead of merge_commit_sha (SOFTWARE-3325)

    apparently pull/N/merge and merge_commit_sha only get updated when the
    PR is initially created, and then upon explicit api request, which then
    requires further polling until the mergeable attribute is set ...

    instead we'll just do a separate check that the pull commit contains the
    base commit, and we now explicitly need to fetch both refs.

    https://developer.github.com/enterprise/2.1/changes/2013-04-25-deprecating-merge-commit-sha/

    https://developer.github.com/v3/git/#checking-mergeability-of-pull-requests

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -7,6 +7,8 @@ from flask import Flask, Response, request, render_template
 import logging
 import os
 import re
+import subprocess
+from subprocess import PIPE
 import sys
 import urllib.parse
 
@@ -55,6 +57,7 @@ if not app.config.get("SECRET_KEY"):
 
 global_data = GlobalData(app.config)
 
+src_dir = os.path.abspath(os.path.dirname(__file__))
 
 def _fix_unicode(text):
     """Convert a partial unicode string to full unicode"""
@@ -193,6 +196,87 @@ def generate_downtime():
                        edit_url=edit_url, site_dir_url=site_dir_url,
                        new_url=new_url)
 
+
+@app.route("/pull_request_hook", methods=["POST"])
+def pull_request_hook():
+    event = request.headers.get('X-GitHub-Event')
+    if event == "ping":
+        return Response('Pong')
+    elif event != "pull_request":
+        return Response("Wrong event type", status=400)
+
+    payload = request.get_json()
+    action = payload['action']
+    if action not in ("opened", "edited", "reopened", "synchronize"):
+        return Response("Not Interested")
+    # status=204 : No Content
+
+    sender     = payload['sender']['login']
+
+    head_sha   = payload['pull_request']['head']['sha']
+    head_label = payload['pull_request']['head']['label']
+    head_ref   = payload['pull_request']['head']['ref']
+
+    base_sha   = payload['pull_request']['base']['sha']
+    base_label = payload['pull_request']['base']['label']
+    base_ref   = payload['pull_request']['base']['ref']
+
+    pull_num   = payload['pull_request']['number']
+    pull_url   = payload['pull_request']['html_url']
+
+    pull_ref   = "pull/{pull_num}/head".format(**locals())
+
+    # make sure data repo contains relevant commits
+    stdout, stderr, ret = fetch_data_ref(base_ref, pull_ref)
+
+    if ret == 0:
+        script = src_dir + "/tests/automerge_downtime_ok.py"
+        cmd = [script, base_sha, head_sha, sender]
+        stdout, stderr, ret = runcmd(cmd, cwd=global_data.topology_data_dir)
+
+    OK = "Yes" if ret == 0 else "No"
+
+    subject = "Pull Request {pull_url} {action}".format(**locals())
+
+    out = """\
+In Pull Request: {pull_url}
+GitHub User '{sender}' wants to merge branch {head_label}
+        (at commit {head_sha})
+into {base_label}
+        (at commit {base_sha})
+
+Eligible for downtime automerge? {OK}
+
+automerge_downtime script output:
+---
+{stdout}
+---
+{stderr}
+---
+""".format(**locals())
+
+    recipients = ["edquist@cs.wisc.edu"]
+    _,_,_ = send_mailx_email(subject, out, recipients)
+
+    return Response(out)
+
+
+def runcmd(cmd, input=None, **kw):
+    if input is None:
+        stdin = None
+    else:
+        stdin = PIPE
+    p = subprocess.Popen(cmd, stdout=PIPE, stderr=PIPE, stdin=stdin,
+                         encoding='utf-8', **kw)
+    stdout, stderr = p.communicate(input)
+    return stdout, stderr, p.returncode
+
+def fetch_data_ref(*refs):
+    return runcmd(['git', 'fetch', 'origin'] + list(refs),
+                  cwd=global_data.topology_data_dir)
+
+def send_mailx_email(subject, body, recipients):
+    return runcmd(["mailx", "-s", subject] + recipients, input=body)
 
 def _make_choices(iterable, select_one=False):
     c = [(_fix_unicode(x), _fix_unicode(x)) for x in sorted(iterable)]
--- a/src/app.py
+++ b/src/app.py
@@ -255,7 +255,7 @@ automerge_downtime script output:
 ---
 """.format(**locals())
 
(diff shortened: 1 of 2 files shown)


src/app.py
```diff
@@ -194,6 +197,87 @@ def render_form(**kwargs):
...
+                         encoding='utf-8', **kw)
+    stdout, stderr = p.communicate(input)
+    return stdout, stderr, p.returncode
+
+def fetch_data_ref(*refs):
+    return runcmd(['git', 'fetch', 'origin'] + list(refs),
+                  cwd=global_data.topology_data_dir)
+
+def send_mailx_email(subject, body, recipients):
+    return runcmd(["mailx", "-s", subject] + recipients, input=body)
```
matyasselmeci: Can you make this run in the background instead? You're not getting the results so don't block the webapp.

src/app.py
```diff
@@ -194,6 +197,87 @@ def render_form(**kwargs):
...
+    stdout, stderr, ret = fetch_data_ref(base_ref, pull_ref)
+
+    if ret == 0:
+        script = src_dir + "/tests/automerge_downtime_ok.py"
+        cmd = [script, base_sha, head_sha, sender]
+        stdout, stderr, ret = runcmd(cmd, cwd=global_data.topology_data_dir)
+
+    OK = "Yes" if ret == 0 else "No"
+
+    subject = "Pull Request {pull_url} {action}".format(**locals())
```
matyasselmeci: Can you put automergeable-ness right in the subject?

src/app.py
```diff
@@ -194,6 +197,87 @@ def render_form(**kwargs):
...
+    pull_num   = payload['pull_request']['number']
+    pull_url   = payload['pull_request']['html_url']
+
+    pull_ref   = "pull/{pull_num}/head".format(**locals())
+
+    # make sure data repo contains relevant commits
+    stdout, stderr, ret = fetch_data_ref(base_ref, pull_ref)
+
+    if ret == 0:
+        script = src_dir + "/tests/automerge_downtime_ok.py"
```
bbockelm: Why invoke a subprocess instead of directly executing from the python interpreter?

src/app.py
```diff
@@ -194,6 +197,87 @@ def render_form(**kwargs):
...
+
+automerge_downtime script output:
+---
+{stdout}
+---
+{stderr}
+---
+""".format(**locals())
+
+    recipients = [ x + "@cs.wisc.edu" for x in ("edquist", "matyas", "blin") ]
```
bbockelm: Can we leave a comment on the PR itself instead of emailing this set of people?

bbockelm:
I find the idea of a webapp having the ability to change its own code a bit unsettling.  Is it possible to have this run in webhook-only mode?

That way, we could have a separate endpoint for the webhook - with appropriate credentials to make merges - and disable all other routed requests that theoretically could have exploits.

Makes sense to have it in the same code base, but perhaps deploy it in different ways.

src/app.py
matyasselmeci: I think we should do that once we're out of 'alpha testing mode'.

matyasselmeci:
It might be even better if the web hook used a separate entry point and not `app.py`. It can keep using the libraries in the `webapp` dir and the same config file if it wants to; that will also keep it from blocking the web site while the web hook is working.

src/app.py
edquist: > Can we leave a comment on the PR itself instead of emailing this set of people?

Just to note it here, we can post a comment on the PR with the [GitHub Issue Comments API](https://developer.github.com/v3/issues/comments/#create-a-comment).

Which is just to send a `POST /repos/:owner/:repo/issues/:number/comments` with post data `{"body": "comment text"}`.

I presume this requires github auth, which means deciding what user to use (do we have something like a service account/user for the `opensciencegrid` org?), and also means we have to decide how to store the auth token.  I'd guess we can create a separate account with minimal permissions, and store the auth token on disk rather than in the git repo.

edquist:
> I find the idea of a webapp having the ability to change its own code a bit unsettling.

Even after adding the github api call to merge the PR, the repo containing the `app.py` running on `topology*` doesn't get automatically updated.  As I understand it, somebody that knows what they're doing has to manually git pull the app repo.  So it won't be updating itself.  Also, a key part of the mergeability check script is to confirm that no files other than downtime files are modified (and additional requirements on the downtime files that are modified).  Of course that relies on the correctness of the check.  But I think the most important thing is the code getting executed only gets updated manually by a human.

Anyway, i'm not sure if i've misunderstood your concern - please clarify if i've missed something.

src/app.py
edquist: >Why

I like the idea of process separation -- let the check script do whatever it needs to do, and don't worry about the check having side effects on `app.py`'s python interpreter.  Eg, at some point in its development history, the check script called `chdir`, which i think should be fair game for a separate script to do.  Now the `chdir` is done via the `cwd` parameter passed to `subprocess.Popen`, which still happens in the subprocess.

Also, even the fact that the check script happens to be named with a `.py` extension, i consider incidental.  I'd rather not have to care at all what language it's implemented in.

And, why _not_? I don't really see any disadvantages, apart from the small bit of overhead of invoking a second python interpreter.

bbockelm:
> Even after adding the github api call to merge the PR

That implies the deployed webapp has the credentials necessary to merge a PR, right?

If the topology application itself has a security vulnerability in the topology web _outside_ the GitHub webhook, then an attacker may be able to trick the webapp into merging bad topology data.  As we try to automate factory configuration, that could be used to, for example, direct pilots to attacker-controlled CEs.

Now, I understand that this is not bulletproof: if the webhook itself has a security vulnerability, the game is up.  I'd just like to minimize the potential attack surface.

Make sense?

src/app.py
edquist: I've never seen `mailx` block on the topology hosts (for more than .03s), even if sending to an invalid domain ... certainly the call takes much less time to complete than the other parts of this webhook.  So i'm not really convinced that running this in the background (not waiting on the result) will be of any benefit.  Especially if we will soon be posting a PR comment rather than sending an email.

But um, despite my feedback, do you still want me to do it anyway?

matyasselmeci: I think you don't need to do this if you do as @bbockelm recommends and put the webhook in a separate webapp.

edquist:
> Make sense?

Yeah, thanks for the explanation!  The first time around i was confused by what you meant by "webhook-only mode" and the "webapp having the ability to change its own code".

So, you want the webhook to be run from a separate app/process listening on a separate port, so that the rest of the webapp will not be able to use the webhook's Merge PR credentials.  I'm all for the idea, although i don't know what kind of fancy separation/containerization would be needed to actually block the main webapp from getting ahold of the webhook app's access token.  Will we have to run them under separate users?

As for "webapp having the ability to change its own code", i think you are talking about the webapp being able to merge an arbitrary PR into the upstream master branch.  On the other hand, iiuc, the copy of `app.py` that is actually run in the webapp (along with the rest of that topology git repo checkout) is owned by `root`, while the webapp runs as the `apache` user.  So all updates to the code that actually gets run are still done manually by a human, or in any case by someone with root on the machine.

bbockelm:
> Will we have to run them under separate users?

I was thinking more simply than that: either a separate Heroku instance or just a separate VM.

matyasselmeci:
> I was thinking more simply than that: either a separate Heroku instance or just a separate VM.

If the choice is between those two, I'd prefer a separate VM. We already know how to set one of those up whereas with Heroku we have to figure out deployment, work around its limitations, etc.

matyasselmeci:
Superseded by #347

PR #244, opened 2018-10-18
Fixes for osg-topology and topology_utils (SOFTWARE-3446)

matyasselmeci:
- Add the `src` directory to `sys.path` so `bin/osg-topology` finds `topology_utils`
- Never cache the results so we don't get too little info when following up an unauth query with an auth query (or worse, too much info when following up an auth query with an unauth query)

### DIFF
--- a/src/topology_utils.py
+++ b/src/topology_utils.py
@@ -36,6 +36,7 @@ def get_auth_session(args):
         key = args.key
 
     session = requests.Session()
+    session.headers.update({'Cache-Control': 'no-cache, no-store, must-revalidate'})
 
     if os.path.exists(cert):
         session.cert = cert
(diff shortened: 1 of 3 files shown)


matyasselmeci:
Yes to both.

matyasselmeci:
OK, _now_ it works.

src/topology_utils.py
```diff
@@ -36,6 +36,7 @@ def get_auth_session(args):
         key = args.key
 
     session = requests.Session()
+    session.headers.update({'Cache-Control': 'no-cache, no-store, must-revalidate'})
```
brianhlin: Do we still need this with the `no_proxy` env var?

brianhlin:
After you merge this, can you update the following docs?

- https://opensciencegrid.org//operations/services/sending-announcements/
- https://opensciencegrid.org/technology/release/cut-sw-release/
- https://opensciencegrid.org/technology/release/cut-data-release/

PR #245, opened 2018-10-19
Decommission T3_US_UCDavis

brianhlin:
https://support.opensciencegrid.org/a/tickets/7430

Not deleting the resource group outright since they've contributed some hours over the past year (https://gracc.opensciencegrid.org/dashboard/db/site-summary?orgId=1&from=now-1y&to=now&var-interval=$__auto_interval&var-site=UC%20Davis%20CMS%20T3&var-type=Batch)


matyasselmeci:
Isn't `Disable:True` the same as deleting the entry, from GRACC's point of view?

brianhlin:
Oops, yeah. Fixed.

PR #250, opened 2018-10-24
Upgrading dCache minor version

a2bobball:
Bringing us up to date in the 4.2 release chain.


a2bobball:
copy/paste of the yaml was odd, LOTS of extra spaces, more at the beginning of each successive line.  This entry now looks formatted like the previous one.

PR #251, opened 2018-10-24
Log authenticated user DN

matyasselmeci:
This ends up in `/var/log/httpd/error_log`. This also adds a config knob to set the loglevel.


matyasselmeci:
This is running on topology-itb.

PR #261, opened 2018-10-26
Use a wrapper around yaml.safe_load instead of anymarkup so I get file name info for parse errors

matyasselmeci:
This is a PR against the `itb` branch. On parse errors, `anymarkup.parse_file` was providing less than helpful error messages like `error in <file> line 5 column 7` and not actually telling us _which_ file. This wraps `yaml.safe_load` to load the file and prints out the file path when parsing fails.

### DIFF
--- a/src/webapp/common.py
+++ b/src/webapp/common.py
@@ -9,6 +9,7 @@ import sys
 from typing import Dict, List, Union, AnyStr
 
 import xmltodict
+import yaml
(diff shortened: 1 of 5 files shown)


matyasselmeci:
WORKSFORME

matyasselmeci:
Well, it worked until I deployed it ugh

PR #264, opened 2018-10-26
Use a wrapper around yaml.safe_load

matyasselmeci:
(this time it works)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,3 @@
-anymarkup>=0.7.0,<0.8
-anymarkup-core>=0.7.1,<0.8
 certifi>=2018.4.16
 chardet>=3.0.4,<3.1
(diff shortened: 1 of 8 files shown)


matyasselmeci:
Yep. Tell you what, before you merge, I'll try killing that one too.

matyasselmeci:
I find your comment disturbing. How sure are you that no resources were dropped?

brianhlin:
Mat and I took this off-thread for a bit but to clarify my comment above, the ordering of resources in the resource dropdown of the downtime generation form were different for the facility that I spot checked (Arizona State University)

matyasselmeci:
OK, the XMLs have the same number of elements (ran them through `xml2 <$XML | wc -l`) so I'm calling this good.

PR #265, opened 2018-10-27
Add downtime for several UTA_SWT2 resouces due to network maintenance

ptmcguigan:


ptmcguigan:
I just realized after the pull request was submitted that the UTC is off by one hour.  The work will occur from 05:30 to 09:30 UTC.  Either kill this pull request and I'll submit another or adjust the submitted request for the correct time.

PR #270, opened 2018-10-31
Don't crash the webapp

matyasselmeci:
- The rg, vo, and project readers log and skip files with errors when run from the webapp
- The webapp itself also catches and logs exceptions

This should avoid 503 errors in production. When running the readers from the command line (as we do in the CI), they still die to exceptions. The webapp can also die to exceptions if running in devel mode or `STRICT=True` in the config.

### DIFF
--- a/src/webapp/project_reader.py
+++ b/src/webapp/project_reader.py
@@ -32,7 +32,8 @@ def get_projects(indir="../projects", strict=False):
             if strict:
                 raise
             else:
-                log.error("skipping")
+                # load_yaml_file() already logs the specific error
+                log.error("skipping (non-strict mode)")
                 continue
         project.update(data)
         projects.append(project)
--- a/src/webapp/project_reader.py
+++ b/src/webapp/project_reader.py
@@ -2,8 +2,11 @@
 from argparse import ArgumentParser
 from collections import OrderedDict
 
+import logging
 import os
 import sys
+
+import yaml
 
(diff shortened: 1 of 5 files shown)


matyasselmeci:
Running on topology-itb. Errors are logged to `/var/log/httpd/errors_log`.

src/webapp/project_reader.py
```diff
@@ -20,27 +26,36 @@ def get_projects(indir="../projects"):
...
         project = OrderedDict.fromkeys(["ID", "Name", "Description", "PIName", "Organization", "Department",
                                         "FieldOfScience", "Sponsor"])
-        project.update(load_yaml_file(os.path.join(indir, file)))
+        try:
+            data = load_yaml_file(os.path.join(indir, file))
+        except yaml.YAMLError:
+            if strict:
+                raise
+            else:
+                log.error("skipping")
```
brianhlin: Could we make this more specific to describe what we're skipping and why?

brianhlin: Same goes for all the other "skipping" messages

matyasselmeci: The reason there's nothing more specific is because load_yaml_file() already logs the specifics and I didn't want the duplication.

brianhlin: Well we could say that we're skipping it instead of failing because it's running in non-strict mode

PR #276, opened 2018-11-06
Sample iCal Generator

bbockelm:
Simple silly script that generates an iCal file we can import.

Not meant to merge directly, but perhaps to copy/paste into the webapp itself.


matyasselmeci:
Neat! I ran it and imported the result into Thunderbird and the calendar entry seemed normal enough. I'll see what I can do about integrating this into the webapp.

matyasselmeci:
Done in #321 .

PR #277, opened 2018-11-07
Improve error message for getting the XML arguments wrong

matyasselmeci:


bbockelm:
@matyasselmeci - are there sanitization issues here?

I see HTML being embedded in error messages and wonder if the other formatting parameters are appropriately sanitized to remove HTML meta-characters.

matyasselmeci:
@bbockelm they're string literals. I have them in a loop to avoid code duplication.

bbockelm:
Roger that.  In that case, I'm happy if you are.

PR #284, opened 2018-11-08
Autogenerate IDs for some things that don't have them

matyasselmeci:
If some entry is missing an ID, one will be generated by running the Name through a hash. This works for:
- facilities (a FACILITY.yaml file still needs to exist but can be empty)
- sites
- rgs
- NOT resources (we want to use the FQDN, not the Name, but it's not unique yet -- see SOFTWARE-3373)
- VOs
- projects


matyasselmeci:
Closing for now...

PR #285, opened 2018-11-08
Add topology entries for CHTC squid servers

matyasselmeci:


matyasselmeci:
By the way, I don't know if the FQDNAliases for DNS round-robin thing really is true: I just did a grep through are data and there are _no_ resources with more than one alias.

brianhlin:
@DrDaveD you wrote the original documentation saying that squids on a DNS round-robin should be consolidated to a single resource entry. Who at the WLCG or CERN is actually consuming the squid resource registrations and do they use `FQDNAliases`? I don't believe OIM previously supported multiple FQDNs in the `FQDN` field so I believe `FQDNAliases` is the correct way for us to specify multiple FQDNs.

DrDaveD:
The consumer is [this script](http://svnweb.cern.ch/world/wsvn/wlcgsquidmon/trunk/scripts/grid-services/get_squids.py) on wlcg-squid-monitor.cern.ch.  It only uses FQDN, not FQDNAliases.  It looks up the round-robin name and monitors each of the IP addresses in it, it does not need the indvidual names.  I don't know where the idea of FQDNAliases came from.

matyasselmeci:
So if I understand correctly, for [GGUS 138162](https://ggus.eu/index.php?mode=ticket_info&ticket_id=138162), we only need to provide the round-robin names and _not_ all of the hosts mentioned in the ticket?

DrDaveD:
Yes, if they're covered in a round-robin, you're good.

matyasselmeci:
@aaronmoate, could you take a look at this and see if it's accurate?

aaronmoate:
@matyasselmeci Looks good to me!

PR #287, opened 2018-11-12
Added registration of four CVMFS Squids

cvuosalo:


brianhlin:
@cvuosalo do these squid servers serve the same resource? Do you have a DNS round robin entry set up for these three squids perchance?

cvuosalo:
@brianhlin: cvmfscache01 and 02 serve our data and worker nodes, while cvmfscache03 and 04 serve our interactive nodes. They are not set up as round-robin.

brianhlin:
@cvuosalo: would you be able to get two DNS round-robin entries, one for cvmfscache{01,02} and one for cvmfscache{03,04}? I believe that's what the WLCG prefers for registration.

cvuosalo:
@brianhlin: I don't understand. You want round-robin entries for Squids that are not round robin???
Could you please specify the exact syntax that is required?

brianhlin:
@cvuosalo perhaps I'm misreading your initial comment. cvmfscache01 and 02 server your data and worker nodes but they're not round-robined? Similarly for cvmfscache03 and 04 for the interactive nodes?

cvuosalo:
@brianhlin: Yes, that is correct. No round robins involved.

PR #290, opened 2018-11-14
Fix New Mexico State registrations

brianhlin:
- Added AggieGrid site (the admins requested that we separate the two into two different sites)
- Updated the hosted CE contact info
- Made the site names more human-readable since they show up in the GRACC dashboards
- The resource name doesn't need to match the factory entry name so I simplified those as well

We should coordinate with factory ops so they can update their factory entries for these two CEs to reflect the new resource and site names


matyasselmeci:
Should this be merged before or after the factory entries get updated?

PR #295, opened 2018-11-21
Initial commit of DataFederations config for HCC VO

jthiltges:
Related to development discussion in SOFTWARE-3468

### DIFF
--- a/src/webapp/vos_data.py
+++ b/src/webapp/vos_data.py
@@ -31,6 +31,10 @@ class VOsData(object):
         for vo_name in sorted(self.vos.keys(), key=lambda x: x.lower()):
             try:
                 expanded_vo_data = self._expand_vo(vo_name, authorized=authorized, filters=filters)
+
+                # Remove DataFederations block from VOSummary
+                expanded_vo_data.pop('DataFederations', None)
+
                 if expanded_vo_data:
                     if 'DataFederations' in expanded_vo_data:
                         del expanded_vo_data['DataFederations']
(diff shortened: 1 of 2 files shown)


bbockelm:
@matyasselmeci - note that the schema evolved a bit between #299 and this one.  In #299, we called the list of directories {{Authorizations}} while here we call them {{Namespaces}}.

I like {{Namespaces}} better as the {{AllowedCaches}} list is also an authorization; {{Namespaces}} feels more precise.

src/webapp/vos_data.py
```diff
@@ -31,6 +31,10 @@ def get_tree(self, authorized=False, filters: Filters = None) -> Dict:
         for vo_name in sorted(self.vos.keys(), key=lambda x: x.lower()):
             try:
                 expanded_vo_data = self._expand_vo(vo_name, authorized=authorized, filters=filters)
+
+                # Remove DataFederations block from VOSummary
+                expanded_vo_data.pop('DataFederations', None)
```
matyasselmeci: This needs to be inside the `if expanded_vo_data` block to avoid a potential TypeError if expanded_vo_data is None.

bbockelm: Got this fixed in #299 also.  Just removed this hunk.

bbockelm:
I can confirm that, with these changes, we see the `/hcc` FQAN authorized to access `/hcc/focusday` and the Brian Bockelman DN can access both HCC and LIGO's areas:
```
$ PYTHONPATH=src ./bin/osg-authfile 2>/dev/null | grep hcc
g /hcc /hcc/focusday rl
u 6fb7593d.0 /hcc/focusday rl /user/ligo rl
```
@matyasselmeci - can you re-review?  I think this is ready to go.

PR #298, opened 2018-11-21
Adding DN to cache entry for origins to parse

djw8605:
Prototyping cache DN


djw8605:
For future reference, how to get the service 156 (XRootD cache)
https://topology.opensciencegrid.org/rgsummary/xml?service&service_sel[]=156

matyasselmeci:
What will this be used for? Generating the mapfile for the stats collector?

matyasselmeci:
Makes sense. So it looks like you want to expose this information in the XML? You'll have to edit `Resource.get_tree()` in `src/webapp/topology.py`. I *think* adding `"DN"` to the end of the `OrderedDict.fromkeys()` call in line 90 should be sufficient since that data requires no special processing, but it's the kind of thing we should test out on topology-itb first.

djw8605:
@matyasselmeci maybe I don't understand the code.  The XML already contains the DN without the changes you mention.

matyasselmeci:
/me reads code again
You're right, it does. The `update()` on line 94 already puts the data in there. However, since the key is not already in the OrderedDict, the ordering won't be guaranteed, which may cause schema validation problems if we were to add additional attributes in the future.

bbockelm:
> So it looks like you want to expose this information in the XML?

Actually, I wanted to ask @djw8605 about that!

What's the upside of doing this via XML versus generating an Authfile as we are doing for the caches?  I'd guess the origin admins would find it easier to integrate a flat list of DNs instead of XML into their Authfiles.

@matyasselmeci - you're thinking about running one of these.  What approach would you prefer?

djw8605:
My first goal was to get the DN into the topology.

I suppose it is more of a question of where the Authfile should be generated.  You integrated the authfile generation into Topology itself.  That was going to be my next step.  Though, I think there is an argument that could be made that whatever daemon that runs on the caches / origins could just download the XML and generate the Authfile.  It already has to download the Authfile, and one could imagine tying this daemon to the version of XRootD, if they decide to alter the format of the Authfile.

matyasselmeci:
I assumed Derek added it to the schema because he wanted it in the XML; didn't realize that it was already making it into the XML and he made the addition just to get the CI to pass.

@bbockelm are you suggesting that this is a script people will run on checkouts of the raw topology data? In the interest of making sure people have current data, I'd rather have something that queries the webapp, though the response doesn't need to be XML.

bbockelm:
@matyasselmeci - like in #299, it's probably useful to have a CLI for developers, but for the real-world, the only reasonable way is to have this downloaded from a webapp.

bbockelm:
For metadata purposes - note this is work toward SOFTWARE-3469.

Let's keep this particular PR for the changes to the resource itself; we can do the changes to the webapp for generating the DN list separately.

However, prior to merging this - I would like to have the listing of allowed VOs for the resource as well.

matyasselmeci:
I've lost track of where this stands with respect to the other SC support we're adding. Can we merge this (after update)? Should it wait for #336? Or is this no longer needed?

bbockelm:
Superseded by the other PR

PR #299, opened 2018-11-22
Add capability to generate Authfiles

bbockelm:
This adds an initial capability to generate the public and private {{Authfile}} for StashCache cache servers.  The YAML design is described in https://opensciencegrid.atlassian.net/browse/SOFTWARE-3468

This PR also includes a tweak of the VO registration for the known participants (or near-future participants) of StashCache.

It also contains a pure-python algorithm for calculating the OpenSSL-style hash given a DN.  It might be only a dozen-or-so lines long, but it was one of the trickier reverse-engineering feats I've had to put together.


src/stashcache.py
```diff
@@ -0,0 +1,136 @@
...
+    for vo_name, vo_data in vo_data.vos.items():
+        stashcache_data = vo_data.get('DataFederations', {}).get('StashCache')
+        if not stashcache_data:
+            continue
+
+        for dirname, authz_list in stashcache_data.get("Authorizations", {}).items():
+            for authz in authz_list:
+                if authz.startswith("FQAN:"):
+                    id_to_dir["g {}".format(authz[5:])].append(dirname)
+                if authz.startswith("DN:"):
```
42B: Couldn't this be `elif`?

bin/osg-authfile
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/python3.6
```
matyasselmeci: Have you tried this? Does using the system python work with packages installed inside the virtualenv?

src/app.py
```diff
@@ -120,6 +120,12 @@ def rgdowntime_xml():
     return _get_xml_or_fail(global_data.get_topology().get_downtimes, request.args)
 
 
+@app.route("/authfile")
+def authfile():
+    return get_authfile(global_data.get_topology().get_resources(),
```
matyasselmeci: Does this work? I don't see you importing `stashcache.get_authfile` anywhere...

src/stashcache.py
```diff
@@ -0,0 +1,136 @@
...
+   "L": "2.5.4.7",
+   "postalCode": "2.5.4.17",
+   "street": "2.5.4.9",
+   "emailAddress": "1.2.840.113549.1.9.1",
+   }
+
+
+__dn_split_re = re.compile("/([A-Za-z]+)=")
+
+def _generate_ligo_dns():
```
matyasselmeci: Not a big fan of having LIGO-specific code right inside the webapp -- I'd rather have some sort of configuration. If you can't fix it now, please make a JIRA ticket and assign it to me.

bin/osg-authfile
bbockelm: Yeah, this is how I run things locally.  However, given we do `venv`-based installs, I agree that an absolute path isn't the best.  Is `/usr/bin/env python3` more kosher?

bin/osg-authfile
```diff
@@ -0,0 +1,17 @@
...
+# -*- coding: utf-8 -*-
+
+"""
+A script to generate Authfiles for XCaches in the OSG
+"""
+
+import webapp.models
+import stashcache
+
+global_data = webapp.models.GlobalData()
```
matyasselmeci: This needs to be configurable -- I certainly don't have a "contacts" checkout inside my topology dir. Since you probably don't want to make your own config system, I suggest you use Flask's, like the webapp does -- see lines 38-58 of `src/app.py`.

bin/osg-authfile
matyasselmeci: Yep, that's what the other scripts do.

bin/osg-authfile
bbockelm: Right now, I see `osg-authfile` as a testing tool for developers (since the primary use will be via the web interface).  Hence, I don't worry too much about providing a separate config file for this.

I also don't have a `contacts` directory checked out -- it warns about it and moves on.

matyasselmeci: In that case, do you think it's worth adding `osg-authfile` to the CI?

bbockelm: Yes!  What's the best way to do that?

bin/osg-authfile
bbockelm: Done in the latest update.

bin/osg-authfile
matyasselmeci: For a minimal test (run it and see if it crashes), just add it to `.travis.yml`. If need to make a wrapper script, put it in `src/tests`.

src/stashcache.py
bbockelm: Right - so this is basically grandfathered in until SciTokens is more mature.  How would you like to handle it?

bin/osg-authfile
bbockelm: Done in the latest commit.

src/stashcache.py
bbockelm: Decided to go ahead and make this a config knob.  Resolving for now.

PR #307, opened 2018-12-05
Scheduled CIT_CMS_T2 downtime on Dec 10 - Dec 11

juztas:
We have accumulated several upgrades pending for the whole Tier2 and we schedule downtime for 24h from Dec 10th 16:00 +0000 for 24h.


juztas:
@jpata fyi

bbockelm:
@matyasselmeci - it appears something broke in the authfile verification unrelated to this PR.  Can you take a look at it?

@juztas - the downtime itself looks fine so I'll go ahead with the merge.

PR #308, opened 2018-12-06
Use secure connection to LIGO LDAP.

bbockelm:


bbockelm:
@brianhlin - this fixes the build failures.

PR #317, opened 2018-12-13
Update ENMR.yaml

marcoverl:
updated with my new X509 personal certificate DN

### DIFF
--- a/virtual-organizations/ENMR.yaml
+++ b/virtual-organizations/ENMR.yaml
@@ -50,7 +50,7 @@ MembershipServicesURL: https://voms2.cnaf.infn.it:8443/voms/enmr.eu/
 OASIS:
   Managers:
     Marco Verlato:
-      DNs: /C=IT/O=INFN/OU=Personal Certificate/L=Padova/CN=Marco Verlato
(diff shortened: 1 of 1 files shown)


virtual-organizations/ENMR.yaml
```diff
@@ -50,7 +50,7 @@ MembershipServicesURL: https://voms2.cnaf.infn.it:8443/voms/enmr.eu/
 OASIS:
   Managers:
     Marco Verlato:
-      DNs: /C=IT/O=INFN/OU=Personal Certificate/L=Padova/CN=Marco Verlato
+      DNs: /DC=org/DC=terena/DC=tcs/C=IT/O=Istituto Nazionale di Fisica Nucleare/CN=Marco Verlato verlato@infn.it
```
brianhlin: Hi Marco,

Is the email address really part of the CN? I don't think I've ever seen that before!

Thanks,
Brian

marcoverl: No, verlato@infn.it is not a mail address, it is just part of my identity name in the INFN IdP

PR #321, opened 2018-12-15
iCal generator script and route

matyasselmeci:
#276 merged into the webapp. Accepts the same filters and parameters the downtime XML does. The script is also included.


matyasselmeci:
This is running on topology-itb right now.

bbockelm:
Looks great!  I was able to view it on my iCal.  Let's push it through.

PR #329, opened 2018-12-20
handle missing resource group files (SOFTWARE-3325)

edquist:
this fixed the backtrace we saw earlier in a topo merge email

### DIFF
--- a/src/tests/automerge_downtime_ok.py
+++ b/src/tests/automerge_downtime_ok.py
@@ -109,15 +109,24 @@ def commit_is_merged(sha_a, sha_b):
     ret, out = runcmd(args, stderr=_devnull)
     return ret == 0
 
-def get_downtime_dict_at_version(sha, fname):
+def parse_yaml_at_version(sha, fname):
     txt = get_file_at_version(sha, fname)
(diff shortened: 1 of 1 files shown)


src/tests/automerge_downtime_ok.py
```diff
@@ -116,8 +116,10 @@ def get_downtime_dict_at_version(sha, fname):
 
 def get_rg_resources_at_version(sha, fname):
     txt = get_file_at_version(sha, fname)
+    if not txt:
+        return {}
     rg = yaml.safe_load(txt)
-    return rg["Resources"]
+    return rg.get("Resources") or {}
```
matyasselmeci: 'get' takes a second parameter for what to use as the default.
```suggestion
     return rg.get("Resources", {})
```
Also, what if yaml.safe_load fails?

edquist: I can add a non-exception-throwing wrapper around this which (more gently) writes some error message to stderr

edquist:
Alright @matyasselmeci, i think this is better.  It'll include yaml parse errors in the list of failures.

PR #331, opened 2018-12-21
separate webhook app & updates (SOFTWARE-3325)

edquist:
This is what's been running on topology-itb, which includes the separate app for handling pull_request webhooks and emailing the downtime-automergeability check results.

The work for performing the actual automerge is on a separate branch, but i'd like to get this here merged as it's been getting stale.  I've rebased on top of `master` just to make the commit history a bit cleaner.


src/webapp/models.py
```diff
@@ -55,13 +55,28 @@ def __init__(self, config=None, strict=False):
...
         self.projects_dir = os.path.join(self.topology_data_dir, "projects")
         self.topology_dir = os.path.join(self.topology_data_dir, "topology")
         self.vos_dir = os.path.join(self.topology_data_dir, "virtual-organizations")
         self.config = config
         self.strict = strict
 
+    def _update_webhook_repo(self):
+        if not self.config["NO_GIT"]:
+            parent = os.path.dirname(self.webhook_data_dir)
+            os.makedirs(parent, mode=0o755, exist_ok=True)
```
matyasselmeci: "mode" means something different in makedirs: it's ANDed against the umask.

PR #343, opened 2019-01-08
UCSD T2 outage due to power out

jdost321:


topology/University of California San Diego/UCSD CMS Tier2/UCSDT2_downtime.yaml
```diff
@@ -388,3 +388,47 @@
   Severity: Outage
   StartTime: Mar 28, 2018 21:00 +0000
 # ----------------------------------------------------------
+- Class: SCHEDULED
+  ID: 120032287
+  Description: power outage
```
edquist: Was this `power outage` in fact `SCHEDULED` ?

jdost321: Yes, it was. We knew about it in advance but Terrence didn't know how to mark it in downtime, so I just gave it a shot. Also, we weren't sure if duplicate id's were ok.. I just copy pasted one entry for the CE to all of the others. In the future is it ok to just bump the integer manually for each CE?

jdost321: Let's leave it as Terrence for now. Also, I just added a missing SE downtime to the original commit.

PR #348, opened 2019-01-11
let the shell do its own quoting

edquist:
happened to see how this was implemented when i was trying to figure out how to git-push with the configured ssh key.  it crossed some red-line threshold for me so i tried to digest it, and this is what came out.

anyway whaddya think?


matyasselmeci:
Looks good to me; can you try it out on topology-itb, just in case?

matyasselmeci:
We discussed an entirely better way of calling out to git over IM (using GIT_SSH_COMMAND instead of ssh-agent); if you're still willing to take the time to test this PR on topology-itb, I can merge it; if you don't want to do that, then let's close the PR in favor of the better solution.

PR #349, opened 2019-01-11
Unscheduled CIT downtime

juztas:
There are power issues in one of the racks and also several other datanodes are down and affect all HDFS. Files are not lost, but nodes are not accessible and this affects all SE. 

p.s. There is no xrootd as a service for SE, shouldn't this be implemented in some way, same as xrootd caches? Or this is not something what CMS would look/need?


juztas:
correct, I need to add these things to site topo... Thanks!

PR #351, opened 2019-01-11
allow passing git_dir explicitly to run_git_cmd (SOFTWARE-3325)

edquist:
previously, passing a `dir` parameter to `run_git_cmd` would be interpreted as a working-tree, and it would assume that the git-dir was also in a `.git` subdir.

in the case of the webhook app, the webhook data dir is a bare repo; that is, a git-dir without a working-tree.

by accepting an explicit `git_dir` parameter, this `run_git_cmd` can now be run on a bare git repo, as well as non-bare repos with a git-dir and working-tree in separate locations.


edquist:
Added docstring per @matyasselmeci's request

PR #353, opened 2019-01-17
Use LibYAML for faster loading of topology data

jthiltges:
Quick testing with rg_reader.py shows a 4x speed-up after switching to the PyYAML CSafeLoader: 8.8 sec down to 1.9 sec.

### DIFF
--- a/src/tests/automerge_downtime_ok.py
+++ b/src/tests/automerge_downtime_ok.py
@@ -125,7 +125,7 @@ def parse_yaml_at_version(sha, fname, default):
     if not txt:
         return default
     try:
-        return yaml.safe_load(txt)
+        return yaml.load(txt, Loader=yaml.CSafeLoader)
     except yaml.error.YAMLError:
         return None
 
--- a/src/tests/automerge_downtime_ok.py
+++ b/src/tests/automerge_downtime_ok.py
@@ -12,6 +12,11 @@ try:
 except ImportError:
     from urllib2 import urlopen
 
+try:
+    from yaml import CSafeLoader as SafeLoader
+except ImportError:
+    from yaml import SafeLoader
+
 import xml.etree.ElementTree as et
 
 
@@ -125,7 +130,7 @@ def parse_yaml_at_version(sha, fname, default):
     if not txt:
         return default
     try:
-        return yaml.load(txt, Loader=yaml.CSafeLoader)
+        return yaml.load(txt, Loader=SafeLoader)
     except yaml.error.YAMLError:
         return None
 
--- a/src/tests/verify_downtimes.py
+++ b/src/tests/verify_downtimes.py
@@ -8,6 +8,11 @@ import sys
 import os
 import re
 
+try:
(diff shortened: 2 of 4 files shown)


matyasselmeci:
Jeez, here I am writing a ticket to write something to pickle all the data for performance reasons and you come up with this :)

It looks good; I will test it on topology-itb first and approve the PR if it works.

djw8605:
I still vote for pickle'ing in addition to this.  But this is certainly a good improvement.

jthiltges:
Sorry for the confusion. I believe it should work as committed. PyYAML should already include LibYAML.

```
$ docker run -it python:3.6 /bin/bash

root@5a300d26ce5b:/# pip3 install PyYAML
...
root@5a300d26ce5b:/# python3
Python 3.6.7 (default, Oct 24 2018, 22:47:56) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> print(yaml.load("hello: world", Loader=yaml.CSafeLoader))
{'hello': 'world'}
```

matyasselmeci:
Not getting that result:
```
(venv) [root@hcc-osg-topology2 topology-itb]# pip install PyYAML
Requirement already satisfied: PyYAML in ./venv/lib/python3.6/site-packages (4.2b4)
(venv) [root@hcc-osg-topology2 topology-itb]# python
Python 3.6.3 (default, Jan  4 2018, 16:40:53) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> print(yaml.load("hello: world", Loader=yaml.CSafeLoader))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'yaml' has no attribute 'CSafeLoader'
```

jthiltges:
Ah, it looks like the libyaml-devel RPM needs to be present before PyYAML is installed for the `site-packages/_yaml.cpython-*.so` to be built. Reinstalling PyYAML should hopefully sort it out.

The Debian based python:3.6 Docker image didn't show the issue. Sorry for the trouble.

I confirmed using these steps, starting from a CentOS image this time:
```
$ docker run -it --rm=true centos:7 /bin/bash
# yum install -y centos-release-scl
# yum install -y rh-python36 gcc libyaml-devel
# scl enable rh-python36 bash
# python -m venv topo
# source topo/bin/activate
# pip install 'PyYAML>=4.2b1,<=4.2'
```

matyasselmeci:
Looks like that works. I updated the log message for the fallback. Can you add Puppet code to install libyaml-devel on the topology machines?

djw8605:
I added the puppet code to add `libyaml-devel` to topology.  Can take up to 15 minutes for puppet to run on the nodes.

matyasselmeci:
OK, both topology and topology-itb are running the new code and I reinstalled PyYAML on both of them. It's a lot faster now!

PR #383, opened 2019-01-24
Adding Freshdesk support widget

djw8605:


brianhlin:
Does this result in the widget appearing on all the topology pages, e.g. the XML pages?

djw8605:
Nope, only the page which import the "base.html.j2" file.  So only the HTML pages.  Not the XML pages.

PR #387, opened 2019-01-28
Add AllowedOrigins and AllowedCaches entries for all VOs in StashCache

matyasselmeci:
There need to be _some_ values, even if they're empty, for authfile
generation to work.  Some of the values will need to be replaced once we
know what the origin servers are.


matyasselmeci:
Better?

PR #399, opened 2019-01-30
Initial osg-topology Dockerfile

brianhlin:
Just something I was playing around with yesterday. Currently requires mounting a user proxy to /user.proxy to view emails. I'd like to also support the case where users mount their cert/key so they don't have to have `{voms,grid}-proxy-init` installed locally. I think the main hurdle there is getting certs installed but I haven't gotten that far.

I learned a bunch about docker image development practices and niceties of multi-stage builds from here: https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3?gi=fc1cb2613305

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:2.7-alpine as base
+
+FROM base as builder
+
+RUN mkdir /install
+WORKDIR /install
+
+COPY requirements-rootless.txt /
+
+RUN pip install --install-option="--prefix=/install" -r /requirements-rootless.txt
+
+FROM base
+
+LABEL maintainer OSG Software <help@opensciencegrid.org>
+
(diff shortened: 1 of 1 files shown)


matyasselmeci:
I like the idea and also thank you for the link. After we split the repos, `osg-topology` is on my short list of scripts to modify to use the web server instead of the libraries...

I don't think you need CA certs to use the user cert/key so if you mount `~/.globus`, you should be good -- try it out!

I wonder what this will all look like in our Glorious COManage Future. :)

matyasselmeci:
and now we already have a Dockerfile, but it's for the webapp. What do you want to do with this issue? Rename your new Dockerfile? Close it?

brianhlin:
Let's just abandon this

PR #401, opened 2019-01-31
Add ticket opening from topology

djw8605:
Still need to be completed:

- [x] [Captcha](https://developers.google.com/recaptcha/docs/v3) from Google

This adds 2 new configuration values:

- `GOOGLE_CAPTCHA_KEY`: Registered CAPTCHA secret key.
- `FRESHDESK_API_KEY`: API key for FreshDesk. I believe each person has their own API key, and I don't know how to make it general for the entire project.


djw8605:
Read for review.

src/app.py
```diff
@@ -108,6 +110,65 @@ def contacts():
...
+                resource_tree = resource.get_tree(authorized = True)
+                contact_list = resource_tree['ContactLists']['ContactList']
+                for contact in contact_list:
+                    if contact["ContactType"] == "Administrative Contact":
+                        for inner_contact in contact['Contacts']['Contact']:
+                            cc_emails.add(inner_contact['Email'])
+
+    print(cc_emails)
+    ticket = {
+        'subject': "Test API ticket",
```
matyasselmeci: We should change the default subject before this goes live.

bbockelm:
@djw8605 - can you attach some screenshots to the ticket so we can see what the UI looks like?

Also - I don't want to deploy any ticket UI that might ask for an X509 certificate.  Can we limit this to a specific vhost?

brianhlin:
> @djw8605 - can you attach some screenshots to the ticket so we can see what the UI looks like?

Or better yet, deployment to the ITB?

djw8605:
Multiple resources select with smart search.
image

It will add all of the "Admin" contacts for those resources.

brianhlin:
Do we have a drop-down for VOs in case sites are having issues with a specific VO's?

djw8605:
Nope, for now, this was specific to opening ticket against sites.

src/templates/ticket.html.j2
```diff
@@ -0,0 +1,150 @@
...
+        <label for="requester">Requester Email</label>
+        <input type="email" name="requester" class="form-control" id="requester" required placeholder="Requester Email" pattern="[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$"></input>
+    </div>
+    <div class="form-group">
+        <label for="Subject">Subject</label>
+        <input type="text" name="subject" class="form-control" id="subject" required></input>
+    </div>
+    <div class="form-group">
+      <label for="ticket-message">
+        Ticket Content
```
brianhlin: I prefer "Description of the issue" or simply "Description"

src/templates/ticket.html.j2
```diff
@@ -0,0 +1,150 @@
...
+    <div class="form-group">
+      <label for="resource-list">Resource</label>
+      <select name="resource-list" id="resource-list" multiple placeholder="Pick resources..."></select>
+    </div>
+    <div class="form-group">
+        <label for="requester">Requester Email</label>
+        <input type="email" name="requester" class="form-control" id="requester" required placeholder="Requester Email" pattern="[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$"></input>
+    </div>
+    <div class="form-group">
+        <label for="Subject">Subject</label>
```
brianhlin: "Summary of the issue" or "Summary"

src/templates/ticket.html.j2
```diff
@@ -0,0 +1,150 @@
...
+
+
+{% endblock %}
+
+{% block content %}
+<div class="container">
+
+  <form action="/submitTicket" method="POST">
+    <div class="form-group">
+      <label for="resource-list">Resource</label>
```
brianhlin: > Pick OSG Resource(s):

src/templates/ticket.html.j2
djw8605: I like "Description of the issue"

src/templates/ticket.html.j2
djw8605: That "Pick Resource(s)" is the placeholder text in the search box.  Should it be repeated?

brianhlin: Ah, I missed that. `Resource(s) is fine`

djw8605:
Added a few sentences at the top:
image

brianhlin:
That looks good. I forgot to ask, can we enable the Freshdesk "support" widget on this page, too?

djw8605:
the freshdesk support widget is enabled.  You can see it in the screenshot on the right.

brianhlin:
Oops, so it is. Could we get a link added to the main page under "Human-readable pages"?

src/app.py
brianhlin: I believe this is fixed. Resolving

src/templates/base.html.j2
```diff
@@ -8,6 +8,9 @@
       href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.1.3/css/bootstrap.min.css" 
       integrity="sha256-eSi1q2PG6J7g7ib17yAaWMcrr5GrtohYChqibrV7PBE="
       crossorigin="anonymous" />
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js"
```
matyasselmeci: What's the purpose of this change? If want_js is false, then javascript shouldn't get loaded...

djw8605: JQuery needs to be loaded before the script section so that `$` is defined.  I suppose this should be surrounded by the `want_js` if's.

matyasselmeci: Actually it looks like you added the Freshdesk Support Widget to base.html.j2 which is all JavaScript, so at this point I guess you might as well get rid of `want_js` altogether...

djw8605: Removed want_js

matyasselmeci: Can you edit contacts.html.j2 and generate_downtime_form.html.j2 to remove the `set want_js = true` lines?

brianhlin:
Let's hold off on merging this until we clear up the issues mentioned here: https://support.opensciencegrid.org/a/tickets/8380

djw8605:
The issues mentioned in that ticket are related to public ticket, not necessarily with this ticket creation.  I see them as unrelated.

src/app.py
```diff
@@ -108,6 +112,69 @@ def contacts():
...
+
+    api_key = app.config['FRESHDESK_API_KEY']
+    url_request = requests.post("https://opensciencegrid.freshdesk.com/api/v2/tickets", 
+                                data = json.dumps(ticket),
+                                auth = (api_key, 'x'),
+                                headers = headers)
+
+    if url_request.status_code in [ 200, 201 ]:
+        json_return = url_request.json()
+        return_str = "Created Ticket {}".format(json_return['id'])
```
brianhlin: We should tell users that they'll receive an email shortly. Would it be possible to have this show up as a message on the form page? The page with the return_str is a little bare.

matyasselmeci:
This seems like a useful tool but it's five years old and hasn't been merged. Sorry; closing.

PR #402, opened 2019-02-04
Add SDSC-PRP-CE

sfiligoi:


matyasselmeci:
I'm confused. Is this a CE or a Cache?

sfiligoi:
This is a CE.
I just happens to be hosted on a node that has "cache" in the DNS name.

matyasselmeci:
OK.
You don't need an AllowedVOs section if it's not a StashCache cache or origin.

PR #411, opened 2019-02-06
Retire old LIGO resource at LSU

brianhlin:
Edgar/Marco are working on setting up a hosted CE with another group
for LIGO


brianhlin:
I honestly don't know what we use the test XML for. IIRC, the schema tests generate the XML each time. @matyasselmeci ?

matyasselmeci:
The test XMLs were for testing initial translation (i.e. the XML -> YAML bits). I'm not sure we need them anymore.

PR #416, opened 2019-02-06
Remove XMLs and converters

matyasselmeci:
These were needed for the initial conversion but are only of historical value and created noise in searches.


matyasselmeci:
I tagged the last commit `drop_converters`; we can get the converters back by referring to its parent i.e. `drop_converters~`

PR #418, opened 2019-02-07
Drop I2U2 VO

brianhlin:
And update the list of active VOs. Not really sure why the diff is so ugly for the `README.md` but it's what my one-off script produced /shrug


matyasselmeci:
You added <> around all the links, that's the difference. Does #36 depend on this?

brianhlin:
Well, that's what I get for not putting my one-off script in version control, I guess. The `<>` are nice to have because then the URLs render properly after being passed through static HTML generators.

I assume you meant https://github.com/opensciencegrid/opensciencegrid.github.io/pull/36? The PRs can be treated independently

PR #431, opened 2019-02-09
Document web API for XML pages

matyasselmeci:


matyasselmeci:
https://github.com/opensciencegrid/topology/pull/431

README.md
```diff
@@ -54,6 +54,85 @@ Alternatively, the registry data can be accessed in XML format at the following
...
+#### For the Resource Topology (rgsummary) and Resource Downtime (rgdowntime) pages
+
+Boolean filters:
+
+| The following argument(s)... | Will have the effect of...                         |
+|------------------------------|----------------------------------------------------|
+| `active=on&active_value=0`   | showing only inactive resources                    |
+| `active=on&active_value=1`   | showing only active resources                      |
+| `disable=on&disable_value=0` | showing only disabled resources                    |
+| `disable=on&disable_value=1` | showing only enabled resources                     |
```
brianhlin: It seems like maybe /we/ got this wrong. Reading this over, I would expect `disable_value=1` to correspond to showing only disabled resources.

README.md
```diff
@@ -54,6 +54,85 @@ Alternatively, the registry data can be accessed in XML format at the following
...
+
+ID-based filters:
+
+These select by a numeric ID on facilities, sites, support centers, resource groups, and services, and VOs.
+Multiple IDs may be specified for an attribute, in which case resources that match any of the IDs listed will be shown.
+There are two equivalent formats for the arguments:
+- `<ATTRIB>_<ID1>=on&<ATTRIB>_<ID2>=on&<ATTRIB>_<ID3>=on`
+- `<ATTRIB>_sel[]=<ID1>&<ATTRIB>_sel[]=<ID2>&<ATTRIB>_sel[]=<ID3>`
+
+| The following attributes... | Will filter by...             |
```
brianhlin: For continuity, I think this should read something like this:

```suggestion
| The following attributes (`<ATTRIB>`)... | Will filter by (`<ID>`)...             |
```

README.md
```diff
@@ -54,6 +54,85 @@ Alternatively, the registry data can be accessed in XML format at the following
...
+| `rg`                        | ResourceGroup GroupID         |
+| `service`                   | Service ID                    |
+| `sc`                        | SupportCenter ID              |
+| `voown`                     | VO Ownership (rgsummary only) |
+
+For example, either of the two following sets of arguments will show resources that are in ANL (10089), Fermilab (10009), or UW-Madison (10011):
+- `facility_10089=on&facility_10009=on&facility_10011=on`
+- `facility_sel[]=10089&facility_sel[]=10009&facility_sel[]=10011`
+
+If multiple filters are specified, the resources must match all of them to be shown.
```
brianhlin:
```suggestion
If multiple filters are specified, only resources that match all filters will be shown.
```

README.md
```diff
@@ -54,6 +54,85 @@ Alternatively, the registry data can be accessed in XML format at the following
...
+
+Finally, you may restrict how much _past_ downtime is shown in the Resource Downtime XML with:
+- `downtime_attrs_showpast=all`: show all past downtime
+- `downtime_attrs_showpast=7`: show downtime that ended no earlier than 7 days (168 hours) ago
+- `downtime_attrs_showpast=0`: show no past downtime; this is the default if not specified
```
brianhlin: Are the only acceptable values `all`, `7`, and `0`? Because that's the impression that I get reading this

README.md
matyasselmeci: Nope. I checked their source; if you don't have `disable=on`, `disable_value` does nothing.  Which is good because of the number of people that have `active=on&active_value=1&disable_value=1` in their queries, which would show resources that are both "active" and "disabled" which is nonsensical.

README.md
matyasselmeci: Fixed.

brianhlin: I think the parenthetical is confusing, am I supposed to enter number of days or number of hours? An example would be illustrative:

```
- `downtime_attrs_showpast=<DAYS>`: show downtime that ended no earlier than `<DAYS>` days ago. For example, `downtime_attrs_showpast=7` would result in showing all downtimes that ended within the past 7 days (168 hours).
```

README.md
brianhlin: Ah ok, I was confused by our earlier conversation, then. Yikes.

README.md
matyasselmeci: OK, better?

PR #433, opened 2019-02-11
WLCG now only requires 24 hours notice of downtime (see FD #8275)

matyasselmeci:


matyasselmeci:
The question of validating times against scheduled/unscheduled came up at one point but we decided that enforcing WLCG policy was none of our business so the webform only gives a warning.

PR #436, opened 2019-02-12
Revert "ITB test downtime for VDT-ITB-MADISON"

edquist:
This reverts commit fc0f224fcca5cfd9cda62d35efd2042dea4cf964.


matyasselmeci:
I think the middle comment is enough.

I do appreciate your newfound love of emoji, but does anyone else think that :+1: looks a lot like :middle_finger: at a distance?

brianhlin:
Agreed, on only needing the one comment. I'd also prefer that we spell out "automatic merge" instead of using "automerge". We should also leave a comment about requesting help via our help@ email address.

edquist:
Glad you like the emoji!

No i do not think the standard +1 signal looks offensive.

Sheesh you guys are picking apart osg-bot when he's just learning to dream.

Anyway yeah in this case, where everything goes well, the middle approval comment seems sufficient, but there are a number places where things can go wrong. See #437, #438, #439, for starters, but also (not yet showcased in an existing PR) after the ci checks and automerge approval message, if the merge fails because two eligible PRs were opened in quick succession, it seems appropriate to add a final comment about the merge error.

Also, just something to keep in mind, if we only keep the middle message, that means potentially having to wait a long time without seeing anything if the ci build hangs for a long time (which does occasionally happen).

Maybe i just enjoy osg-bot's first comment appearing a split second after the PR is created?  :)

I think what you are after is to post as little as possible, and i think that could mean one comment if everything goes right, or possibly more if there's an unexpected error.

I _do_ also really like osg-bot saying

>Have a nice day! :racehorse: 

though i feel like it's most appropriate only if and after everything succeeds...

So you see i have some mixed feelings.  But yeah i can see it being best to only have a single comment.

Anyway since you guys are clearly itching to unleash osg-bot on the world, i will open up a `WIP` PR for the actual github api work...  which will probably be a better home for this kind of discussion.

PR #458, opened 2019-02-14
test double dt merge conflict 2

edquist:


matyasselmeci:
Can you make it skip the first message?

edquist:
>Can you make it skip the first message?

Actually @matyasselmeci that's a fine idea.

The thing is the first message is actually the PR review approval, and the second message only shows up after the \[Merge Button] attempt if the merge attempt unexpectedly fails.  (eg, a "mid-air collision" of multiple approved PRs.)

The PR review approval really does belong before the merge attempt (and is required if we hit the merge button instead of doing the manual merge+git-push thing), but we could simply issue a PR approval with no message, and then osg-bot can make a single comment after the merge succeeds or fails.

How does that sound?

matyasselmeci:
Sounds good. We just don't want to get the user's hopes up and then have them dashed again :grin:

edquist:
Makes sense.

As a non-user, i kinda liked watching the play-by-play in eager anticipation with my :popcorn:, but we don't need the suspense to be that killer for the real users.

PR #459, opened 2019-02-14
Add StashCache to VO template (SOFTWARE-3518)

matyasselmeci:


template-virtual-organization.yaml
```diff
@@ -85,3 +85,16 @@ SupportURL: http://www.usatlas.bnl.gov/index.shtml
...
+### DataFederations (optional) is information about data federations that the VO is part of
+### This template is for the StashCache data federation; for instructions,
+### see the documentation at https://opensciencegrid.org/docs/data/stashcache/vo-data/
+# DataFederations:
+#   StashCache:
+#     Namespaces:
+#       /<VONAME>/PUBLIC:
+#         - ANY
+#     AllowedCaches:
+#       - ANY
```
brianhlin: Do we want this to be the default suggestion? I'm inclined to say that we should just have:

```
- <CACHE_RESOURCE>
- ...
```

template-virtual-organization.yaml
```diff
@@ -85,3 +85,16 @@ SupportURL: http://www.usatlas.bnl.gov/index.shtml
...
+#     AllowedOrigins:
+#       - <ORIGIN_RESOURCE>
```
brianhlin: Since there can be many, should we add `- ...`?

matyasselmeci: Done

template-virtual-organization.yaml
matyasselmeci: Done

PR #468, opened 2019-02-20
CIT_CMS_T2 downtime due to power maitenance,renovations.

juztas:
Part of the Tier2 cluster will be without electricity on Saturday, February 23rd due to needed maitenance and renovations. As this happens on weekend, I extend this downtime from Friday to Monday. In case all will be back sooner, we will shorten it.


bbockelm:
@juztas - @edquist is looking in to why the robot didn't auto-merge this.  One moment please...

PR #475, opened 2019-02-21
Filter by FQDN for osg-notify and osg-topology

matyasselmeci:


bin/osg-topology
```diff
@@ -151,6 +156,11 @@ def main():
     oparser = get_parser()
 
     args = oparser.parse_args()
+    if getattr(args, "fqdn") is not None:
+        if args.fqdn:
```
brianhlin: Aren't these two if statements superfluous? Can we ever reach the else block?

matyasselmeci: Yeah, we can reach the else block if args.fqdn is False, which is what I'm testing for.

bin/osg-notify
```diff
@@ -99,7 +99,11 @@ def parseargs():
                          help="Print out the email instead of sending it.")
 
     oparser.add_argument("--oim-name-filter", dest="name_filter", action="store", nargs="?",
-                         help="Shell expression filter on the VO or resource name")
+                         help="Shell expression filter on the VO or resource name."
+                         " Can't be specified along with --oim-fqdn-filter.")
+    oparser.add_argument("--oim-fqdn-filter", dest="fqdn_filter", action="store", nargs="?",
```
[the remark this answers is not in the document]
matyasselmeci: Sorry :P

PR #482, opened 2019-02-25
Update AGLT2_downtime.yaml

ShawnMcKee:
AGLT2 at the UM site will have a complete power outage on March 2, 2019.  We are reserving a window through March 3, 2019 at 15:00 Eastern time to recover all services and go back into production.


matyasselmeci:
I merged this, but you're not listed as a contact on AGLT2_SL6, only Wenjing Wu. Should we fix that?

matyasselmeci:
Looks like Brian took care of it in #483.

PR #494, opened 2019-03-04
Adding stashcache UCSD to topology

efajardo:


matyasselmeci:
Looks like you got the wrong Terrence: you put Terrence Martin in the name but the ID you put was for Terrence Lee Hart.

efajardo:
Thanks @matyasselmeci IT should be fixed now.

matyasselmeci:
Still failing. Looks like you'll need to use your full name "Edgar Mauricio Fajardo Hernandez" because that's what we have for you in the contact repo.

PR #505, opened 2019-03-07
Catch missing SITE.yaml and FACILITY.yaml

matyasselmeci:
and fix missing ones for our resources


brianhlin:
This looks good but how can we turn on strict checking for our tests? https://travis-ci.org/opensciencegrid/topology/builds/502137898#L618-L657

I suspect that we have to pass something to rg_reader.py

matyasselmeci:
strict is the default; we only turn it off for the webapp.

PR #511, opened 2019-03-08
Add downtime for UCSD for maintenance

jdost321:


brianhlin:
@jdost321 if you're going to be submitting downtimes for UCSD, we should just add you as an admin contact.

jdost321:
@brianhlin Go ahead. Can you add me from your side or is there something you need from me?

PR #522, opened 2019-03-26
Create UCIAtlas.yaml

djw8605:
Did I do this right?

### DIFF
--- a/projects/UCIAtlas.yaml
+++ b/projects/UCIAtlas.yaml
@@ -22,6 +22,6 @@ Sponsor:
   #   Name: <NAME>
 
   ### Uncomment the VirtualOrganization block if the project was sponsored by a VO
-  # VirtualOrganization:
+  VirtualOrganization:
   ##### Name is the name of the VO that sponsored the project
-  #   Name: <VO>
+    Name: ATLAS
--- a/projects/UCIAtlas.yaml
+++ b/projects/UCIAtlas.yaml
@@ -1,0 +1,27 @@
+# Name is the short name of the project. It must match the file name
+Name: UCIAtlas
+# Description is a human-readable description of what the project is about
+Description: University of California, Irvine Atlas Group
+# Department is the department of the project, e.g. "Physics"
+Department: Physics
+# FieldOfScience is a more specific description, e.g. "High Energy Physics"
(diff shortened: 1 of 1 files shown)


projects/UCIAtlas.yaml
```diff
@@ -0,0 +1,27 @@
...
+# FieldOfScience is a more specific description, e.g. "High Energy Physics"
+FieldOfScience: High Energy Physics
+# Organization is the organization where the project is hosted, e.g. "UCSD"
+Organization: UCI
+# PIName is the prinicipal investigator for the project
+PIName: Anyes Taffard
+
+# If you have an up-to-date local git clone, fill ID with the output from `bin/next_project_id`
+# Otherwise, leave it blank and we will fill in the appropriate value for you.
+ID:
```
matyasselmeci:
```suggestion
ID: 537
```

projects/UCIAtlas.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+Sponsor:
+  ### Uncomment the CampusGrid block if the project was sponsored by a campus grid e.g. OSG Connect
```
matyasselmeci: You need a sponsor here. Since it's atlas, perhaps
```
Sponsor:
  VirtualOrganization:
    Name: ATLAS
```

djw8605:
I don't understand what's wrong.  Any ideas?

projects/UCIAtlas.yaml
```diff
@@ -0,0 +1,27 @@
...
+  # CampusGrid:
+  ##### Name is the specific campus grid that sponsored the project
+  #   Name: <NAME>
+
+  ### Uncomment the VirtualOrganization block if the project was sponsored by a VO
+  VirtualOrganization:
+  ##### Name is the name of the VO that sponsored the project
+    Name: ATLAS
```
matyasselmeci:
```suggestion
    ID: 35
    Name: ATLAS
```
Looking at the other projects, it looks like you need the VO ID as well, not just the name. I think this is a bug.

djw8605: Makes sense in the DB world.  Linking things together...

PR #535, opened 2019-04-04
Make requirements.txt rootless

matyasselmeci:
requirements.txt becomes a symlink to requirements-rootless.txt.  The
old requirements.txt (used for the production service) becomes
requirements-apache.txt.

This is needed for Heroku because Heroku fails to compile mod_wsgi, and
we can't tell Heroku to skip it by using requirements-rootless.txt.


matyasselmeci:
We'll need to fix the deployment instructions at https://opensciencegrid.org/operations/services/topology/ if this is merged.

PR #537, opened 2019-04-04
Make /Gluex PUBLIC

matyasselmeci:
This is necessary to avoid errors due to caches not having DNs


matyasselmeci:
one sec, let me try it out locally before merging

matyasselmeci:
ok, works

matyasselmeci:
For the record, Richard Jones said in https://support.opensciencegrid.org/a/tickets/8513 that he doesn't have any protected data right now.

PR #545, opened 2019-04-10
Get rid of MaybeOrderedDict

matyasselmeci:
Turns out there's an existing type that does the same thing.


brianhlin:
Have you tried this out on the ITB?

matyasselmeci:
Yep.

PR #352, opened 2019-01-15
SciTokens configuration file support

bbockelm:
This is preliminary support for generating the SciTokens configuration file for caches.

Meant to be merged after #336.

PR #547, opened 2019-04-11
SciTokens config support

matyasselmeci:
#352 rebased on master with some code style changes and the webapp route added.

Added a SciTokens entry for GLOW and tested (locally) with the URL http://127.0.0.1:5000/stashcache/scitokens?cache_fqdn=sc-cache.chtc.wisc.edu to give the result:
```
[Global]
audience = CHTC_STASHCACHE_CACHE, https://sc-cache.chtc.wisc.edu

[Issuer /chtc/PROTECTED]
issuer = https://scitokens.org/chtc
base_path = /chtc
```

### DIFF
--- a/src/stashcache.py
+++ b/src/stashcache.py
@@ -168,7 +168,7 @@ def generate_cache_authfile(vo_data: VOsData, resource_groups: List[ResourceGrou
             else:
                 raise DataError("VO {} in StashCache does not provide a Namespaces list.".format(vo_name))
 
-        has_non_public = False
+        needs_authz = False
         for namespace, authz_list in namespaces.items():
             if not authz_list:
                 if suppress_errors:
@@ -176,9 +176,9 @@ def generate_cache_authfile(vo_data: VOsData, resource_groups: List[ResourceGrou
                 else:
                     raise DataError("Namespace {} (VO {}) does not provide any authorizations.".format(namespace, vo_name))
             if authz_list != ["PUBLIC"]:
-                has_non_public = True
+                needs_authz = True
                 break
-        if not has_non_public:
+        if not needs_authz:
             continue
 
         if resource and not _cache_is_allowed(resource, vo_name, stashcache_data, False, suppress_errors):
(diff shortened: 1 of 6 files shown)


matyasselmeci:
Should I add `bin/scitokens-config` script as well?

matyasselmeci:
Hold on a sec. The GLOW config change breaks origin-authfile generation for sc-origin ("Resource SU_STASHCACHE_CACHE is an allowed cache for VO GLOW but does not provide a DN.")

matyasselmeci:
SOFTWARE-3659 is done so this no longer causes a problem.

matyasselmeci:
This is ready for review...

src/stashcache.py
```diff
@@ -234,6 +241,109 @@ def generate_public_cache_authfile(vo_data, resource_groups, fqdn=None, legacy=T
...
+    for vo_name, vo_data in vo_data.vos.items():
+        stashcache_data = vo_data.get('DataFederations', {}).get('StashCache')
+        if not stashcache_data:
+            continue
+
+        namespaces = stashcache_data.get("Namespaces", {})
+        if not namespaces:
+            continue
+
+        has_non_public = False
```
bbockelm: Oof.  The `not has_non_public` double-negative later on is hard on the brain.

How about calling this `needs_authz`?  Then it's obvious what 

```
if not needs_authz:
    continue
```

is doing later on.

src/stashcache.py
```diff
@@ -234,6 +241,109 @@ def generate_public_cache_authfile(vo_data, resource_groups, fqdn=None, legacy=T
...
+                                                                 dirname, suppress_errors))
+
+    issuer_blocks_str = "\n".join(issuer_blocks)
+
+    return template.format(**locals()).rstrip() + "\n"
+
+
+def _get_scitokens_issuer_block(vo_name: str, scitokens: Dict, dirname: str, suppress_errors: bool) -> str:
+    template = """\
+[Issuer {dirname}]
```
bbockelm: Although I admit the issuer names are mostly arbitrary, using the directory name seems a bit strange.  I normally think of the issuer name as a human readable form of the `issuer` URL -- however, since there might need to be multiple issuer blocks per issuer, the `dirname` might be the best option.

Nothing to change, just noting something weird.

src/stashcache.py
matyasselmeci: Sure!

src/stashcache.py
matyasselmeci: blockname perhaps?

bbockelm: Nah, we don't need a change.  Just wanted to write down my thoughts for posterity.

matyasselmeci:
Deployed on topology.opensciencegrid.org.

matyasselmeci:
Resolves #352

PR #557, opened 2019-04-23
Update Caltech Normalization figures

djw8605:
...from CMS spreadsheet

I will wait for Caltech to review this change before merging.


juztas:
lgtm. I will review the list of all nodes and in case needed will update these values.

PR #574, opened 2019-05-01
reject PRs with new organizations (SOFTWARE-3671)

edquist:
Example rejection message:

https://github.com/opensciencegrid/topology/pull/573


bbockelm:
Suggested text:
```
Thank you for your pull request.

Your requested changes add a new organization to a project file; this requires manual review from OSG Staff.
```

edquist:
>I still hate your numeric return code thing

Yeah - an integer ends up being a convenient way to store the state of the check between the initial PR webhook and the CI status webhook - but i kind of hate it too.  I'd be open to redesigning things a bit, but for this PR it was easy enough just to extend the existing model.

edquist:
>The rejection message seems a little harsh

Fwiw, the current language as seen in #573 ("not allowed" and "please consider ...") was meant to explain the "requested changes", in contrast to a normal review comment, which also requires manual review from OSG Staff.

But I'm happy to use @bbockelm's text, or whatever.

matyasselmeci:
"additional manual review" perhaps?

The submitter sees this but IMO the comment is mostly intended for reviewers so they know they should check for this while reviewing the PR.

edquist:
How does this look now in #575 ?

matyasselmeci:
Looks good; if it's easy to omit the message about it not being a downtime commit, can you do that? If not, feel free to merge.

edquist:
>if it's easy to omit the message about it not being a downtime commit

Unfortunately, as it stands, the new-org rejection check is combined with the downtime automerge eligibility checks and it's not so easy to separate them.  Maybe later we can come up with a nicer way to show just what's "relevant".

But note also that, while the message about not being a downtime file in the above example is not so relevant, it does become a potentially useful message to keep there if there are both downtime and non-downtime files involved.

matyasselmeci:
I just deployed this on topology. Sorry I didn't do that sooner. Topology seems to still work.

edquist:
Erm, from SOFTWARE-3671,


>Carl Edquist added a comment - 2019/05/01 12:49 PM
>
>Updated on production topology instance.

I had updated it under /opt/topology-webhook, which is the instance that this matters for.

PR #578, opened 2019-05-02
use named constants for automerge check return codes

edquist:
to address some of the hate discussed in #574


edquist:
I did some tests on ITB; didn't seem to break anything too bad  :)

PR #590, opened 2019-05-07
Adding KISTI cache

efajardo:
I am setting a cache in KISTI for LIGO.  I am waiting on the hosts certs but getting this out of the way now.


brianhlin:
@efajardo what's the status on this cache?

efajardo:
@matyasselmeci can We merge this and I change the DN when we get the correct DN?

Thanks

PR #609, opened 2019-05-16
disable NERSC-PDSF resources

zvada:
Per discussion with @lagerhardt all resources here retired.


matyasselmeci:
Since you're disabling the resource group, you should just delete the whole file and the corresponding downtime.yaml file.

bbockelm:
Are we sure there are no GRACC records from the past 12 months for PDSF?

brianhlin:
Yup, last records I saw were from January 2017

bbockelm:
Sounds good.  In the future, we should record the fact we performed the check when we delete -- just in case we ever need to revisit our old decisions!

brianhlin:
@zvada please remove the resource group and corresponding downtime file instead

zvada:
> @zvada please remove the resource group and corresponding downtime file instead

sorry, slipped off of attention. @brianhlin should we add into an appropriate place in docs how to retire resource clean way? Also, what quick URL it is to check for the last records. Or, at a minimum, reflect the process in a note in the [development section](https://github.com/opensciencegrid/topology/tree/master/src) which explains `active|disable` or something...

matyasselmeci:
I fixed the merge conflicts and merged the changes in #632.

PR #610, opened 2019-05-17
Pixley lab

CarrieBrown:


edquist:
Hi @CarrieBrown,

It looks like the existing name for Rutgers University is `Rutgers, The State University of New Jersey` - if you could use the existing name instead.

CarrieBrown:
@edquist Thanks for pointing that out! I seem to have accidentally made two PRs. The PR with all the corrected info is #611 

I am not sure how to correct this.

PR #611, opened 2019-05-17
Adding PixleyLab and SimCenter projects

CarrieBrown:


edquist:
I don't think `University of Tennessee at Chattanooga` is the same as the main `University of Tennessee`, so this is fine.

brianhlin:
This resolves #610

PR #616, opened 2019-05-20
Adding missing XSEDE projects

rynge:


edquist:
Is `Johns Hopkins University Applied Physics Lab` a distinct organization from `Johns Hopkins University` ?

rynge:
I don't know - this information is mirrored from the XSEDE central database, so I would prefer to keep the information like they have it. If you feel that it will cause issue on our end, we can change it.

PR #617, opened 2019-05-21
Put UCSDT2-A in indefinite downtime

jdost321:


jdost321:
@brianhlin we will reinstall it eventually, Terrence just can't commit to a date yet so we put an artificially high year. Is that ok?

brianhlin:
That's fine, whenever you reinstall it you can set `Active: True`

brianhlin:
(note that we have this documented here: https://opensciencegrid.org/docs/common/registration/#registering-resource-downtimes)

topology/University of California San Diego/UCSD CMS Tier2/UCSDT2_downtime.yaml
```diff
@@ -620,3 +620,4 @@
   - CE
 # ---------------------------------------------------------
 
+
```
brianhlin: Looks like there's still an extra whitespace change that snuck in

```suggestion
```

brianhlin:
LGTM!

PR #618, opened 2019-05-21
Consolidate same FQDN (SOFTWARE-3373)

yzz971115:
topology/Harvard University/Harvard University/HU_ATLAS_Tier2.yaml
This group has several CE-service resources with same FQDN. So I just leave it as it is.


brianhlin:
I took a quick glance at the PR and there are some things that we have to clarify with resource owners before we merge some of these. For example, the Vanderbilt services on the same host have different `InteropAccounting` values. @PerilousApricot is that on purpose?

PerilousApricot:
We're not in the WLCG, so I don't think the Accounting matters to us (???), but I don't know enough about the meaning to understand the consequences

brianhlin:
@stlammel is`WLCGInformation/InteropAccounting` used by SiteDB? If so, does it make sense to disable it for some services on a single host but not others?

stlammel:
Hallo Brian,
   CMS does not use any information in the WLCGInformation tag.
   Thanks,
- Stephan

bbockelm:
The only place that I know `InteropAccounting` is used would be touch-points with WLCG.  Things to verify would be the GRACC-APEL uploader and (potentially) the uploads of XRootD data to MONiT.  For the latter, I would assume it is considered more "monitoring" than "accounting" -- but @djw8605 would be the authority.

djw8605:
If vanderbilt is not in the WLCG, then they do not care about the `InteropAccounting` and it should be `false`.  The `InteropAccounting` is used when deciding whether or not to send the usage information to the WLCG.

topology/Columbia University_Nevis Labs/Nevis/Nevis.yaml
```diff
@@ -82,40 +87,4 @@ Resources:
       StorageCapacityMax: 100
       StorageCapacityMin: 1
       TapeCapacity: 0
-  Nevis_gridftp:
```
brianhlin: Let's use `Nevis_gridftp` as the unified resource name instead

topology/Fermi National Accelerator Laboratory/FNAL USCMS Tier1/USCMS-FNAL-WC1.yaml
```diff
@@ -88,31 +88,6 @@ Resources:
...
-    Description: This is a PerfSonar dual stack instance
-    FQDN: psonartest1.fnal.gov
-    ID: 744
-    Services:
-      net.perfSONAR.Bandwidth:
-        Description: PerfSonar Bandwidth monitoring node
-        Details:
-          endpoint: psonartest1.fnal.gov
-    VOOwnership:
-      Fermilab: 100
```
brianhlin: @hyunwoo18 does this look ok to you? `USCMS PerfSonar IPV6 Test` appears to be the exact same resource as [FNAL NRTB PerfSONAR Bandwidth](https://github.com/opensciencegrid/topology/blob/1a6bae8e7f8bab0e713800f6ce4794604bcd0915/topology/Fermi%20National%20Accelerator%20Laboratory/FNAL%20USCMS%20Tier1/USCMS-FNAL-WC1.yaml#L5]

topology/Langston University/LUCCRE/LUCILLE.yaml
```diff
@@ -80,53 +90,6 @@ Resources:
       StorageCapacityMax: 110
       StorageCapacityMin: 110
       TapeCapacity: 0
-  Lucille_SE:
```
brianhlin: Let's merge the squid resource into the SE instead

topology/Langston University/LUCCRE/LUCILLE_downtime.yaml
```diff
@@ -9,17 +9,6 @@
   Severity: Outage
   StartTime: Jun 14, 2016 11:00 +0000
 # ----------------------------------------------------------
-- Class: SCHEDULED
```
brianhlin: The SE entries in this file will need to be restored and instead you'll want to remove the squid entries

topology/Princeton University/Princeton_ICSE/Princeton_ICSE_T3_CMS.yaml
```diff
@@ -42,35 +42,4 @@ Resources:
       KSI2KMin: 0
       StorageCapacityMax: 0
       StorageCapacityMin: 0
-  T3_US_Princeton_ICSE:
```
brianhlin: Let's keep this resource and remove `Princeton_ICSE_T3_CMS_SE` instead

topology/Stanford University/SLAC/WT2.yaml
```diff
@@ -132,46 +132,6 @@ Resources:
...
-      APELNormalFactor: 0
-      AccountingName: US-WT2
-      HEPSPEC: 0
-      InteropAccounting: true
-      InteropBDII: true
-      InteropMonitoring: true
-      KSI2KMax: 0
-      KSI2KMin: 0
-      StorageCapacityMax: 3890
-      StorageCapacityMin: 3890
```
brianhlin: @wyang007 this resource appears to be a duplicate of `SLAC-ATLAS-T3_GRIDFTP` is there a good reason for that? If not, do you have a preference for which one we remove?

topology/University of Nebraska/Nebraska-CMS/Nebraska.yaml
```diff
@@ -303,35 +307,6 @@ Resources:
...
-    Description: perfSONAR testbed latency host at the University of Nebraska-Lincoln.
-    FQDN: perfsonar.unl.edu
-    ID: 692
-    Services:
-      net.perfSONAR.Latency:
-        Description: PerfSonar Latency monitoring node
-        Details:
-          endpoint: perfsonar.unl.edu
-    VOOwnership:
-      HCC: 100
```
brianhlin: @gattebury is there a good reason why there's separate resource entries for the bandwidth and latency monitors? Can we just merge the two since they're on the same host?

hyunwoo18:
Brian,
I am not the right person about this perfSONAR resource.
I vaguely remember who were in charge of that, but they already left
Fermilab.
So, Tanya Levshina (tlevshin@fnal.gov) might have the best chance of
knowing what to say about that.
HyunWoo

topology/University of Nebraska/Nebraska-CMS/Nebraska.yaml
gattebury: There's no good reason I can think of. That host was originally in a pS development test mesh and there's a chance having separate entries made sense to someone, but that's in the past. Feel free to squash it down to a single entry.

brianhlin: Will do, thanks for the quick response!

topology/Stanford University/SLAC/WT2.yaml
brianhlin: @wyang007 any thoughts on this duplicate resource?

wyang007:
Hi Brian, WT2 and SLAC-ATLAS-T3 always share the same GridFTP.

topology/Fermi National Accelerator Laboratory/FNAL USCMS Tier1/USCMS-FNAL-WC1.yaml
edquist: Looks ok to me to drop `USCMS PerfSonar IPV6 Test`.

As you mentioned it's identical to `FNAL NRTB PerfSONAR Bandwidth` except that Burt Holzman was listed as a Miscellaneous Contact, and of course the resources had different IDs.

brianhlin: Agreed, let's drop the test one

topology/Stanford University/SLAC/WT2.yaml
edquist: @brianhlin, as @wyang007 noted,

> WT2 and SLAC-ATLAS-T3 always share the same GridFTP.

The actual resource names that share the duplicate FQDN `osggridftp02.slac.stanford.edu` are `WT2_SE` and `SLAC-ATLAS-T3_GRIDFTP`.

Anyway, the resources are identical except for the Description and WLCGInformation.  The relevant fields in the diff:

```
-WT2_SE:
+SLAC-ATLAS-T3_GRIDFTP:
-  Description: SE at SLAC
+  Description: GridFTP server for SLAC ATLAS Tier 3 and others
   FQDN: osggridftp02.slac.stanford.edu
-  ID: 166
+  ID: 491
   ...
   WLCGInformation:
     APELNormalFactor: 0
-    AccountingName: US-WT2
     HEPSPEC: 0
-    InteropAccounting: true
+    InteropAccounting: false
     InteropBDII: true
-    InteropMonitoring: true
+    InteropMonitoring: false
     KSI2KMax: 0
     KSI2KMin: 0
-    StorageCapacityMax: 3890
-    StorageCapacityMin: 3890
+    StorageCapacityMax: 100
+    StorageCapacityMin: 5
+    TapeCapacity: 0
```

So the question is, if we want to combine these, is it OK to drop the `WT2_SE` resource name?  (Or do they both need to exist?)  And then, should the WLCGInformation be combined, just using the values from `WT2_SE`, which has InteropAccounting enabled?

brianhlin: Let's keep both of those separate for now.

Though since it's the same host, I would think the WLCG info should be the same (@wyang007?) but that's not important to get into this PR

edquist: Alright, I've added back `WT2_SE` to match what's in `master`.

edquist:
I added back in the LUCILLE downtimes -- we are merging the resources rather than retiring them, so the downtimes can stay, but i've updated the ResourceName from Lucille-squid to Lucille_SE.

edquist:
So, looks like the current set of duplicate FQDNs remaining is:

```
      2     FQDN: osg-comet-frontend.sdsc.edu
      2     FQDN: osggridftp02.slac.stanford.edu
      2     FQDN: ps-development.bnl.gov
      2     FQDN: psuta01.atlas-swt2.org
      2     FQDN: psuta02.atlas-swt2.org
      2     FQDN: rccf-osg.ci-connect.net
      2     FQDN: t3serv007.mit.edu
      3     FQDN: net2.rc.fas.harvard.edu
```

PR #641, opened 2019-05-29
add FACILTY and SITE check (SOFTWARE-3425)

yzz971115:


src/tests/verify_resources.py
```diff
@@ -436,6 +439,26 @@ def test_14_res_contacts_match(rgs, rgfns, contacts):
 
     return errors
 
+def test_15_FACILITY_SITE_check():
+    # verify the required FACILITY.yaml and SITE.yaml files
+    errors = 0
+
+    groups = glob.glob(_topdir + '/topology/*')
```
edquist: Note that we do a `os.chdir(_topdir + "/topology")` earlier, so all `glob()` calls can start relative to the `topology` dir, rather than needing to specify `_topdir` and `/topology/` as part of the glob path.

(In fact, including `_topdir`, which the program does not control, in the argument to `glob()` is technically unsafe, as the path could contain glob characters ...)

src/tests/verify_resources.py
```diff
@@ -436,6 +439,26 @@ def test_14_res_contacts_match(rgs, rgfns, contacts):
 
     return errors
 
+def test_15_FACILITY_SITE_check():
```
edquist: If you notice the rest of the `test_*` function names, none of them ends with "check" or similar - they are all tests.  You can drop the `_check` suffix.  If you want to be more specific you can call it `test_15_facility_site_files`, or make two separate tests (one for facility and one for site), or even just work these tests for `SITE.yaml` and `FACILITY.yaml` into `test_6_site`.

src/tests/verify_resources.py
```diff
@@ -436,6 +439,26 @@ def test_14_res_contacts_match(rgs, rgfns, contacts):
...
+    # verify the required FACILITY.yaml and SITE.yaml files
+    errors = 0
+
+    groups = glob.glob(_topdir + '/topology/*')
+    for group in groups:
+        for root, dirs, files in os.walk(group):
+            if root == group:
+                if 'FACILITY.yaml' not in files:
```
edquist: I'm not wild about using `os.walk` for this and then testing membership.

Can we just to a `for facdir in glob.glob("*/")` and test `os.path.exists(facdir + "FACILITY.yaml")`?

And likewise `for sitedir in glob.glob("*/*/")` and test `os.path.exists(sitedir + "SITE.yaml")` ?

yzz971115: Thanks, I will make the modification.

PR #644, opened 2019-05-30
Create FIFE.yaml

shreyb:
The one thing I couldn't put in here is a project ID, so I left it as an empty quoted string - I assume OSG assigns that?

We're creating this project so we can associate a new CVMFS repo with it.

### DIFF
--- a/projects/FIFE.yaml
+++ b/projects/FIFE.yaml
@@ -1,0 +1,11 @@
+Department: N/A
+Description: Project entry corresponding to FIFE experiments within the Fermilab VO.
+FieldOfScience: High Energy Physics
+ID: ''
+Name: FIFE
+Organization: Fermilab
+PIName: Ken Herner
+Sponsor:
+  VirtualOrganization:
+    ID: '9'
+    Name: Fermilab
--- a/projects/FIFE.yaml
+++ b/projects/FIFE.yaml
@@ -1,7 +1,7 @@
 Department: N/A
 Description: Project entry corresponding to FIFE experiments within the Fermilab VO.
 FieldOfScience: High Energy Physics
-ID: ''
+ID: '564'
 Name: FIFE
 Organization: Fermilab
 PIName: Ken Herner


brianhlin:
Hey @shreyb, what exactly do you mean by a new FIFE CVMFS repo? Something like `/cvmfs/fife.fnal.gov/`? I wasn't aware of projects having anything to do with CVMFS repos.

djw8605:
CVMFS repos are associated with VO's, not projects.  See the Nova VO for example: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/Nova.yaml#L48

shreyb:
FIFE is a part of the Fermilab VO, and thus should be a project.  I just checked again with Dave Dykstra, and we believe the policy has been that repo names have to be derived from either projects or VOs, which is why we're requesting to register FIFE a project.  Is that not right?

Our plan is to create multiple fife CVMFS repositories for a new service we're standing up here.  The repos will be something like /cvmfs/fife-user1.opensciencegrid.org; /cvmfs/fife-user2.opensciencegrid.org, etc.

brianhlin:
I think that would just require adding entries under `OASISRepoURLs` here: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/Fermilab.yaml#L61. Projects are really just for accounting for smaller groups within a VO.

shreyb:
But to have the names on the repos, doesn't this documentation say that we have to at least have it be a project?  We're thinking this would be similar to how Larsoft, another fermilab project, is set up.

https://opensciencegrid.org/docs/data/external-oasis-repos/#creating-a-repository

brianhlin:
Do you mean this?

> Select a repository name; typically, this is derived from the VO or project's name

That's really just a suggestion for choosing a name and doesn't read like a requirement to me. @DrDaveD can correct me if I'm wrong. The key portion is requesting that the repo be added to the VO it's serving, which would then be added to that VO's list of `OASISRepoURLs`

But going back to your previous comment:

> FIFE is a part of the Fermilab VO, and thus should be a project.

This is only true if FIFE jobs specify `+Project = FIFE` in their submit file so that you can separate out FIFE accounting records: https://gracc.opensciencegrid.org/dashboard/db/project-summary?orgId=1

DrDaveD:
> That's really just a suggestion for choosing a name and doesn't read like a requirement to me. 

It's true that I softened the language at some point, but originally it was a requirement and I probably weakened the language on my own initiative because I thought the people who originally pushed for it to be a requirement were no longer involved.  Still, I do think it makes sense to register the project name even if no jobs are submitted with that name, just to reserve the prefix for cvmfs repositories.  I don't think Larsoft submits any jobs either, but we had them register as a project in order to get a cvmfs repository.

shreyb:
So, what are people's thoughts? Should we make this a project given what Dave said?

brianhlin:
I don't really have any objection to it as long as everyone understands that this has nothing to do with CVMFS repositories. Merging this doesn't reserve any CVMFS repo namespaces since we don't check project names or VOs when adding to a VO's `OASISRepoURLs`

projects/FIFE.yaml
```diff
@@ -0,0 +1,11 @@
+Department: N/A
+Description: Project entry corresponding to FIFE experiments within the Fermilab VO.
+FieldOfScience: High Energy Physics
+ID: ''
```
brianhlin: @shreyb before we merge this into master, could you approve this suggestion?

```suggestion
ID: '564'
```

DrDaveD:
The [instructions for the operations person](https://opensciencegrid.org/operations/external-oasis-repos/#requests-to-host-a-repository-on-oasis) does say that they're supposed to make sure that the name is "appropriate" for the VO.  We could make that instruction more explicit to ask that they check that the name is derived from the VO name or one of its projects.  Otherwise at least if somebody else tries to register a project by the same name they'll find a clash.

brianhlin:
> We could make that instruction more explicit to ask that they check that the name is derived from the VO name or one of its projects. 

Yeah, I think that'd be appropriate. I've created a ticket: https://opensciencegrid.atlassian.net/browse/OPS-5

projects/FIFE.yaml
shreyb: Yep - done.

shreyb:
Thanks all!

PR #656, opened 2019-06-04
Fix spelling of "University of Wisconsin-Madison"

matyasselmeci:


brianhlin:
Maybe we should just remove this entirely? I don't think we use projects anymore in the user school (@lmichael107, @ChristinaLK, @osg-cat)

osg-cat:
Last year (2018), I think we had special, temporary training accounts on a separate server … but that would not preclude having a group, right? Also, the correct formatting for our institution is University of Wisconsin–Madison (en-dash).

brianhlin:
Having a registered project would just be useful for GRACC purposes and would require additional attributes at submit time, i.e. having `ProjectName = UserSchool2015` set in their job ads. There isn't really any other purpose to it.

Though if we end up merging GLOW/OSG VOs or just requiring projects to GLOW user job submissions, then we should rename this to `UserSchool` or something along those lines.

matyasselmeci:
Can we make a decision on this?

brianhlin:
I vote that we just get rid of it entirely

lmichael107:
I agree with Brian. The OSG School usage falls under GLOW, which we divvy by project for the annual report. With my general OSG hat on (and thinking of FoCaS services for institutions), we should aim for users to NOT have to indicate a project (unless they have a second project they'd like to indicate), which means we need a way for GLOW and others to have a default project inserted at submit time. We don't have that yet, so I think we can keep GLOW in its current situation, until then.

matyasselmeci:
Closing, see #684 instead.

PR #659, opened 2019-06-05
Adding WayneStateU_TDA group, per FreshDesk ticket 8725

lmichael107:


projects/WayneStateU_TDA.yaml
```diff
@@ -0,0 +1,13 @@
+Name: WayneStateU_TDA
+Description: Topological Data Analysis of fMRI Signals during Learning: Function to Structure
```
matyasselmeci:
```suggestion
Description: "Topological Data Analysis of fMRI Signals during Learning: Function to Structure"
```
Looks like this needs to be quoted because it has a colon in it.

PR #665, opened 2019-06-10
Update UColorado_HEP.yaml

doug1977:
Attempting to register multiple squid servers

### DIFF
--- a/topology/University of Colorado/UColorado_HEP/UColorado_HEP.yaml
+++ b/topology/University of Colorado/UColorado_HEP/UColorado_HEP.yaml
@@ -34,6 +34,27 @@ Resources:
       GridFtp:
         Description: GridFtp Storage Element
         Details:
+          hidden: false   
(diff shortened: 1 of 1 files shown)


brianhlin:
Hi Doug, do these squids serve all the same worker nodes? Because we suggest having a single DNS round-robin entry in that case (https://opensciencegrid.org/docs/data/frontier-squid/#registering-frontier-squid). If not, I'll add a few suggestions that you'll need to accept so that we can merge this in.

Also, do you have a backup contact for the resources?

matyasselmeci:
Hi Doug,
We just need to know if there's a round-robin DNS entry for these squids. If there isn't, we register the individual squids; if there is, we register that DNS entry instead.
Either way, I can take care of the syntax.

doug1977:
Greetings,
No there is not a DNS round robin.
Thank you,
doug

matyasselmeci:
Ah, Brian just told me that you weren't doing round-robin.

PR #667, opened 2019-06-10
Add check for support center and VO contacts (SOFTWARE-3329)

yzz971115:
For this commit, there exist a problem that some of the error messages for unknown ID will be displayed in the wrong error message category -- name mismatch, not all of them, just a portion. I do not know how to fix this problem by modifying my code. I need some suggestions.

### DIFF
--- a/src/tests/verify_resources.py
+++ b/src/tests/verify_resources.py
@@ -44,7 +44,7 @@ def sumvals(d):
 
 def get_vo_names():
     return set( re.search(r'/([^/]+)\.yaml$', path).group(1) for path in
-                glob.glob(_topdir + "/virtual-organizations/*.yaml") )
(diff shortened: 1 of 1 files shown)


src/tests/verify_resources.py
```diff
@@ -44,7 +44,7 @@ def sumvals(d):
 
 def get_vo_names():
     return set( re.search(r'/([^/]+)\.yaml$', path).group(1) for path in
-                glob.glob(_topdir + "/virtual-organizations/*.yaml") )
+                glob.glob(_topdir + "/virtual-organizations/*.yaml"))
```
edquist: This space was intentional, as part of the generator expression started with `set( `

edquist:
I'll have to take a closer look at the main body of changes for this PR to study what's going on.

PR #669, opened 2019-06-11
Update SyracuseCachingInfrastructure.yaml

rcameele:
Added DN


brianhlin:
Hey @rcameele, has that cert been issued by the InCommon IGTF CA? I noticed that it has `postalCode` and `street` fields, which I believe the IGTF CA has stopped including recently.

rcameele:
Yes it was, but it’s only valid until June 28 so I will be requesting a new one shortly.

Rich

PR #672, opened 2019-06-11
clarify language in README about REPORTING_GROUPS.yaml

edquist:
"each VO has a file" sounded like there is a separate file for each VO.

Also, why are reporting groups defined in a separate file anyway?  (As opposed to the `<VO>.yaml` files.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ as follows:
 -   `virtual-organizations/` contains information about Virtual Organizations
     (VOs).
     Each VO has its own file, named `<VO>.yaml`.
-    Additionally, each VO has a file that contains information about WLCG metric reporting groups called
-    `REPORTING_GROUPS.yaml`.
+    Additionally, there is a `REPORTING_GROUPS.yaml` file that contains information about WLCG metric reporting groups
+    for each VO.
 
 -   `topology/` contains information about the topology of the resources that OSG sites provide.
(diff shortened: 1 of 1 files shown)


matyasselmeci:
The reporting groups don't match up 1:1 with the VOs which is why they're in a separate file. Not every VO will have a reporting group (and it's possible that multiple VOs will report to the same group, though I didn't look). This is all stuff I imported from the database.

edquist:
Thanks @matyasselmeci for the explanation.

So, is the structure of `REPORTING_GROUPS.yaml` keyed off of VO name at the top level?  And then the reporting groups are under `FQANs->GroupName` ?

Because if that's the case, i don't see why we can't just include a `ReportingGroups` field in `<VO>.yaml` -- even if they are not 1:1 and even if multiple VOs report to the same group.

(Thoughts?)

matyasselmeci:
I don't have the conversion script any more but $20 says they are keyed off of the "name" field in the "vo_report_name" table, and there is no foreign key constraint linking that field to the "vo" table. So no.

Also I don't want to mess with the format unless we absolutely have to.

edquist:
> $20

That's like 3.5 beers!

Alright.  Let's just hope nobody wants to add a VO called `REPORTING_GROUPS`  :-p

matyasselmeci:
> Let's just hope nobody wants to add a VO

Fixed that for you :-P

PR #673, opened 2019-06-12
Update Vanderbilt resources

PerilousApricot:
We are temporarily having to move all our external grid hosts to new
domain names. Update the Vanderbilt resource group to take that into
account


brianhlin:
Resolved by #674

PR #677, opened 2019-06-12
Add test for registered CEs and reporting CEs (SOFTWARE-3370)

yzz971115:


djw8605:
I don't think this should be part of the topology code base, nor should it directly read in the topology URL.  It should be separate, and query topology through the standard HTTP web interface.

brianhlin:
> I don't think this should be part of the topology code base, nor should it directly read in the topology URL. It should be separate, and query topology through the standard HTTP web interface.

This is a good point, back to the drawing board

PR #680, opened 2019-06-13
Adding LancasterPPS.yaml to correct MuonG2.yaml

lmichael107:


lmichael107:
Fixes #540, which can be closed.

PR #682, opened 2019-06-14
Adding stashcache-origin-01.ultralight.org into CIT infrastructure

andres-moya:
Hi, 

I am doing it my first time. Please check i am doing it right. 

Not sure multiple Secondary allowed in ContactLists

PR #683, opened 2019-06-17
Add DN for stashcache-origin-01.ultralight.org

matyasselmeci:


matyasselmeci:
This was part of #682.

PR #686, opened 2019-06-17
Fix perms for icecube namespace

matyasselmeci:


brianhlin:
This seems to be the opposite of what we tell users to do via https://opensciencegrid.org/docs/data/stashcache/vo-data/#namespaces-section

matyasselmeci:
Apparently the doc is wrong.

matyasselmeci:
https://github.com/opensciencegrid/docs/pull/576

PR #694, opened 2019-06-21
change the notify script and add site contact (software-3539)

yzz971115:
The updated tool has not been tested due to limit of authority


brianhlin:
@yzz971115 great, this fails just as expected! Could you submit the pull request to add the relevant contacts to our database?

PR #700, opened 2019-06-24
Added missing XSEDE projects

rynge:


projects/TG-DMR190045.yaml
```diff
@@ -0,0 +1,14 @@
+Name: TG-DMR190045
+Description: 1D nanoconfined helium: A versatile platform for exploring Luttinger liquid physics
```
matyasselmeci:
```suggestion
Description: "1D nanoconfined helium: A versatile platform for exploring Luttinger liquid physics"
```
Looks like you'll need to quote text with colons in it.

PR #711, opened 2019-06-30
Update SPRACE_downtime.yaml

jmarra13:
Please update the downtime for SPRACE, it was turned off because of 2 out of 3 air conditioners are not working well and temperature is constantly raising.


jmarra13:
i doing another one

PR #722, opened 2019-07-03
place the DN entry for osg-gftp under osg-gftp FQDN instead of under …

amcneill3:
…osg-sched FQDN

place the DN entry for osg-gftp under osg-gftp FQDN instead of under osg-sched FQDN


matyasselmeci:
I didn't even notice that. Thanks!

PR #723, opened 2019-07-03
modify the osg-notify tool to send one email at one time (software-3390)

yzz971115:


bin/osg-notify
```diff
@@ -185,11 +186,13 @@ def main():
...
-    msg = generateFullMessage(subject=args.subject, to=recipients, from_name=args.from_name, message=contents,
-                              sign=args.sign, keyid=args.keyid)
+    msgs = []
+    for recipient in recipients:
+        msgs.append(generateFullMessage(subject=args.subject, to=recipient, from_name=args.from_name, message=contents,
+                              sign=args.sign, keyid=args.keyid))
 
     if args.dryrun:
-        print(msg)
+        print('\n'.join(msg for msg in msgs))
```
matyasselmeci: This is gonna print out a huge amount of output. I suggest doing the following instead:
```
if args.dryrun:
    if not recipients:
        raise Exception("Message has no recipients!")
    print("Recipients: " + ", ".join(recipients))
    print("Example message:")
    print(msg[0])
```

and making the same change below (after the "Not sending email...")

bin/osg-notify
```diff
@@ -65,13 +66,13 @@ def generateFullMessage(subject, to, from_name, message, sign=True, keyid=None):
         msg.attach(basemsg)
         msg.attach(signmsg)
         msg['Subject'] = subject
-        msg['Bcc'] = ", ".join(to)
+        msg['Bcc'] = to
```
matyasselmeci: Since we're now only sending to one person at a time, use the "To" field instead of the "Bcc" field for the recipient.

edquist: @matyasselmeci, the current "To" is set to `"Open Science Grid Helpdesk" <help@opensciencegrid.org>` - do you think that should be kept also for To, or should the real recipient replace it?

matyasselmeci: No, only use the real recipient.

bin/osg-notify
yzz971115: ok, thanks, I will modify it

bin/osg-notify
```diff
@@ -199,12 +205,16 @@ def main():
 
             if verify_send.lower() == 'y':
                 session = smtplib.SMTP('localhost')
-                session.sendmail("help@opensciencegrid.org", recipients, msg)
+                for recipient, msg in zip(recipients, msgs):
+                    session.sendmail("help@opensciencegrid.org", recipient, msg)
+                    time.sleep(4)
```
matyasselmeci: Put a `print(".", end="")` in the for loop to let the user know that something is happening.

PR #727, opened 2019-07-09
Revert SOFTWARE-3390 to avoid sending multiple announcements

brianhlin:
This reverts the following commits:

- f36187e113323fc76c3395658036dadd239ca871
- def892ece575c6709e2ca5c683810ed7d750cd3a
- 6b999618992b32c0bfd5cc23b2969ab385131ddc.
- 64d7394bd563e782b4999f3b652a81b0651a3776.

### DIFF
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -11,7 +11,6 @@ import sys
 import getpass
 import smtplib
 import argparse
-import time
 import email.message
 import email.mime.text
 import email.mime.multipart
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Por qué?

brianhlin:
This came up in a discussion of mailing lists with the Tims: we send announcements to a combination of topology contacts and mailing lists. If we send separate emails for each recipient, someone registered in topology would get one email + an additional email for each mailing list they're subscribed to.

matyasselmeci:
Gotcha. We should put that note into SOFTWARE-3390 and change the resolution from 'done' to 'won't do'.

brianhlin:
Done

PR #728, opened 2019-07-10
Adding Comet Virtual CLuster info

efajardo:


efajardo:
@matyasselmeci ok I updated the name. @brianhlin thanks I would prefer to add it to the gitignore

PR #736, opened 2019-07-12
vo part of software-3345

yzz971115:


src/webapp/project_reader.py
```diff
@@ -23,6 +25,7 @@ def get_projects(indir="../projects", strict=False):
     projects = []
 
     mapping = load_yaml_file(os.path.join(indir, "_CAMPUS_GRIDS.yaml"))
+    VOsData = get_vos_data("../virtual-organizations", None)
```
matyasselmeci:
```suggestion
    vos_data = get_vos_data(os.path.join(indir, "../virtual-organizations"), None)
```

src/webapp/project_reader.py
```diff
@@ -38,6 +41,10 @@ def get_projects(indir="../projects", strict=False):
                 name = data['Sponsor']['CampusGrid']['Name']
                 ID = mapping[name]
                 data['Sponsor']['CampusGrid'] = OrderedDict([("ID", ID), ("Name", name)])
+            elif 'VirtualOrganization' in data['Sponsor']:
+                name = data['Sponsor']['VirtualOrganization']['Name']
+                ID = VOsData.vos[name]['ID']
```
matyasselmeci:
```suggestion
                ID = vos_data.vos[name]['ID']
```

matyasselmeci:
Also, Python convention is to only capitalize constants and class names, so local variables should be all lowercase.

brianhlin:
@yzz971115 for further reading about Python conventions, check out PEP8:

- https://www.python.org/dev/peps/pep-0008/#constants
- https://www.python.org/dev/peps/pep-0008/#class-names
- https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

matyasselmeci:
This looks good, but I'm going to hold off on merging it until I get back from vacation on Thursday.

brianhlin:
Labeling `do-not-merge` for the time being

PR #742, opened 2019-07-12
adding chemml.yaml per FreshDesk 6892, 12048

lmichael107:


matyasselmeci:
Hi Lauren,
We dropped the requirement for putting in the ID for Campus Grids (the `ID: '14'` right above the `Name: OSG Connect`), so you won't have to put it in for future PRs.
(You'll still need the main ID for the project.)

PR #751, opened 2019-07-16
Add SDSC-PRP-GRIDFTP-IceCube

sfiligoi:


topology/University of California San Diego/SDSC OSG/SDSC-PRP.yaml
```diff
@@ -39,4 +39,29 @@ Resources:
...
+          Name: Igor Sfiligoi
+          ID: 593daf4f680b3135849201036e1055f03f89cfac
+    FQDN: xcache-11.t2.ucsd.edu
+    Services:
+      GridFtp:
+        Description: GridFtp Storage Element
+        Details:
+          hidden: false
+    VOOwnership:
+      icecube: 100
```
brianhlin:
```suggestion
      IceCube: 100
```

bbockelm:
WAIT - NACK NACK.

Endpoints correspond to resource names (this was accidentally not enforced and we are trying to fix some existing examples).

We need to add another service type to the existing endpoint instead.

sfiligoi:
Not sure what do you mean by this.
The fact that I have two services on the same node?

bbockelm:
Right - each resource (the top-level YAML object inside `Resources`) should have a _unique_ FQDN.  So, instead of adding a second resource with a duplicate FQDN (`xcache-11.t2.ucsd.edu`), the existing resource should have a second service running on it.  Example:

```
    Services:
      CE:
        Description: Compute Element
        Details:
          hidden: false
          sam_uri: htcondor://xcache-11.t2.ucsd.edu
```

becomes

```
    Services:
      CE:
        Description: Compute Element
        Details:
          hidden: false
          sam_uri: htcondor://xcache-11.t2.ucsd.edu
      GridFtp:
        Description: GridFtp Storage Element
        Details:
          hidden: false
```

sfiligoi:
Ugh... that's really ugly, since I am forced to use the same logical name for it, which has no resemblance to its actual semantics!

PR #757, opened 2019-07-18
Move gridftp into the CE resource

sfiligoi:
As requested by blin.


brianhlin:
It doesn't look like we actually added the GridFtp service to the CE resource, though

matyasselmeci:
?

brianhlin:
Oops, misread the diff. Carry on.

PR #767, opened 2019-07-20
Support mailing OASIS managers with osg-notify (SOFTWARE-3757)

yzz971115:


matyasselmeci:
This tool needs to query topology.opensciencegrid.org instead of reading the data directly.

brianhlin:
Back to the drawing board on this one

PR #770, opened 2019-07-24
 Rename SDCC entry for GRACC readability.

jhover:
Looking to clean up the view in GRACC, and refer to BNL in the name.


efajardo:
This looks fine. I would double check with factory Ops so we remain as consistent as possible.

brianhlin:
@jdost321 @mmascher @zvada this PR changes the topology site name for the BNL-SDCC resource group. Does anything need to be changed in the factory for consistency?

PR #642, opened 2019-05-29
add test to check required DN for Xrootd service (SOFTWARE-3667)

yzz971115:
Some resources do not have required DN now.

PR #774, opened 2019-07-26
Add test to verify that XRootD caches have a DN attribute (SOFTWARE-3667)

brianhlin:
Additional commits on that ignores inactive caches, fixing #642


matyasselmeci:
Looks like it didn't catch one of them -- I see this in the warnings:
> `# WARNING: Resource UCSD_STASHCACHE_CACHE was skipped for VO HCC because the resource does not provide a DN.`

brianhlin:
That's a separate test. I think you actually added that warning text to the authfile generation yourself :)

matyasselmeci:
In https://github.com/opensciencegrid/docker-osg-build/blob/master/input/build-from-github#L10, you don't specify the mock config to use, so it uses the default mock config, which appears to be epel-7. So you don't have the osg repos available in the mock chroot. I'd suggest changing build-from-github to add `--mock-config-from-koji=osg-3.4-el7-build`. Read-only koji access should be sufficient for that, but you'll still have to run `osg-build setup` to put the right things into `~/.osg-koji`.

matyasselmeci:
(We talked about this in person; it's OK for that one not to have a DN because it's inactive.)

PR #779, opened 2019-07-31
Add HCC as a LIGO-recognized origin

brianhlin:
@efajardo said that the HCC origin is responsible for hosting LIGO's private data so I think it should be registered as such. Are there any reason that we shouldn't?


djw8605:
Although it's reasonable to have a Ligo HCC Origin, this particular host is not the ligo origin (we have a few, I guess).

The real origin is our CMS Tier2, behind xrootd-local.unl.edu.  We will have to register that origin.

PR #790, opened 2019-08-08
Georgia Tech Quarterly Maintenance Downtime

mbatgt:
Systems will be offline between 8/8 (Thr) and 8/10 (Sat) for scheduled quarterly maintenance


mbatgt:
Sorry for the last minute notice! Also a question... Do we need to keep the entire history of old downtimes in this file, or can we simply replace the content with the next one? Thanks!

edquist:
Hi @mbatgt,

>Do we need to keep the entire history of old downtimes in this file, or can we simply replace the content with the next one?

The idea is just to append new ones, rather than replacing old ones.  In theory we only get rid of downtime information when a resource is decommissioned and removed altogether from topology.

PR #791, opened 2019-08-08
Register Squid for brux11.hep.brown.edu (FD #26994)

matyasselmeci:


matyasselmeci:
Edgar did this in another PR.

PR #807, opened 2019-08-15
Adding AMNH_Smith and Venda_Arrey

lmichael107:


brianhlin:
@lmichael107 why a separate project from https://github.com/opensciencegrid/topology/blob/master/projects/AMNH.yaml?

This seems to be going down the project hierarchy route that @osg-cat, Frank, and @rynge were discussing in the "projects in OSG" doc that we have going. I'd like to sort out naming schemes sooner rather than later.

lmichael107:
My understanding is that, across OSG (in general) we definitely want better granularity in the "Project" designation than having only one project per institution. In OSG Connect, alone, WE (Research Facilitation) definitely want more granularity than one project per institution.

"AMNH" probably should have been "AMNH_Staff", as is our convention for other institutions, but it was created a bit hastily and without a clear vision for how/when they'd pursue their own submit server.

brianhlin:
That makes sense but why do we need the institution in the namespace if we specify that in the yaml with the `Organization` field?

Also, what's the difference between a project associated with OSG Connect and the OSG VO? Is that a useful distinction that needs to be in the registration? In other words, could we just gather this data from the accounting records based on which submit hosts the jobs are coming from?

lmichael107:
Hi Brian,

We decided (in OSG Connect support) that letting researchers choose their
project names led to numerous issues, and that going with
Institution_PILastName was far more informative than any other convention
(or any of the project names that have been chosen by researchers), even if
redundant with some other details. UNL and UW-Madison use similar
conventions for project names, because it makes it easy to see whom a
report is referring to, at first glance (both for the researcher, and for
OSG staff).

Lauren

Lauren Michael - Research Computing Facilitator, Center for High Throughput
Computing <http://chtc.cs.wisc.edu/>, University of Wisconsin - Madison
Other Affiliations: Wisconsin Institute for Discovery
<http://wid.wisc.edu/people/lauren-michael/>; Open Science Grid
<https://opensciencegrid.org/>; CaRCC <https://carcc.org/>
lmichael@wisc.edu, tinyurl.com/LMichaelCalendar, Discovery 2262,
(608)316-4430

brianhlin:
What if a PI has multiple projects?

Also, any thoughts about the OSG Connect vs OSG VO distinction above?

lmichael107:
They shouldn't need multiple projects. One is sufficient granularity.

Lauren

Lauren Michael - Research Computing Facilitator, Center for High Throughput
Computing <http://chtc.cs.wisc.edu/>, University of Wisconsin - Madison
Other Affiliations: Wisconsin Institute for Discovery
<http://wid.wisc.edu/people/lauren-michael/>; Open Science Grid
<https://opensciencegrid.org/>; CaRCC <https://carcc.org/>
lmichael@wisc.edu, tinyurl.com/LMichaelCalendar, Discovery 2262,
(608)316-4430

brianhlin:
Are there any multi-disciplinary PIs that may have projects with different fields of science?

How about OSG Connect vs OSG VO?

osg-cat:
Brian, let’s not try to solve every possible corner case with the project identifier and field-of-science tagging. Close counts.

And isn’t it the case that every user on OSG-Connect is a member of the OSG VO? In other words, we are talking about the organization of projects within the OSG VO. When we move GLOW users to the OSG VO, we will use the same conventions.

— Tim

brianhlin:
> Brian, let’s not try to solve every possible corner case with the project identifier and field-of-science tagging. Close counts.

Sure but I'm just trying to get a better picture of how projects are used across VOs and what naming conventions are used. If we want to do hierarchies of projects, that's going to affect how we re-organize the data within topology and ultimately it would behoove us to have at least one recommended naming convention.

> And isn’t it the case that every user on OSG-Connect is a member of the OSG VO? In other words, we are talking about the organization of projects within the OSG VO. When we move GLOW users to the OSG VO, we will use the same conventions.

Yes that is the case but I'm wondering why some projects are associated with OSG Connect and some with the OSG VO. What's the purpose of this distinction within our registered projects between "campus grids" and VOs?

matyasselmeci:
Perhaps this ticket is not the best place for a discussion...

lmichael107:
Agree with Mat.

Lauren Michael - Research Computing Facilitator, Center for High Throughput
Computing <http://chtc.cs.wisc.edu/>, University of Wisconsin - Madison
Other Affiliations: Wisconsin Institute for Discovery
<http://wid.wisc.edu/people/lauren-michael/>; Open Science Grid
<https://opensciencegrid.org/>; CaRCC <https://carcc.org/>
lmichael@wisc.edu, tinyurl.com/LMichaelCalendar, Discovery 2262,
(608)316-4430

PR #808, opened 2019-08-15
new project from colorado satellite

ChristinaLK:


ChristinaLK:
@brianhlin what do you mean about "having a sponsor without an associated VO"?

I confess I don't understand what that part means - I just copied an old project file.

brianhlin:
Projects can have their sponsor set to a Campus Grid and/or a VO (https://github.com/opensciencegrid/topology/blob/master/projects/MicroBooNE.yaml#L9-L11). I'm not entirely sure how the sponsor is used (if it all) but some of the conversations we're having about the definition of projects suggests that they should always belong to a VO.

Long story short, though, you should continue to register projects the way you have been and we'll let you know if that changes.

PR #819, opened 2019-09-03
OSG upgrade to 3.5.0

bockjoo:
We would like to upgrade one of our CEs to OSG 3.5.0.


brianhlin:
Hi @bockjoo as noted [here](https://opensciencegrid.org/docs/release/3.5/release-3-5-0/#summary-of-changes), OSG 3.5.0 does not contain HTCondor-CE but I expect that it will make it into 3.5.1.

bockjoo:
Thanks Brian! I will keep that in mind. Just wanted to test the installation.

PR #839, opened 2019-09-18
Added group for Professor Mondal from Clarkson University

CarrieBrown:


matyasselmeci:
Should this be `clarkson_mondal` instead of `clarson_mondal`?

CarrieBrown:
Yes, yes it should. I’ll correct that.

PR #849, opened 2019-10-01
Added UH_Gorham

CarrieBrown:


djw8605:
Are you sure it should be UH?  uh.edu is university of houston.

CarrieBrown:
Hawaii uses hawaii.edu - UH was what the user gave me as an acronym. Which would be preferred?

CarrieBrown:
According to their website, they use UH - https://www.hawaii.edu/about-uh/

PR #856, opened 2019-10-11
replace Hover with Xhao in updated ATLAS topology.

deesto:
John Hover is no longer associated with BNL or ATLAS.

I believe Caballero should also be removed from OASIS.


deesto:
> LGTM. I'll get this merged in for now but feel free to submit another PR to remove Jose from OASIS

Thank you, Brian.

PR #871, opened 2019-10-18
Add missing CHTC submit hosts

brianhlin:
cc @aaronmoate


edquist:
Wow, there's a learn and an osg-learn?  Who knew.

PR #875, opened 2019-10-22
Update BNL-Belle-II.yaml

xzhao87:
Add WLCGInformation section, setting InteropAccounting: true, so GRACC can monitor the accounting data from these CEs.


brianhlin:
Hi @xzhao87, I think you're going to need to populate the other fields before we can accept this

djw8605:
I think we only need `APELNormalFactor` set.

brianhlin:
@xzhao87 @djw8605 is this something that we still want to get in?

xzhao87:
Sorry, this one fell through the crack. Yes I think we still want to add the norm factor, which I don't have right now unfortunately. The person in charge is on vacation, I will talk to him late this week or early next week to get the norm factor. Thanks. Xin

PR #879, opened 2019-10-29
Added FF15IPQEXT

CarrieBrown:


projects/FF15IPQEXT.yaml
```diff
@@ -0,0 +1,10 @@
+Department: Chemistry
+Description: Parameterizing force fields for artificial amino acids through molecular dynamic simulations 
+FieldOfScience: Chemistry
+ID: '611'
+Name: FF15IPQEXT
+Organization: University of Pittsburg
```
brianhlin:
```suggestion
Organization: University of Pittsburgh
```

projects/FF15IPQEXT.yaml
```diff
@@ -0,0 +1,10 @@
+Department: Chemistry
+Description: Parameterizing force fields for artificial amino acids through molecular dynamic simulations 
+FieldOfScience: Chemistry
+ID: '611'
+Name: FF15IPQEXT
```
brianhlin: I thought we were using the `<Institution>_<PI Last name>` format?

projects/FF15IPQEXT.yaml
CarrieBrown: Good catch! I'll fix this.

projects/FF15IPQEXT.yaml
CarrieBrown: We are. This is a project that was created before this standardization was implemented but wasn't added to topology.

PR #889, opened 2019-11-05
Added new project: NII

rynge:


brianhlin:
Is this project a holdover from before the new `<INSTITUTION>_<PI Last Name>` naming scheme?

rynge:
Yes, it is direct collaboration and not a research group.

PR #902, opened 2019-12-02
Update AGLT2.yaml

wuwj6269:
deleting gate04 (machine retired) as a gatekeeper

PR #905, opened 2019-12-02
Retire AGLT2_SL6

brianhlin:
Closes #902 and closes #903. Downtimes already removed in #904


djw8605:
GRACC says that that specific CE stopped sending usage over a year ago.  So from a GRACC perspective, since we don't re-summarize more than a year back, we're clear for this removal.

PR #907, opened 2019-12-03
Combine Purdue resources into a single group

goughes:
This was requested by @jdost321. This is a pretty big change so let's discuss any additional things things that should happen in this PR.

I combined all our resources into a new Purdue.yaml, using Purdue-Hadoop.yaml as the base. I added all the resources from the Purdue-*.yamls and then deleted them. I wasn't sure what to do with the downtime yaml files so I left them for now. Should I combine them all into a single file called Purdue_downtime.yaml?


matyasselmeci:
Yes, the downtime files need to be combined as well.

PR #908, opened 2019-12-03
Replace smart quotes and dashes (software-3893)

ddavila0:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-3893


bin/osg-notify
```diff
@@ -115,6 +115,9 @@ def parseargs():
                          "(default: all)", )
     oparser.add_argument("--bypass-dns-check", action="store_true", dest="bypass_dns_check",
                          help="Bypass checking that one of the host's IP addresses matches with the hostanme resolution")
+    oparser.add_argument("--force-non-ascii", action="store_true", dest="force_non_ascii",
```
edquist: How about calling the option `--allow-non-ascii`?  (rather than `force`)

bin/osg-notify
```diff
@@ -173,6 +191,15 @@ def main():
     with open(args.message, 'r') as fp:
         contents = fp.read()
 
+        # Replace smart quotes and em/en dashes
+        contents = replace_smart_quotes_and_dashes(contents)
```
edquist: Note that this smart quote/dash conversion to ascii is done even if the option to explicitly allow non-ascii characters is specified

bin/osg-notify
```diff
@@ -144,6 +147,21 @@ def network_ok(bypass_dns_check):
...
+    # Replace smart quotes and em/en dashes
+    replaced = contents.replace('“','"').replace('”','"').replace("–","-").replace("—","-")
+    return replaced
+
+def has_non_ascii_characters(contents):
+    ret = False
+    # For each character in the message
+    for c in contents:
+        # Chek if the numeric value of the character is outside the ascii range (0-127)
+        if ord(c)< 0 or ord(c) > 127:
```
edquist: If this was really the check you wanted, note that you can equivalently write it as
```
if not 0 <= ord(c) <= 127:
```

Though:
- `ord(c)` can't be `< 0`, so this part of the check is sort of meaningless
- the printable range is `32-126`; do you really want to allow `NUL` bytes (`0`), control characters (`1-31`) and ascii `DEL` (`127`) ?

bin/osg-notify
```diff
@@ -173,6 +191,15 @@ def main():
     with open(args.message, 'r') as fp:
         contents = fp.read()
 
+        # Replace smart quotes and em/en dashes
+        contents = replace_smart_quotes_and_dashes(contents)
+
+        # Check for non-ascii characters
+        if has_non_ascii_characters(contents) and args.force_non_ascii == False:
+            print("ERROR: message contains non-ascii characters.")
```
edquist: consider sending such error messages to stderr with `print("...", file=sys.stderr)`

bin/osg-notify
ddavila0: done

bin/osg-notify
ddavila0: that's the expected behavior. Form the ticket:https://opensciencegrid.atlassian.net/browse/SOFTWARE-3893
"Also add --force option that allows the user to override the aforementioned error but still replaces smart quotes/dashes"

bin/osg-notify
ddavila0: I've changed it to only allow the 32-126 range

ddavila0: well that range and 10 which is '\n'

ddavila0: and 9 which is a tab. Hopefully I'm not leaving any character that can be used out

bin/osg-notify
ddavila0: done

bin/osg-notify
edquist: Ok!  (@brianhlin - i guess i'm surprised that even when explicitly allowing non-ascii characters, smartquotes still get zapped, but it sounds like this was intentional in any case.)

bin/osg-notify
```diff
@@ -144,6 +147,22 @@ def network_ok(bypass_dns_check):
         print("  https://opensciencegrid.org//operations/services/sending-announcements/")
         return False
 
+def replace_smart_quotes_and_dashes(contents):
+    # Replace smart quotes and em/en dashes
+    replaced = contents.replace('“','"').replace('”','"').replace("–","-").replace("—","-")
+    return replaced
+
+def has_non_printable_ascii_characters(contents):
```
edquist: Diego i'm sorry i prompted you to mess with this so much.  It should work fine as you have it, especially since the main use-case it to catch (non-ascii) unicode codepoints, which are all > 126.

Though i'm just gonna throw it out there, this whole function could perhaps be equivalently written as

```
def has_non_ascii(s):
    return re.search(r'[^ -~\t\n]', s) is not None
```

that is, if the search for a regex pattern {that contains any characters _not_ in the character set of the printable ascii-range and tab and newline} matches, return true, else false.  I think it's identical to what you have, just short-hand with the regex.  Feel free to use whichever you like best.

ddavila0: Thanks fo the suggestion Carl but if both functions get the same result, I will just not bother to make another change ;)

PR #915, opened 2019-12-09
In CHTC.yaml changed submit-4 to submit4

aaronmoate:
To make it consistent with new submit node naming convention

### DIFF
--- a/topology/University of Wisconsin/CHTC/CHTC.yaml
+++ b/topology/University of Wisconsin/CHTC/CHTC.yaml
@@ -188,7 +188,7 @@ Resources:
     VOOwnership:
       GLOW: 100
 
-  CHTC_SUBMIT_4:
+  CHTC-submit4:
     Active: true
     ContactLists:
       Administrative Contact:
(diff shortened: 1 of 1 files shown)


brianhlin:
@aaronmoate do you need to do the same thing for submit-1 or will that change come later?

aaronmoate:
Later;  submit-1.chtc.wisc.edu's name isn't changing until we
eventually reprovision it, which may be a while.

-Moate

brianhlin:
@djw8605 I forgot to ask during the meeting. Will this change have any effect on records coming from submit 4 over the past year? How do we handle an FQDN and/or resource rename for payload records?

aaronmoate:
FWIW, I changed the resource ID, so I imagine it will be seen as
a completely new resource.

-Moate

djw8605:
In effect, this commit deletes submit-4 and replaces it with submit4.  By convention, we don't delete resources from Topology until a year after the last usage.

Yes, this commit would affect usage in GRACC, as it is deleting a resource.

aaronmoate:
Changes made as requested, I think.

-Moate

PR #931, opened 2019-12-16
Change PSC Bridges resource name: it may not be just CMS in the future

mmascher:


mmascher:
FYI I have also added the CE to the hosted ce repo. PR to merge: https://gitlab.mwt2.org/osg/hosted-ce-config/merge_requests/5

PR #932, opened 2019-12-16
Update Purdue APEL normalization factor and HS06

goughes:


goughes:
That is good information. Is this true for all resources? Even ones where we set:
```
VOOwnership:
      CMS: 0
```
The Brown, Rice and Halstead resources are opportunistic slots we do not own (so CMS ownership is set to 0). The Hammer and Hadoop ones are nodes purchased by CMS so the value is 100.

I think we might just need to set a single value for all APELNormFactor entries if the Brown/Rice/Halstead APEL entries are included in the average that you send.

goughes:
@djw8605 can you take a look at my comment above and see if it makes sense? I'm wondering if I need to update the APELNormFactor for every resource to be 15.07, instead of having different values for each resource.

The 15.07 value is what we report to CMS here: https://docs.google.com/spreadsheets/d/1Ps-BGET7AlGvCFWyE0cT17tJjHreAkKF4cgNLb2MdOA/edit?pli=1#gid=0

djw8605:
15.07 for every resource makes sense.  At least it will be a predictable value, rather than the averaged off all of the cluster's values.

goughes:
Thanks @djw8605! I made a new PR.

PR #938, opened 2019-12-20
Added the Pitt_Koes project

CarrieBrown:

PR #939, opened 2019-12-20
end of december projects

ChristinaLK:


matyasselmeci:
Hi Christina,
Same thing as #938 , can you add the abbreviations for MSU and UCLA to `mappings/project_institution.yml`? The lines should be
```
MSU: "Montana State University"
UCLA: "University of California, Los Angeles"
```
(ASU is already in there)

ChristinaLK:
done!

PR #950, opened 2020-01-07
Adding the Internet 2 houstson cache

efajardo:
The facility and the site are kinda odd since the site is all over the place. Let me know if this is ok


brianhlin:
Do we have any other Internet2 registered caches? I'm open to the idea of having an `Internet2` facility and having location-specific site names, e.g. `I2 - Houston` based off of https://github.com/opensciencegrid/docs/pull/642/files. In that case, we'll also want to make the resource group yaml file names unique between the different Internet2 sites.

You'll also need to fill in more details for the `SITE.yaml`. See https://github.com/opensciencegrid/topology/blob/master/template-SITE.yaml

@djw8605 if we rename/move already registered caches around, what effect does that have on accounting?

efajardo:
@brianhlin there are no other registered but I can try to register them all now. So each location would be a site. But the only thing I have for each site is long and lat is that enoug, or are there other hard requirements?

brianhlin:
Also add a `LongName` and `Description`. @matyasselmeci are the city/address required?

matyasselmeci:
Technically no. Latitude/Longitude would be nice so it shows up on the map.

djw8605:
Topology is not consulted in stashcache accounting.  Nothing done here will affect accounting.

efajardo:
@brianhlin I added the rest of the sites of Internet2 can you please check it so I can bring the houston cache to production please?

topology/Internet2/Internet2Chicago/I2ChicagoInfrastructure.yaml
```diff
@@ -0,0 +1,26 @@
+Production: true
+SupportCenter: Community Support Center
+GroupDescription: Services located at the Internet2 Chicago POP
```
brianhlin: What's a "POP"?

efajardo: "Point of Presence"

brianhlin: Could you spell that out in all of the descriptions?

PR #957, opened 2020-01-09
Create new EIC yaml topology file

deesto:
New "virtual" VO topology file for EIC, for the purpose of creating a new OASIS repository, eic.opensciencegrid.org. Note that the ID 'ID: 137' on line 25 was copied from another file and may need to be modified?

### DIFF
--- a/virtual-organizations/EIC.yaml
+++ b/virtual-organizations/EIC.yaml
@@ -22,7 +22,7 @@ FieldsOfScience:
   - Nuclear Physics
   SecondaryFields:
   - Physics and astronomy
-ID: 137
+ID: 139
 LongName: The Electron-Ion Collider eXperiment at RHIC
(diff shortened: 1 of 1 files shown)


virtual-organizations/EIC.yaml
```diff
@@ -0,0 +1,41 @@
...
+  VO Manager:
+  - ID: acbb79b1cef1d22d7ca443ccd1aa9556750a8f3f
+    Name: Christopher Pinkenburg
+Disable: false
+FieldsOfScience:
+  PrimaryFields:
+  - Nuclear Physics
+  SecondaryFields:
+  - Physics and astronomy
+ID: 137
```
brianhlin:
```suggestion
ID: 139
```

Here's the ID that you want. Once you accept the suggestion we can get this merged.

PR #958, opened 2020-01-10
Add downtime for UCSDT2 for work on cooling system

jdost321:


jdost321:
sorry i forgot to bump the DT ID numbers for each successiver service..

PR #964, opened 2020-01-15
Adjust structure of OASIS Managers in VOs (SOFTWARE-3947)

edquist:
So for SOFTWARE-3947, we want to change the OASIS Managers section from a dict of dicts (keyed off of the Manager Name) to a list of dicts, with a new Name attribute instead of the key from the original dict.

Adjusting the yaml data itself isn't too bad, but we have to modify the webapp code that processes this section.  I ended up replacing the call to `expand_attr_list` (which took the old dict) with a new `order_dict` function that provides the same ordering without moving the key to a Name attribute.

@matyasselmeci, you wrote the original code that i'm modifying here, so if you could sanity check it that'd be great.

### DIFF
--- a/src/webapp/common.py
+++ b/src/webapp/common.py
@@ -144,6 +144,19 @@ def expand_attr_list(data: Dict, namekey: str, ordering: Union[List, None]=None,
     return newdata
 
 
+def order_dict(value: Dict, ordering: List, ignore_missing=False) -> OrderedDict:
+    """
+    Convert a dict to an OrderedDict with key order provided by ``ordering``.
+    """
+    new_value = OrderedDict()
+    for elem in ordering:
+        if elem in value:
+            new_value[elem] = value[elem]
+        elif not ignore_missing:
+            new_value[elem] = None
+    return new_value
+
+
 def to_xml(data) -> str:
     return xmltodict.unparse(data, pretty=True, encoding="utf-8")
 
--- a/src/webapp/vos_data.py
+++ b/src/webapp/vos_data.py
@@ -140,7 +140,6 @@ class VOsData(object):
         """
         new_managers = copy.deepcopy(managers)
         for i, data in enumerate(managers):
-            name = data["Name"]
             if not is_null(data, "DNs"):
                 new_managers[i]["DNs"] = {"DN": data["DNs"]}
             else:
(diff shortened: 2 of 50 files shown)


matyasselmeci:
Tagged this do-not-merge so we don't accidentally do it; the webapp must be updated and restarted _immediately_ after the merge or it will barf on the data.

(Personally I would prefer that the code understand both the old and the new data...)

edquist:
> (Personally I would prefer that the code understand both the old and the new data...)

Updated to support both now.

(Though they are being updated together so i'm not sure what the use case for supporting both might be.)

matyasselmeci:
> > (Personally I would prefer that the code understand both the old and the new data...)
> 
> Updated to support both now.
> 
> (Though they are being updated together so i'm not sure what the use case for supporting both might be.)

Thanks! The point for supporting both is that the code and the data are accessed from two different clones (on disk) so even though they're updated together upstream, it's possible for them to get out of sync, and I don't want the webapp to crash because of that.

matyasselmeci:
Looks good; you just need to update the template file as well (`/template-virtual-organization.yaml`)

edquist:
Ah right, updated.

matyasselmeci:
This is deployed on topology-itb: compare [production](https://topology.opensciencegrid.org/vosummary/xml) with [itb](https://topology-itb.opensciencegrid.org/vosummary/xml). Looks good to me; @brianhlin , wanna give it a once-over?

edquist:
Thanks @matyasselmeci! Just to double-check, is the itb instance using the updated data (for virtual-organization definitions) also, in addition to the webapp sources?

matyasselmeci:
As far as I can tell, yes:
```
# grep "^TOPOLOGY_" /etc/opt/topology/config-itb.py
TOPOLOGY_DATA_DIR = "/var/cache/topology-itb/topology-itb"
TOPOLOGY_DATA_REPO = "https://github.com/opensciencegrid/topology"
TOPOLOGY_DATA_BRANCH = "itb"

# git --git-dir /var/cache/topology-itb/topology-itb/.git log -1
commit 0b101471d2fbe5a690a8b318144e7ed3afc93323
Merge: 3b1eef9 daa2958
Author: Matyas Selmeci <matyas@cs.wisc.edu>
Date:   Thu Jan 16 13:16:04 2020 -0600

    Merge remote-tracking branch 'edquist/SOFTWARE-3947.OASIS-CILogon' into itb
...
```

matyasselmeci:
Merge and deployed.

PR #971, opened 2020-01-24
Reorganize University of Utah site topology

brianhlin:
Follow site-naming conventions (https://opensciencegrid.org/docs/common/registration/#new-site) also including SLATE by request of the SLATE team.

@djw8605 we have some records in June/July 2019  associated with FQDNs associated with the CHPC site that are no longer registered in Topology (these surprisingly show up in the GRACC dashboards [2]). Is removing the CHPC site kosher here or should we keep it around? Should we re-register the old FQDN as an inactive resource?

[1] https://gracc.opensciencegrid.org/kibana/goto/3fb36dc35b65ffac0c77950fbb606a7d
[2] https://gracc.opensciencegrid.org/dashboard/db/site-summary?orgId=1&from=now-1y&to=now&var-interval=$__auto_interval&var-site=CHPC&var-type=Batch


topology/University of Utah/Utah-SLATE-Kingspeak/Utah-SLATE-Kingspeak.yaml
```diff
@@ -0,0 +1,32 @@
...
+Production: true
+Resources:
+  SLATE_US_UUTAH_KINGSPEAK:
+    Active: true
+    ContactLists:
+      Administrative Contact:
+        Primary:
+          ID:  40a181cc3b7a43593188c9607310d35cb079bda9
+          Name: Mitchell Steinman
+      Resource Report Contact:
```
matyasselmeci: Do we still need a resource report contact?

brianhlin:
Marking as `do-not-merge` until we get word from @djw8605 about the GRACC consequences

djw8605:
I think you will need to keep the CHPC around in topology for summarization to work.

brianhlin:
Instead of re-adding the CHPC site, I moved the old resource so that it would be resummarized under the new `Utah-SLATE-Notchpeak` site

PR #987, opened 2020-02-06
Consolidate Augustine Abarises

matyasselmeci:
(There's a duplicate contact entry for Augustine which I want to get rid of)


matyasselmeci:
The other one does have separate email addresses, it's just they're very similar. It's also the one that has his GitHub username, which is why it's the one I want to keep.

PR #1001, opened 2020-02-14
Add cedar UConn hosted CE

mmascher:


mmascher:
FYI: it would be nice to recover the data here https://gracc.opensciencegrid.org/kibana/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-1y,mode:quick,to:now))&_a=(columns:!(_source),index:gracc.osg.summary,interval:auto,query:(query_string:(analyze_wildcard:!t,lowercase_expanded_terms:!f,query:'ProbeName:htcondor-ce%5C:hosted-ce28.grid.uchicago.edu')),sort:!(EndTime,desc))

so that it shows up in Gracc. I have recently re.created the 99-local.ini using the conventions of the hsoted-ce github repo. Also, I have jsut changed Glidein Resource Name in the factory to match that.

matyasselmeci:
Hi Marco, do you have a Site Contact for this?

mmascher:
Yes, the site contact is Richard Jones.

matyasselmeci:
@djw8605 , when you're back, can you take a look at recovering the GRACC data as @mmascher mentioned in the comment above?

djw8605:
What does "recover" mean.  It looks like the records are tagged with Connecticut correctly.

mmascher:
@djw8605 I thought that changing GLIDEIN_ResourceName in the factory, and resource_group in the 99-local.ini file (the osg-configure one) meant the data was not available in GRACC anymore. The only thing that I kept is "resource" in 99-local-ini. In general, is this enough to guarantee that everything shows up in GRACC?

djw8605:
I'm not fully certain what 99-local.ini does with osg-configure, @matyasselmeci can answer that.

But, for GRACC.  The matching is between the ProbeName from the record and the fqdn in Topology.  GLIDEIN_ResourceName only affect Payload jobs.

The key is that the:
- (Pilot Records) ProbeName needs to match the fqdn in Topology
- (Payload Records) GLIDEIN_ResourceName is the same as the Resource, Site, or Resource Group.

matyasselmeci:
The resource setting in your osg-configure configs is used for:
- the `OSG_Resource` attribute in the classad that gets sent to the CE collector (collector.opensciencegrid.org)
- the `ProbeName` in the gratia probes

djw8605:
Do you mean SiteName?  ProbeName is usually `<scheduler>:<fqdn>`

SiteName is **ONLY** used if the GRACC cannot find a match to the fqdn.

matyasselmeci:
Yes, my mistake. It's used for SiteName.

PR #1005, opened 2020-02-17
Feb projects

ChristinaLK:


matyasselmeci:
Got a few questions about the Internet2 project:
- is this going to be the only project from Internet2? Can't we use the INSTITUTION_PROJECTNAME scheme?
- we get our fields of science from the left column of [this table](https://topology.opensciencegrid.org/nsfscience/csv). Is there something more appropriate than "other"? "Infrastructure Development" perhaps? If there isn't, that's fine, though you'll need to write it as "Other" so it can be found in the table.

ChristinaLK:
I just blanked when I created the project and didn't use the PI name. 

Infrastructure Development is a good idea. I'll update.

PR #1038, opened 2020-03-05
add functions for oasis managers endpoint (SOFTWARE-3947)

edquist:
For an authorized request, the new endpoint will take a VO and
return the associated OASIS Managers with ssh pubkeys and DNs.

The new `get_oasis_manager_endpoint_info` function facilitates
collecting and organizing all the relevant information from
contacts, topology, and cilogon ldap to return the information
for this endpoint.

### DIFF
--- a/src/webapp/oasis_managers.py
+++ b/src/webapp/oasis_managers.py
@@ -23,7 +23,7 @@ def get_oasis_manager_endpoint_info(global_data, vo, ldappass):
             ContactID = manager['ID']
             Name = manager['Name']
             DNs = manager['DNs']
-            CILogonID = contact_cilogon_ids[ContactID]['CILogonID']
+            CILogonID = contact_cilogon_ids[ContactID].cilogon_id
             ssh_keys = ssh_keys_map.get(CILogonID, [])
             info.append({
                 'ContactID': ContactID,
@@ -37,8 +37,8 @@ def get_oasis_manager_endpoint_info(global_data, vo, ldappass):
 
 def get_contact_cilogon_id_map(global_data):
     """ return contacts dict, limited to users with a CILogonID """
-    contacts = global_data.get_contacts_data()
-    return { k: v for k, v in contacts.items() if "CILogonID" in v }
+    contacts = global_data.get_contacts_data().users_by_id;
+    return { k: v for k, v in contacts.items() if v.cilogon_id is not None }
 
 
 def get_vo_oasis_managers(global_data, vo):
--- a/src/webapp/oasis_managers.py
+++ b/src/webapp/oasis_managers.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+
+import ldap
+
+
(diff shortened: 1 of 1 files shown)


src/webapp/oasis_managers.py
```diff
@@ -0,0 +1,86 @@
...
+    contacts = global_data.get_contacts_data()
+    return { k: v for k, v in contacts.items() if "CILogonID" in v }
+
+
+def get_vo_oasis_managers(global_data, vo):
+    """return OASIS Managers dict for given vo, if any, else an empty dict"""
+    vos_data = global_data.get_vos_data()
+    if vo in vos_data:
+        if "OASIS" in vos_data[vo]:
+            if "Managers" in vos_data[vo]["OASIS"]:
```
matyasselmeci: The thing you gotta be careful with this is that just because `vos_data[vo]["OASIS"]` exists, it doesn't mean that it's an iterable -- could be a None, for example. Yes, it _should_ be if our data is well-formed, but our data is not necessarily well-formed.

matyasselmeci: Check out `webapp.common.is_null()`, which can check if a dictionary tree exists. If I read it right (sorry for the lack of documentation), `webapp.common.is_null(vos_data, vo, "OASIS", "Managers")` might do what you want.

edquist: Good tip, i've replaced this with a function to traverse the vos_data dict without causing KeyErrors and with a default if the final result is None.

src/webapp/oasis_managers.py
```diff
@@ -0,0 +1,97 @@
...
+def get_contact_cilogon_id_map(global_data):
+    """ return contacts dict, limited to users with a CILogonID """
+    contacts = global_data.get_contacts_data().users_by_id;
+    return { k: v for k, v in contacts.items() if v.cilogon_id is not None }
+
+
+def get_vo_oasis_managers(global_data, vo):
+    """return OASIS Managers dict for given vo, if any, else an empty dict"""
+    vos_data = global_data.get_vos_data()
+    return safe_dict_get(vos_data, vo, "OASIS", "Managers", default=[])
```
matyasselmeci:
```suggestion
    return safe_dict_get(vos_data, vo, "OASIS", "Managers", default={})
```

edquist: Actually in 21526c5b96 i changed this from `{}` _to_ `[]`, because i think OASIS Managers is supposed to be a list.

edquist:
Just that it's not in use yet; but since it's a stand-alone feature-set, it's fine to merge it.  I'm removing the label.

src/webapp/oasis_managers.py
matyasselmeci: You should update the docstring to say list instead of dict, then.

edquist: Good point!  Updated.

edquist:
Force push was just to clean up git history, no changes to the PR.

PR #1047, opened 2020-03-09
Adding project file for Arizona_Paschalidis group

eharstad:


projects/Arizona_Paschalidis.yaml
```diff
@@ -0,0 +1,12 @@
+Name: Arizon_Paschalidis
```
matyasselmeci:
```suggestion
Name: Arizona_Paschalidis
```

matyasselmeci:
One small change (see my suggestion) and it should be fine.

PR #1052, opened 2020-03-12
adding NMSU_staff

lmichael107:
@djw8605 might want to know


projects/NMSU_staff.yaml
```diff
@@ -0,0 +1,12 @@
+Name: NMSU_Staff
+Description: NMSU staff experimenting with OSg
```
brianhlin:
```suggestion
Description: NMSU staff experimenting with OSG
```

lmichael107: good catch!

PR #1053, opened 2020-03-12
Load CILOGON_LDAP_PASSFILE if configured (SOFTWARE-3947)

edquist:
This includes pulling the `_readfile` helper function from the webhook_app into webapp.common


src/webapp/common.py
```diff
@@ -253,3 +253,15 @@ def load_yaml_file(filename) -> Dict:
...
+
+
+def readfile(path, logger):
+    """ return stripped file contents, or None on errors """
+    if path:
+        try:
+            with open(path, mode="rb") as f:
+                return f.read().strip()
+        except IOError as e:
+            logger.error("Failed to read file '%s': %s" % (path, e))
```
matyasselmeci:
```suggestion
            logger.error("Failed to read file '%s': %s", path, e)
```
at least if it's a standard logging object. This is the general pattern used elsewhere; the advantage is that if no output will be printed because of the log level, then Python doesn't bother with the substitution.

Doesn't _matter_ too much in this case, except for consistency's sake.

src/webapp/default_config.py
```diff
@@ -18,6 +18,8 @@
 CONTACT_DATA_BRANCH = "master"
 CONTACT_CACHE_LIFETIME = 60 * 2
 
+CILOGON_LDAP_PASS = None
```
matyasselmeci:
```suggestion
CILOGON_LDAP_PASSFILE = None
```
That's what you use in models.py.

src/webapp/common.py
matyasselmeci: Also you should do `if logger:` first, in case anybody wants to call readfile without a logger.

src/webapp/default_config.py
edquist: Good catch.

matyasselmeci:
OK, looks good. Once you get it working on topology-itb, I'll merge.

edquist:
Right, we still have to merge this.

So I tested this on topology-itb, updating /etc/opt/topology/config-itb.py to add the line
```
CILOGON_LDAP_PASSFILE = "/etc/opt/topology/cilogon-ldap"
```

And then creating that cilogon-ldap password file with permissions 0644  apache:root
```
-rw-------. 1 apache root 21 Mar 19 11:21 /etc/opt/topology/cilogon-ldap
```

I restarted apache and the existing webapp endpoints still work.

matyasselmeci:
Sounds good.

What happens if CILOGON_LDAP_PASSFILE is not configured? Do existing webapp functions work? My real question is: do we have to update the config *before* updating the software, or does it not matter?

edquist:
if it's not configured, `config.get("CILOGON_LDAP_PASSFILE")` should return `None`, then the call to `readfile()` will also, which will log a warning but should not break anything.

edquist:
>My real question is: do we have to update the config before updating the software, or does it not matter?

Just to be clear: it does not matter

matyasselmeci:
OK, thanks Carl. I'll merge and deploy tomorrow; you can also go ahead and ask the Nebraska admins for the config change.

edquist:
ticket for the ops work
https://support.opensciencegrid.org/a/tickets/64659

brianhlin:
Awesome! What's the endpoint for this data?

matyasselmeci:
I don't think there's an endpoint yet.

PR #1060, opened 2020-03-17
new projects for OSG Connect late Feb - early Mar

ChristinaLK:
This one has at least one new university.

### DIFF
--- a/projects/Flightworthy.yaml
+++ b/projects/Flightworthy.yaml
@@ -1,0 +1,12 @@
+Name: Flightworthy
+Description: HTCondor development activities
+Department: Computer Science
+FieldOfScience: Computer Science
+Organization: University of Wisconsin-Madison
+PIName: Miron Livny
+
+ID: '654'
+
+Sponsor:
+  CampusGrid:
+    Name: OSG Connect
--- a/projects/MCBI.yaml
+++ b/projects/MCBI.yaml
@@ -1,0 +1,12 @@
+Name: MCBI
+Description: processing images with Freesurfer
+Department: Martinos Center for Biomedial Imaging
+FieldOfScience: Health
+Organization: Harvard University
+PIName: Davide Momi
+
+ID: '652'
+
+Sponsor:
+  CampusGrid:
+    Name: OSG Connect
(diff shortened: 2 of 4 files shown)


projects/Flightworthy.yaml
```diff
@@ -0,0 +1,12 @@
...
+Department: Computer Science
+FieldOfScience: Computer Science
+Organization: University of Wisconsin-Madison
+PIName: Miron Livny
+
+ID: '654'
+
+Sponsor:
+  CampusGrid:
+    Name: OSG Connect
```
matyasselmeci:
```suggestion
  VirtualOrganization:
    Name: GLOW
```

ChristinaLK: Even if their accounts are on OSG Connect + therefore submitting through the OSG VO?

ChristinaLK: I got the go ahead from Lauren to just call it Flightworthy. Exceptions are occasionally okay. :)

brianhlin: I'd rather keep this as `OSG Connect` instead of introducing a new value for a field that I think is ultimately meaningless: https://opensciencegrid.atlassian.net/browse/SOFTWARE-3763

matyasselmeci: Should we drop the Sponsor field for new registrations?

brianhlin: I don't think we can until we resolve the aforementioned ticket. My guidance in the interim:

- If they're using a CI Connect submit point, use a `CampusGrid` sponsor reflecting the CI Connect submit point they're associated with
- Otherwise, use the relevant VO

projects/MCBI.yaml
```diff
@@ -0,0 +1,12 @@
+Name: MCBI
+Description: processing images with Freesurfer
+Department: Martinos Center for Biomedial Imaging
+FieldOfScience: Health
```
matyasselmeci: Sorry, merged this too soon. Could you change this to one of
- `Biomedical research`
- `Medical Imaging`
- `Medical Sciences`
?
whichever is most appropriate?

The list of fields of science we should use is in the left column of [this spreadsheet](https://topology.opensciencegrid.org/nsfscience/csv).

ChristinaLK: Done in #1062

PR #1081, opened 2020-03-31
Added new project: COVID19_Illinois_Gammie

rynge:


brianhlin:
@lmichael107 is the plan to prepend `COVID19_` to project names like this?

rynge:
Yes, discussed in #facilitation and that is what Lauren wanted.

PR #1082, opened 2020-04-01
Add UChicago_Open_Science_CE resource

matyasselmeci:


topology/University of Chicago/UChicago/UChicago_Open_Science_CE.yaml
```diff
@@ -0,0 +1,61 @@
...
+        Details:
+          hidden: false
+
+      Submit Node:
+        Description: The submit part of this Open Science CE
+        Details:
+          hidden: false
+
+    VOOwnership:
+      OSG: 100
```
brianhlin, line 60: I don't think we ever want to set `VOOwnership` to OSG since accounting records are compared to VOOwnership to determine opportunistic usage. In this case, any OSG VO pilots running through the CE will not be considered opportunistic since OSG owns the resource. It may not actually matter if we're disabling accounting on the CE portion but we should do the right thing here.

matyasselmeci: Gotcha. I'm pretty sure VOOwnership is optional.

PR #1089, opened 2020-04-03
Add new GPU resource to ND

khurtado:
Notre Dame CAML GPU cluster

### DIFF
--- a/topology/University of Notre Dame/NWICG_NDCMS/ND_CAMLGPU.yaml
+++ b/topology/University of Notre Dame/NWICG_NDCMS/ND_CAMLGPU.yaml
@@ -1,0 +1,44 @@
+GroupDescription: ND_CAMLGPU
+GroupID: 80
+Production: true
+Resources:
+  ND_CAMLGPU:
+    Active: true
+    ContactLists:
+      Administrative Contact:
+        Primary:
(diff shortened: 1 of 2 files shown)


djw8605:
Hi,

The current site is ND - CMS_GPU.   Could you also add another resource with that name, even if it's not used in the future?  That would allow Accounting to understand that resource.

-Derek

matyasselmeci:
Hi Kenyi,
What's running on the host you're registering? `HeadNode` is not a valid service type, so I'm trying to figure out what it should be.

khurtado:
Hi Matyas,

It's a condor submit node. There is an OSG Hosted CE submitting to that node. What should I put instead?

matyasselmeci:
I don't think we register those; we just register the hosted CE. @djw8605, do we want to start keeping track of those too? In that case, we should probably create a new service type since none of the others seem to apply.

djw8605:
If this is a hosted-ce site, then only the hosted-ce should be registered.  We haven't registered login nodes in the past.

khurtado:
Ok, I just changed this to CE and put `hosted-ce35.grid.uchicago.edu` instead. I also added "ND - CMS_GPU" with a made-up CE FQDN, but as non-active resource to take care of the current accounting name entry.

PR #1095, opened 2020-04-06
Correct nd

djw8605:
Fea complications:

1. Can resource names have spaces?
2. I changed the fqdn to something that will never match so we don't have 2 resources (the other ND) with the same FQDN, and GRACC won't get confused.


brianhlin:
Why do we have a duplicate resource for this CE? It seems like we should just remove the old one @khurtado

djw8605:
GRACC still has records with the old  `ND - CMS_GPU` name in it.

brianhlin:
@djw8605 payload records I assume?

djw8605:
Yes

brianhlin:
Could you add a comment above the FQDN line to note this?

djw8605:
comment added.

PR #1100, opened 2020-04-09
Update MarkupSafe to avoid bnug reintroduced by a steuptools update

brianhlin:
https://github.com/pallets/markupsafe/issues/57#issuecomment-597105876

Worked fine in my local dev environment. It seems that Topology isn't currently triggering this because we're pulling in Python 3.6.7 for our CI whereas the contacts CI is pulling in 3.6.10.


matyasselmeci:
actually, now that they're using 3-digit versions, do this instead:

brianhlin:
Updated the ITB and the XML looked sane.

PR #1102, opened 2020-04-10
Update EIC contacts

brianhlin:
Based on out-of-band request by Jerome. Related to
https://support.opensciencegrid.org/public/tickets/5bf2be7659c281bca7ab2ba81685809e4a54d692e1b45ab2fb381c6671aa747b


edquist:
Just to confirm since i did not see it explicitly in the ticket (so i suppose this was the out-of-band part) -- Christopher Pinkenburg is intentionally removed here, right?

brianhlin:
Correct!

PR #1112, opened 2020-04-16
Registering Hosted CE for AMNH HEL cluster

jdost321:


jdost321:
I just asked the admin in the ticket if he wants this split as 2 sites in topology. I'll reply and update the PR accordingly when I know.

jdost321:
I fixed ARES active and set to "true" the admins don't want to split in 2 sites, they prefer all usage to be counted for AMNH. I will set the HEL cluster to "true" as well once we've proven it works end to end.

PR #1117, opened 2020-04-21
Create SITE.yaml

RelativePrime:
Creating for ORNL ALICE T2 site


topology/Oak Rigde National Labortory/ORNL/SITE.yaml
```diff
@@ -0,0 +1,9 @@
+LongName: Oak Ridge National Lab
+Description: Institutional computing center for ORNL
+ID:
```
brianhlin, line 3:
```suggestion
ID: 10274
```

PR #1126, opened 2020-04-24
fixing PI for reporting

lmichael107:


projects/COVID19_FoldingAtHome.yaml
```diff
@@ -4,7 +4,7 @@ FieldOfScience: Biochemistry
 ID: '661'
 Name: COVID19_FoldingAtHome
 Organization: University of Nebraska-Lincoln
```
brianhlin, line 6: Should we also update the org?

lmichael107: yes! Did we ever agree on the "Multi-Institutional" option? See: https://foldingathome.org/about/the-foldinghome-consortium/

lmichael107:
We should also fix Organization? I was just asking @brianhlin about "Multi-Institutional", given https://foldingathome.org/about/the-foldinghome-consortium/

brianhlin:
I think we could just use `Folding@Home Consortium` like we do https://github.com/opensciencegrid/topology/blob/master/projects/LIGO.yaml#L6

projects/COVID19_FoldingAtHome.yaml
brianhlin, line 6:
```suggestion
Organization: Folding@Home Consortium (FAHC)
```

djw8605: Just put it now, we can fix it later.  Lets get this merged.

lmichael107:
looks good!

PR #1149, opened 2020-05-06
[CIT_CMS_T2] Add local scheduler to yaml file. Main point to include …

juztas:
…resource usage in EGI Reports

@djw8605 fyi


djw8605:
The FQDN needs to stay the same to match the current records.

brianhlin:
Ah, I see. Let's hope we don't have any collisions in the future!

PR #1154, opened 2020-05-11
Add CILogonID to various XMLs

matyasselmeci:


matyasselmeci:
Tested the changed pages on a local instance on my laptop; the CILogonID attribute showed up as expected. I need to add the CILogonID of someone who's a contact for a VO in order to test the vosummary xml though.

matyasselmeci:
Added it to the OASIS manager section too; turns out Dave Dykstra's CILogonID was also in the contacts DB, and I saw his ID in the vosummary XML, so we should be good there.

matyasselmeci:
@edquist can you review? @brianhlin doesn't have the time

src/webapp/contacts_reader.py
```diff
@@ -31,6 +31,7 @@ def get_tree(self, authorized=False, filters=None) -> Optional[OrderedDict]:
             self.yaml_data["ContactInformation"]["PrimaryEmail"])
         tree["Profile"] = self.yaml_data.get("Profile", None)
         tree["GitHub"] = self.yaml_data.get("GitHub", None)
+        tree["CILogonID"] = self.yaml_data.get("CILogonID", None)
```
edquist, line 34: This reminds me - I think at some point per discussion with you i made a `cilogon_id` property [here](https://github.com/opensciencegrid/topology/blob/6b9eb1d7/src/webapp/contacts_reader.py#L61-L63) rather than touching the yaml data directly.  Is that relevant for any of your changes?

matyasselmeci: I think for use within the same class, the extra layer of indirection from the property isn't necessary; outside of User, I use the cilogon_id property.

src/webapp/topology.py
```diff
@@ -185,13 +185,14 @@ def _expand_contactlists(self, contactlists: Dict, authorized: bool) -> Dict:
...
             for contact in contact_data:
                 contact_id = contact.pop("ID", None)  # ID is for internal use - don't put it in the results
-                if authorized and self.common_data.contacts:
-                    if contact_id in self.common_data.contacts.users_by_id:
-                        extra_data = self.common_data.contacts.users_by_id[contact_id]
-                        contact["Email"] = extra_data.email
-                        contact["Phone"] = extra_data.phone
-                        contact["SMSAddress"] = extra_data.sms_address
-                        dns = extra_data.dns
+                if self.common_data.contacts and contact_id in self.common_data.contacts.users_by_id:
```
edquist, line 188: is the previous `and self.common_data.contacts` check no longer relevant?  Or is it guaranteed to be true/nonempty now or something?

src/webapp/topology.py
```diff
@@ -185,13 +185,14 @@ def _expand_contactlists(self, contactlists: Dict, authorized: bool) -> Dict:
...
+                    user = self.common_data.contacts.users_by_id[contact_id]  # type: User
+                    contact["CILogonID"] = user.cilogon_id
```
edquist, line 190: I guess this is obvious - but the idea here is that the `CILogonID` field does not not require authorization, right?

matyasselmeci: Correct, it's public.

src/webapp/topology.py
matyasselmeci: I'm not sure what you mean. I'm checking for `self.common_data.contacts` right here.

src/webapp/vos_data.py
```diff
@@ -151,11 +150,16 @@ def _expand_oasis_legacy_managers(managers):
...
             else:
                 new_managers[name]["DNs"] = None
-        return {"Manager": expand_attr_list(new_managers, "Name", ordering=["ContactID", "Name", "DNs"],
+
+            new_managers[name]["CILogonID"] = None
+            if self.contacts_data:
+                user = self.contacts_data.users_by_id.get(data["ID"])
+                if user:
+                    new_managers[name]["CILogonID"] = user.cilogon_id
+        return {"Manager": expand_attr_list(new_managers, "Name", ordering=["Name", "CILogonID", "DNs"],
```
edquist, line 159: Just to confirm - does this mean that now with `CILogonID` we stop caring about `ContactID` ?

matyasselmeci: I deleted ContactID because it never existed. I don't know how it got there -- maybe I meant "ID"? But ID is for internal use and we don't expose it in the rgsummary and vosummary XMLs, only miscuser.

src/webapp/topology.py
edquist: Ok my eyes were playing ticks on me

PR #1164, opened 2020-05-14
The WT2 CE no longer need to report to WLCG

djw8605:
FD #64971


djw8605:
Ok to merge, even with that odd "push" CI failure?

matyasselmeci:
Yeah it's fine.

PR #1167, opened 2020-05-15
Added missing XD projects

rynge:


projects/TG-DMR190101.yaml
```diff
@@ -0,0 +1,12 @@
+Name: TG-DMR190101
+Description: 1D Nanoconfined Helium: A Versatile Platform for Exploring
```
matyasselmeci, line 2:
```suggestion
Description: "1D Nanoconfined Helium: A Versatile Platform for Exploring"
```
There's a colon in there so you need to quote it.

PR #1170, opened 2020-05-18
Update Wilson Facility.yaml - New CE, new contacts

tiradani:
The Wilson cluster at Fermilab has been restructured with an HTCondor-CE.  It also has new contacts.


brianhlin:
Hi @tiradani could you also add appropriate security contacts for this resource?

tiradani:
Do I need to include an ID?  I removed that section because FNAL security isn't in the contacts list at https://topology.opensciencegrid.org/contacts.  In the past, I could just input an email address.

tiradani:
Do I need to include an ID?  I removed that section because FNAL security isn't in the contacts list at https://topology.opensciencegrid.org/contacts.  In the past, I could just input an email address.

tiradani:
Thank you for your help in registering the security contacts in the contacts DB.  I have added them to the topology yaml file.

PR #1172, opened 2020-05-18
Fix mapping for Wayne State University: turns out we've been using WayneStateU, not WayneState for their projects

matyasselmeci:


brianhlin:
Is `WayneStateU` preferred?

matyasselmeci:
We have several projects registered "WayneStateU" and none registered with "WayneState".

brianhlin:
Was that a conscious decision or just based on momentum? @lmichael107 ?

lmichael107:
Looks like, according to them (https://wayne.edu/) it should be "Wayne State" for short, so "WayneState" in topology.

lmichael107:
probably momentum, either way

brianhlin:
@djw8605 would it be a problem if we change any of the `WayneStateU_*` projects in topology to `WayneState_*`?

djw8605:
Users have already used the project name WayneStateU_Majumder.  We can do a project name correction though to correct to `WayneState_Majumder`, and ask the user to change their submit files.  GRACC does not read this project institution mapping file.

brianhlin:
@lmichael107 do you want to go through this process of fixing the WayneState project names (notifying the user, updating submit host project lists?), or should we just change the acceptable mapping to `WayneStateU`? We could also grandfather in existing projects and have both the `WayneState` and `WayneStateU` prefixes and have your team use the former from now on

matyasselmeci:
I'm confused. Where does that website say that abbreviations should be "Wayne State"? I see "Wayne State University" right at the top.

lmichael107:
Whenever the institution refers to itself *without* the full name, it uses
"WayneState" (even on their home page). Our intention with the project
names is to use the institutions preferred short-form name.

We should check with Lincoln/UChicago, but the ProjectName is automatically
set for users on OSG Connect (they don't add it to their submit files,
themselves) so he may be able to do some magic. If not, then I suppose we
can keep and move forward with "WayneStateU". Either way, we may have to
play whack-a-mole, but it should all be with known users of OSG Connect.

thoughts? As always, I'm sorry that I and any other RF team members are
still making these mistakes. We try to be careful.

Lauren

Lauren Michael - Research Computing Facilitator, Center for High Throughput
Computing <http://chtc.cs.wisc.edu/>, University of Wisconsin - Madison
Other Affiliations: Wisconsin Institute for Discovery
<http://wid.wisc.edu/people/lauren-michael/>; Open Science Grid
<https://opensciencegrid.org/>; CaRCC <https://carcc.org/>
lmichael@wisc.edu, tinyurl.com/LMichaelCalendar, Discovery 2262,
(608)316-4430

brianhlin:
@LincolnBryant what does it take to change a project name on OSG Connect submit hosts?

PR #1179, opened 2020-05-19
line breaks in COVID19 descriptions

lmichael107:


matyasselmeci:
Sorry, my email was brief; you'll need an extra _blank_ line to add a line break. Like so:
```
"Folding at Home for COVID-19, on the Open Science Grid 

https://foldingathome.org/covid19/"
```

lmichael107:
Yes, I think we may have an off-by-one error in "line break" definition. I was just looking to keep things from being on a single line, by inserting a single (not-blank) line break. :)

PR #1185, opened 2020-05-22
Missed a place in vosummary schema where CILogonID can occur

matyasselmeci:


matyasselmeci:
Ah right, you're on vacation. Well, the CI passes, which was the point, so I'm merging.

PR #1200, opened 2020-06-04
downtime over

hseverini:


matyasselmeci:
Travis reporting got wedged (again) which is probably why @osg-bot didn't trigger, but it looks good and the tests did pass.

PR #1208, opened 2020-06-06
Add /check_suite route to replace /status (SOFTWARE-4119)

edquist:
Travis CI now sends Check Run and Check Suite events instead of Status events.

The payload for `check_suite` events is a bit different than the `status` event payload, but it contains the same essential information.


edquist:
Automerge is working again:
https://github.com/opensciencegrid/topology/pull/1209

And also correctly handling CI errors now:
https://github.com/opensciencegrid/topology/pull/1212

PR #1226, opened 2020-06-11
[CIT_CMS_T2] Add login-2/login-3 as htcondor schedds

juztas:


matyasselmeci:
Are "login-2.tier2" and "login-3.tier2" the correct FQDNs? They do not resolve in DNS.

juztas:
They are private ones. I could change that to public ones (same as login-1.tier2) - but I would have more questions:
Is it enough to change ProbeName in 
```
cat /etc/gratia/condor/ProbeConfig | grep ProbeName
    ProbeName="condor:login-1.tier2"
```
or also HTCondor must be on the public? Maybe you can guide me how these parameters are used and we can change it.

djw8605:
The ProbeName needs to match the `FQDN` in the topology resource.  The DNS name doesn't need to actually resolve, but a unique FQDN does help identify the usage.

You could set the FQDN in the topology registration to login-3.tier2.ultralight.org, even though that won't resolve.  Then set the Probe name to "condor:login-3.tier2.ultralight.org", and everything should link together.  HTCondor does not need to be public, and the DNS does not need to resolve.  This is just used for linking the topology registration with the records in GRACC.

juztas:
Ok, Let me change this on local nodes and I will update pull request

juztas:
Quick question, I need to change only:
```
    FQDN: login-1.hep.caltech.edu
    FQDNAliases:
    - login-1.hep.caltech.edu
```
and leave sam_uri as is? or change it also? Not sure where it is used :)

matyasselmeci:
Not sure, but as a side note, you can leave off FQDNAliases if they're the same as the FQDN.

djw8605:
Don't change the existing login node.  We already have records with that FQDN, and therefore we need to keep that FQDN registered in topology.

juztas:
mh.. puppet change will enforce all nodes to use fqdn from my side, so, login-1.tier2 will not be available, nor reporting anymore. Should I set it back to login-1.tier2 and also add new service with fqdn?

djw8605:
Yes. Go ahead with the puppet change.  Change back the topology for the login-1 to the non full host name. And add a new resource entry with the new login-1 full host name.

matyasselmeci:
Looks good to me too. If you bring it up to date with master, we can merge it.

PR #1229, opened 2020-06-15
Fix service type for Caltech submit nodes

brianhlin:
FYI @juztas


juztas:
p.s. This does not change usage reporting to wlcg, correct?

djw8605:
This should be fine.  It should not affect what is sent to the WLCG.

PR #1250, opened 2020-07-06
Replace CHTC origin server with sc-origin2000.chtc.wisc.edu

matyasselmeci:


djw8605:
If it's a CNAME to the other name, why make any changes at all?

matyasselmeci:
sc-origin2000 is the real name, sc-origin will be the CNAME once we switch over. I suppose I could [manually set the FQDN](https://opensciencegrid.org/docs/data/stashcache/install-origin/#manually-setting-the-fqdn-optional) instead, though the DN is still going to change. Would that be better?

djw8605:
The DN doesn't have to change if it's a CNAME, right?  Or it can be a SAN?

matyasselmeci:
We're also switching CAs from InCommon to Let's Encrypt.

PR #1251, opened 2020-07-06
Add ability to tag resources; tag CC* CEs (SOFTWARE-4000)

edquist:


edquist:
Ok, looks like something is not quite happy

```
Validating rgsummary XML schema...

/tmp/rgsummary.xml:44: element Tags: Schemas validity error : 
Element 'Tags': Missing child element(s). Expected is ( Tag ).
```

edquist:
Ok the problem is that empty `<Tags></Tags>` are being emitted for resources with no tags, which they should not be according to the rules I added to rgsummary.xsd (in fab135716063698682477fd216c023478db9f3ea).

edquist:
I think i was barking up the wrong tree.  The `<Tags></Tags>` would always get generated, possibly due to the xml schema itself rather than the way i was adding objects to the tree.  And it turns out that FQDNAliases does in fact also appear as `<FQDNAliases></FQDNAliases>` in the xml when there are no FQDNAliases.  So to get this working i just need to make the schema allow an empty Tags list, and the Tags element will always appear, even if the list is empty.

Looks like the CI is passing now with my fixes.  I will rebase to clean up all the experimental commits.

edquist:
@matyasselmeci, i think this seems to be working now, if you wanna give it a once-over

edquist:
This is deployed on topology-itb:
https://topology-itb.opensciencegrid.org/rgsummary/xml

Seems to be working as expected:

Resource snippet with no tags:
```
                        <Resource>
                                <ID>446</ID>
                                <Name>ATLab testing node</Name>
                                <Active>True</Active>
                                <Disable>False</Disable>
                                <Services>      
                                        <Service>
                                                <ID>1</ID>
                                                <Name>CE</Name> 
                                                <Description>Compute Element</Description>
                                                <Details>       
                                                        <hidden>True</hidden>
                                                </Details>
                                        </Service>              
                                </Services>                     
                                <Tags></Tags>                   
                                <Description>This is test gatekeeper for integrating the OSG stack with a ROCKS platfrom.</Description>
                                <FQDN>rocksvmfe.uabgrid.uab.edu</FQDN>
                                <FQDNAliases></FQDNAliases>
```

Resource snippet with a tag:
```
                        <Resource>
                                <ID>985</ID>
                                <Name>OSG_US_AMNH_ARES</Name>
                                <Active>True</Active>
                                <Disable>False</Disable>
                                <Services>
                                        <Service>
                                                <ID>1</ID>
                                                <Name>CE</Name>
                                                <Description>American Museum of Natural History Hosted CE</Description>
                                                <Details></Details>
                                        </Service>
                                </Services>
                                <Tags>
                                        <Tag>CC*</Tag>
                                </Tags>
                                <Description>This is a Hosted CE for the AMNH Slurm cluster</Description>
                                <FQDN>hosted-ce22.grid.uchicago.edu</FQDN>
                                <FQDNAliases></FQDNAliases>
```

PR #1265, opened 2020-07-20
added IU StashCache

LincolnBryant:


matyasselmeci:
Hi Lincoln, you need to add a DN to the resource, like in this example: https://github.com/opensciencegrid/topology/blob/master/topology/University%20of%20Wisconsin/CHTC/CHTC.yaml#L322.

LincolnBryant:
Added DN @matyasselmeci

PR #1272, opened 2020-07-27
new projects from VSP and PEARC20 trainings

lmichael107:

PR #1273, opened 2020-07-28
Add SWOSU mapping, fix duplicate project ID

brianhlin:
Sorry for jamming two logically different commits into one but I didn't really want to submit them sequentially and wait on CI/merges for each.

Clearly we don't check for duplicate IDs currently but the question is, do we even have to bother? I don't think any of our consumers care about project IDs. It'd be nice if our RF team if they didn't have to worry about updating IDs each time.


matyasselmeci:
How did you notice the duplicate ID?

If we don't want to keep track of renames then it should be OK to change topology to not require an ID in the YAML. We can have topology:
- omit the ID field altogether
- enter 0 or something for all IDs
- generate a random ID (based on the hash of the name or something)

That depends on our consumers of course. The last is the 'safest' in that we will have unique IDs for each project.

brianhlin:
I saw a strange ID change in #1272 and looked through other projects for that ID. I suspect omitting the field altogether is safe enough since the only consumer of project data is possibly @cnweaver and the OSG Connect team.

PR #1276, opened 2020-07-28
Added clas12MC project

rynge:


projects/clas12MC.yaml
```diff
@@ -0,0 +1,11 @@
+Name: clas12MC
```
brianhlin, line 1:
```suggestion
Name: JLab_clas12MC
```

Perhaps? @lmichael107

rynge: This is for an existing project which was not registered

PR #1277, opened 2020-07-29
adding bob sinkovits project

ChristinaLK:


projects/XSEDE_ECSS.yaml
```diff
@@ -0,0 +1,12 @@
+Name: XSEDE_ECSS 
+Description: Support for XSEDE users through ECSS
+Department: San Diego Supercomputing Center
+FieldOfScience: Computer Science
+Organization: University of California, San Diego
+PIName: Robert Sinkovits
+
+ID: '714'
```
brianhlin, line 8:
```suggestion
ID: '715'
```

There was another project PR sitting in our queue so we've got to update the ID. @matyasselmeci and I have been talking about what it'd take to remove this field so hopefully we can avoid this in the future!

brianhlin:
Why the XSEDE prefix if the institution is UCSD?

ChristinaLK:
He seemed to want the project associated w/ his role as the XSEDE ECSS lead. I could change it though, if that makes sense.

brianhlin:
Makes sense to me, just wanted to check!

PR #1280, opened 2020-07-30
NMSU Discovery: retire site resource

nvonwolf:
Retire non-self-hosted CE in favor of our new self-hosted CE.


nvonwolf:
Travis did some wierd merge thing that added another commit.  I'll remove and readd pull request.

PR #1295, opened 2020-08-11
Add new slate resource for cedar

mmascher:
Used naming convention from[1] while trying to maintain cedar as resource group (backward compatibility).

Setting:
            <attr name="GLIDEIN_ResourceName" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="cedar"/>
            <attr name="GLIDEIN_Site" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="cedar"/>

[1]https://docs.google.com/document/d/16YZpcfegXewEf1WeYLvtFIuDfQCv8U0m_TEgtw8zmJ8/edit#heading=h.ps1ea8hqqem3

### DIFF
--- a/topology/University of Connecticut/UConn - cedar/cedar.yaml
+++ b/topology/University of Connecticut/UConn - cedar/cedar.yaml
@@ -34,3 +34,33 @@ Resources:
           hidden: false
     Tags:
       - CC*
+  UConn-CEDAR:
+    ID: 1073 
+    Description: The Hosted CE serving UConn-cedar
+    ContactLists:
+      Administrative Contact:
+        Primary:
+          Name: Marco Mascheroni
+          ID: 030408ab932e143859b5f97a2d1c9e30ba2a9f0d
+        Secondary:
+          Name: Jeffrey Michael Dost
+          ID: 3a8eb6436a8b78ca50f7e93bb2a4d1f0141212ba
+      Security Contact:
+        Primary:
+          Name: Marco Mascheroni
+          ID: 030408ab932e143859b5f97a2d1c9e30ba2a9f0d
+        Secondary:
+          Name: Jeffrey Michael Dost
+          ID: 3a8eb6436a8b78ca50f7e93bb2a4d1f0141212ba
+      Site Contact:
+        Primary:
+          Name: Richard T. Jones
+          ID: 650bf84a5cf49caf504fed22ce07b98580c6fa12
+    FQDN: hosted-ce28.opensciencegrid.org
+    Services:
+      CE:
+        Description: UConn-CEDAR Hosted CE
+        Details:
+          hidden: false
+    Tags:
+      - CC*


edquist:
@mmascher, do you know what the factory is setting the attribute `GLIDEIN_Resource` to, for payload jobs for this new CE resource?

mmascher:
> @mmascher, do you know what the factory is setting the attribute `GLIDEIN_Resource` to, for payload jobs for this new CE resource?

@edquist This is what we had in the old entry, the one for `hosted-ce28.grid.uchicago.edu`:

```
            <attr name="GLIDEIN_ResourceName" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="UConn - cedar"/>
            <attr name="GLIDEIN_Site" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="GLUEX_US_UConn_cedar"/>
```

For this new CE, `hosted-ce28.opensciencegrid.org`, I am setting (after crosschecking with @jdost321):

```
            <attr name="GLIDEIN_ResourceName" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="cedar"/>
            <attr name="GLIDEIN_Site" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="cedar"/>
```

brianhlin:
@jdost321 To avoid future issues where payload records aren't associated with the resource like with the WSU GPU records [1], can we change the convention for new Hosted CEs so that `GLIDEIN_ResourceName` and the topology resource name match?

Also @mmascher, according to the doc you linked, the resource group should be "UConn-cedar". Could you expand on the backwards compatibility reasoning that you mentioned above?

[1] https://opensciencegrid.atlassian.net/browse/SOFTWARE-4200

mmascher:
> @jdost321 To avoid future issues where payload records aren't associated with the resource like with the WSU GPU records [1], can we change the convention for new Hosted CEs so that `GLIDEIN_ResourceName` and the topology resource name match?
> 
> Also @mmascher, according to the doc you linked, the resource group should be "UConn-cedar". Could you expand on the backwards compatibility reasoning that you mentioned above?
> 
> [1] https://opensciencegrid.atlassian.net/browse/SOFTWARE-4200

@brianhlin The issue is that the yaml file here:
https://github.com/opensciencegrid/topology/tree/master/topology/University%20of%20Connecticut/UConn%20-%20cedar
is called cedar.yaml , and IIUC what @jdost321 told me, that has to be the GLIDEIN_ResourceName

jdost321:
For historical reasons factory GLIDEIN_ResourceName must be the Resource Group name. I want that consistency across all factory entries, not do something different for hosted ces. After we're done with the hosted CE migration I want to take a pass to fix all the topolgy issues to match our new convention in the doc. Long term, "Resrouce Group" should be the same as Resource Name" for this and it will be UConn-cedar.  The yaml file is called cedar.yaml so we have to match that to the resource group, so for not until we fix topology GLIDEIN_ResourceName  should be "cedar". I don't know the implications of how gracc handles changing this.. will it look like a new site? etc..

bbockelm:
@jdost321:

> For historical reasons factory GLIDEIN_ResourceName must be the Resource Group name

Where is this written down?  There's strong reasoning from the accounting side that this is not the right way to go.

Unless we have something written up here, discussed, and approved: while I am sure we have historical reasons, it's not clear that's what we should do going forward.

jdost321:
@bbockelm its not written anywhere but its come up a few times now, GLIDEIN_ResourceName == topology "Resource Group" and this is important in other contexts, especially as the public name WLCG sees in the APEL accounting, it is what used to be exported as the "site" name in BDII, etc.. 

If we just start doing something different now i'm worried about how it could break other external consumers.

brianhlin:
>  especially as the public name WLCG sees in the APEL accounting,

As I understand it, `GLIDEIN_ResourceName` is only relevant for payload records and I didn't think we sent along payload records to APEL [1]. @djw8605?

> it is what used to be exported as the "site" name in BDII

OSG sites, including US-LHC, use the central collector for information services. Maybe we can limit the GLIDEIN_ResourceName = Topology resource name for OSG sites?

[1] https://docs.google.com/document/d/1tHMaYnU5mHaBPmias7J6QWCuL9IpLF5PXtRlWuDannU/edit?ts=5beb2bc3#heading=h.3q7x7nkk4azj

jdost321:
@blin 

> OSG sites, including US-LHC, use the central collector for information services. Maybe we can limit the GLIDEIN_ResourceName = Topology resource name for OSG sites?

This is exactly my fear. OSG entries are only 1/3 of the entries in the factory. from factory operator context GLIDEIN_ResourceName was always intended to mean "the public name in the information system". for EGI where we still use BDII that's the BDII name. I don't want to teach factory operators do this for Europe and that for OSG.. they will never remember the exceptions to the rules.

In fact I thought we had discussed longer term we should really rename GLIDEIN_ResourceName to GLIDEIN_ResourceGroup in the entries to be more consistent.

brianhlin:
How about we just start with the OSG Hosted CEs, then? That way we can have complete payload records and the only folks that need to understand the exception are the Hosted CE operators. You could add comments to the factory config to explain the exception, too.

> In fact I thought we had discussed longer term we should really rename GLIDEIN_ResourceName to GLIDEIN_ResourceGroup in the entries to be more consistent.

Last I remember, we discussed something like `GLIDEIN_OSG_Resource` so that we wouldn't have to worry about a transitioning `GLIDEIN_Resource` users to a new attribute. But we'd have to decide on this attribute, make software changes to look for it, and you'd have to train operators to start populating it for OSG sites.

bbockelm:
Hi -

I like all the exchange of ideas here, but let's turn back to the PR and get the SLATE CE rolling.

@brianhlin @jdost321 @djw8605 - can you write up a word document I can bring to the ET next week?  I would welcome out of the box ideas too.  I wouldn't be opposed to a CMS frontend validation script that overrides the factory settings with legacy settings: if we're not gathering payload data from CMS, I don't see a need to have consistent naming between the pools.

Brian

djw8605:
Payload data is not sent to APEL.  For the most part, we do not receive any CMS payload data (except CMS connect).  Nothing in the payload will affect what data is sent to APEL.

brianhlin:
> @brianhlin The issue is that the yaml file here:
https://github.com/opensciencegrid/topology/tree/master/topology/University%20of%20Connecticut/UConn%20-%20cedar
is called cedar.yaml , and IIUC what @jdost321 told me, that has to be the GLIDEIN_ResourceName

@mmascher Until we sort out the discussion above we have two requirements:

1. From the factory: `GLIDEIN_ResourceName` equals the topology resource group
1. From the GRACC so that payload records can be associated with a resource: topology resource name equals `GLIDEIN_ResourceName`

So if you create a new resource group file `UConn-CEDAR.yaml` and put the `UConn-CEDAR` resource in it, I think we could get this registered.

jdost321:
@djw8605 isn't this PR still valid? for payload hours doesn't the check compare GLIDEIN_ResourceName against both resource name and resource group names?

jdost321:
@blin we don't wnat it to look like a new site, it is to replace the old CE thats why Marco just added the resource into the existing resource group. the old one will be turned off eventually.

brianhlin:
> @djw8605 isn't this PR still valid? for payload hours doesn't the check compare GLIDEIN_ResourceName against both resource name and resource group names?

Yes but if they match on the group, then we can't determine which resource the record should be associated with when there's more than one CE, which can make troubleshooting issues more difficult (which CE are bad records coming from?) and strange differences in accounting visualizations like with the WSU GPU dashboards.

> @brianhlin we don't wnat it to look like a new site, it is to replace the old CE thats why Marco just added the resource into the existing resource group. the old one will be turned off eventually.

It'll still be the same Topology site. My recommendation is the only way I can think of to satisfy the factory and accounting requirements until we decide on a solution to the discussion above.

jdost321:
I discussed more with @brianhlin (sorry for the noise blin whoever you are!) and we agreed in the interest of not stalling the hosted CE slate transitions, we'll make the factory ops exception to the rule for hosted ces only for now and set GLIDEIN_ResourceName to resource name "UConn-CEDAR" in this case.

PR #1304, opened 2020-08-14
add UWM-NEMO SLATE ce

jdost321:


topology/University of Wisconsin Milwaukee/UWM - NEMO/NEMO.yaml
```diff
@@ -29,3 +29,28 @@ Resources:
         Description: UWM-NEMO Hosted CE
         Details:
           hidden: false
+  UWM-NEMO:
+    ID: 1079
+    Disable: false
```
brianhlin, line 34:
```suggestion
    Active: true
```

Presumably this CE is ready to go?

jdost321:
Oops wasn't paying close attention to the flags.. just copy / pasted the old one. does that mean the old CE was never "active" ?

matyasselmeci:
Active is true by default.

PR #1307, opened 2020-08-17
add TACC-Stampede2 SLATE ce

mmascher:


brianhlin:
@edquist since you're looking into the topology/factory/GRACC interactions, I'm requesting your review on this to make sure that records are generated as expected for your dashboards

brianhlin:
@edquist please make this the top priority when you get back

edquist:
Sorry I must have missed this when it was originally assigned to me.  Since this is not tagged for CC*, i am not so concerned about whether the resource name matches the glidein_resourcename configured in the factory (it doesn't reall matter if gracc matches payload records for this CE at the resource level or not).  Otherwise CI passed after bringing up to date with master and i don't see any obvious problems.

edquist (approved):

PR #1314, opened 2020-08-19
register ASU-DELL_M420 SLATE hosted ce

jdost321:


bbockelm:
We don't need to revisit this PR, but it might be strategic to stick to DNS-compatible subdomains in the future...

jdost321:
@bbockelm I don't understand what you're referring to.

brianhlin:
@jdost321 Brian's suggesting that resource names should only conform to a format that would be usable in DNS: https://stackoverflow.com/questions/3523028/valid-characters-of-a-hostname/3523068#3523068

That way, we could set up DNS entries based on the resource name, e.g. `amnh-hel.opensciencegrid.org`

jdost321:
Ah, didn't realize he meant resource names.. well I wish we thought about that sooner
`condor_status -pool collector.opensciencegrid.org:9619 -schedd -af name -af osg_resource | grep hosted.*opensciencegrid.org
hosted-ce04.opensciencegrid.org GSU-ACORE
hosted-ce07.opensciencegrid.org FSU-HNPGRID
hosted-ce13.opensciencegrid.org PSC-Bridges
hosted-ce18.opensciencegrid.org USF-SC
hosted-ce20.opensciencegrid.org ASU-DELL_M240
hosted-ce22.opensciencegrid.org AMNH-ARES
hosted-ce28.opensciencegrid.org UConn-CEDAR
hosted-ce30.opensciencegrid.org TCNJ-ELSA
hosted-ce32.opensciencegrid.org UWM-NEMO
hosted-ce34.opensciencegrid.org UCSD-Comet
hosted-ce35.opensciencegrid.org ND-CAML_gpu
hosted-ce36.opensciencegrid.org AMNH-HEL
hosted-ce37.opensciencegrid.org Clarkson-ACRES
hosted-ce38.opensciencegrid.org SDSC-Triton_Stratus`

What are the implication of changing resource names after GRACC already collected stats?

brianhlin:
Let's not worry about the existing CEs for now, this was something that was brought up recently and I can work with you on how we should register new CEs moving forward. FWIW, @bbockelm this will slow down new CE spin-up as long as we need Chicago to register PTRs.

jdost321:
@bbockelm @brianhlin We should definitely continue this discussion after the migration, but I like having the added level of indirection with hostnames like hosted-ce30.opensciencegrid.org. The DNS record is created one time, and I can reuse them if one CE gets decommissioned and we want to use it for a new one, etc.

matyasselmeci:
Just to be clear, @bbockelm, you're just talking about establishing a convention for resource names, not suggesting that they also match the hostname?

bbockelm:
I think calling it a "convention" might be a bit strong - I was just going to set a limitation so we, in the future, could make the hosted CE names match to make them more readable.

matyasselmeci:
Sure, that sounds fine.

I know in the past we've talked about establishing a 1-1 mapping between FQDN and resource name, but we ended up finding too many duplicates...

PR #1321, opened 2020-08-24
Correct Thomas Britton ID

matyasselmeci:


matyasselmeci:
The CI for the contact repo caught this. I don't think the CI for the topology repo has the key to access the contact repo with, so the check is only in the contact repo.

Since the check only needs Name and ID, which are public, the topology CI can start up the webapp and query /miscuser/xml for that info...

PR #1322, opened 2020-08-24
Add ability to filter contacts and associated VOs/resources on email

brianhlin:
Security wanted a list of resources/VOs associated with specific emails and this was the easiest way to get it done


src/topology_utils.py
```diff
@@ -358,4 +358,12 @@ def filter_contacts(args, results):
             else:
                 results[name] = contact_list
 
+    if getattr(args, 'contact_emails', None):
+        for name in results.keys():
+            contact_list = [contact for contact in results[name] if contact['Email'] in args.contact_emails]
+            if contact_list == []:
```
matyasselmeci, line 364:
```suggestion
            if not contact_list:
```

PR #1332, opened 2020-08-28
Create Baylor-Kodiak_downtime.yaml

hatakeyamak:
Add downtime for Baylor-Kodiak due to testing.


hatakeyamak:
Thanks. BTW, how will this downtime information registered here will be used?

brianhlin:
Downtimes are used by CMS so that dedicated sites aren't docked for their downtimes. The OSG doesn't currently make use of the downtime information but we hope to make better use of it in our job submission and monitoring tools.

hatakeyamak:
I see. Thank you for the clarification.

PR #1373, opened 2020-10-05
Update OSG-Staff project to PI: Frank

djw8605:
I'm open to suggestions on who should be the PI.  But it certainly should not be Chander.


brianhlin (approved):
Fine by me

PR #1375, opened 2020-10-05
Add Semir as a contact for the Georgia Tech PACE resources

matyasselmeci:
[FD #65777](https://support.opensciencegrid.org/a/tickets/65777)

Also fix duplicate 'tertiary' contacts (the contact lists are dicts so the ranks need to be different)


brianhlin (approved):
It pains me that we have to do it this way but LGTM

PR #1378, opened 2020-10-06
Add CHTC tiger stashcache cache

matyasselmeci:


brianhlin (commented):
You'll probably want to update `virtual-organizations/GLOW.yaml`, too!

brianhlin (approved):
Pre-approving

matyasselmeci:
> You'll probably want to update `virtual-organizations/GLOW.yaml`, too!

Nope, that's just if I'm adding an origin. GLOW (and other VOs with only public data) have AllowedCaches: ANY, so the new cache will happily accept their data.

PR #1380, opened 2020-10-07
add WSU SLATE Hosted CE

jdost321:


brianhlin (commented):
@jdost321 I noticed that your corresponding factory commit doesn't appear to include the `GLIDEIN_ResourceName`, was that intentionally left blank?

jdost321:
Yes, I didn't have to override it, it got the right thing from the osg collector:
```<attr name="GLIDEIN_ResourceName" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="string" value="WSU-GRID"/>```

brianhlin (approved):
Interesting, thanks!

brianhlin:
@mmascher does the autoconf script set the `GLIDEIN_ResourceName` to the `OSG_Resource` as advertised in the schedd ad from the central collector?

PR #1381, opened 2020-10-07
Update criteria for extracting gfactory entries in compare-factory-config.py

Caffretro:
This script should only compare active entries between Topology and gfactory.

### DIFF
--- a/bin/compare-factory-config.py
+++ b/bin/compare-factory-config.py
@@ -149,7 +149,7 @@ def run(argv):
     for xml in gfactory:
         get_gfactory_data(gfactoryDB, xml)
 
-    # comparing gfactory with Topology resources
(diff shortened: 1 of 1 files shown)


bin/compare-factory-config.py
```diff
@@ -81,13 +81,17 @@ def get_gfactory_data(gfactoryDB, filename):
         root = tree.getroot()
         # insert Names in Topology database into
         for entry in root.findall('entries/entry'):
+            if entry.get('enabled') == 'False':
+                # only compairing active gfactory entries
```
brianhlin, line 85:
```suggestion
                # only comparing active gfactory entries
```

brianhlin (approved):
A typo but otherwise LGTM!

PR #1385, opened 2020-10-09
Rename JLab resource group and add a JLab CE + Squid

brianhlin:
So that it's not GlueX-specific

Includes #1384


edquist (approved):
Changes look fine.

A procedural note: you are including the PR to add JLab resources, so you might want to include that in your PR title.  Otherwise (as is the case for me), it misleadingly suggests that this PR is only about renaming the resource group.

Relatedly, the description mentions "Closes #1384", which will have the effect of marking #1384 as Closed rather than Merged.  Normally a PR is marked Merged when another PR that includes it gets merged.  Personally i feel like that's the better way to do it; so i'd suggest changing the "Closes #1384" in the description to something like "Includes #1384", which is just informational and doesn't have any undesired semantic consequences.

brianhlin:
Noted, updated PR title + PR body text

PR #1388, opened 2020-10-12
Drop TimC from the MIS OASIS managers per his request

brianhlin:


matyasselmeci (approved):
I don't think any of those certs are still valid, either.

PR #1393, opened 2020-10-13
Fix Cedar: move from UConn to new Compute Canada

mmascher:
See https://opensciencegrid.atlassian.net/browse/OPS-31

Richard Jones asked us move Cedar from the wrong UConn facility to a new Compute Canada one.

We also need to preserve the current accounting. NB The resource is still `hosted-ce28.opensciencegrid.org`. We should also have some records under `hosted-ce28.grid.uchicago.edu` that are pre SLATE translation.


mmascher:
@djw8605 @brianhlin @jdost321 
Is this still pending because of the failing check? How can I know what is wrong?

matyasselmeci (approved):

matyasselmeci:
Hi Marco, sorry we lost track of this. You can see the error by looking at the build 'details' for the check, though the output is sometimes hard to parse. In this case, the error was a missing GroupID; I filled it in for you.

Before merging, I'd like to check with GRACC folks how marking a resource group "Disabled" affects accounting.

matyasselmeci (approved):
Derek sent me the query GRACC is using; this should be fine.

PR #1395, opened 2020-10-14
Adding dependencies to the requirements for compare-factory-config.py

Caffretro:
This commit resolves a non-standard import for `GitPython`, along with a logic adjustment and a minor comment corrections with the script.


brianhlin (approved):
LGTM!

PR #1396, opened 2020-10-14
allow CILogonID format for ContactIDs (SOFTWARE-4310)

edquist:
... And consider a CILogonID to be a "known contact ID", also, if it is associated with a contact.

### DIFF
--- a/src/tests/verify_resources.py
+++ b/src/tests/verify_resources.py
@@ -49,6 +49,9 @@ def get_vo_names():
     return set( re.search(r'/([^/]+)\.yaml$', path).group(1) for path in
                 glob.glob(_topdir + "/virtual-organizations/*.yaml") )
 
+def contact_id_ok(ID):
+    return re.search(r'^[0-9a-f]{40}$', ID) or re.search(r'^OSG\d+$', ID)
+
 def load_yamlfile(fn):
     with open(fn) as f:
         try:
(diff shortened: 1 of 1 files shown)


src/tests/verify_resources.py
```diff
@@ -69,7 +72,12 @@ def get_contacts():
     txt = urlopen(_contacts_url).read()
     xmltree = et.fromstring(txt)
     users = xmltree.findall('User')
-    return dict(map(user_id_name, users))
+    d = dict(map(user_id_name, users))
+    for u in users:
+        cilogonid = u.find('CILogonID')
+        if cilogonid:
+            d[cilogonid] = u.find('FullName').text
```
matyasselmeci, line 79: what does this stuff do?

edquist: the `contacts` dict returned by this function was previously a map of ContactID -> FullName.  This block adds CILogonID -> FullName for the contacts that have a CILogonID, so that both ContactID and CILogonID (if any) will be keys in this contacts dict that map to the FullName.

Given that that was unclear enough for you to ask, i can throw this into a separate, named function.

matyasselmeci (approved):
Looks good. Bring it up to date with master and if the tests pass, feel free to merge.

src/tests/verify_resources.py
edquist: @matyasselmeci, does this help make it clearer?

matyasselmeci: Yeah, it helps.

PR #1399, opened 2020-10-15
Update the FQDN of osg-itb-ce1

brianhlin:


edquist (approved):
i think i missed the memo about this .io

brianhlin:
> i think i missed the memo about this .io

Welcome to the future where Tiger moves fast :D

PR #1401, opened 2020-10-15
Fixing bugs and renaming variables for compare-factory-config.py

Caffretro:
The last version was not outputting correct entry names for factory names that does not match any entity in Topology. CamelCase variables were also renamed to follow PEP8 style guide.


brianhlin (approved):
LGTM! One thing to note for the future: it's a good idea to separate out commits into logical sets of changes, especially when you have changes that change the behavior of the script and others are just superficial in nature.

brianhlin:
Overriding CI failure due to issue with LIGO's LDAP

PR #1403, opened 2020-10-19
Remove accidentally added ara.osgstorage.org

DrDaveD:
This entry for ara.osgstorage.org was probably accidentally copied from some other entry and was never officially requested.  It has been putting warnings into log files on the oasis machine ever since which have been largely ignored, but the line should be removed.

@dsschult


djw8605 (approved):
I always like removing things!

PR #1410, opened 2020-10-23
New project; correcting another

lmichael107:


projects/RIT_ResearchComputing.yaml
```diff
@@ -0,0 +1,12 @@
+Name: RIT_ResearchComputing
+Description: Research Computing staff at the Rochester Institute of Technology (RIT).
+Department: RIT Research Computing
+FieldOfScience: Computer and Information Sciences
+Organization: Rochester Institute of Technology
+PIName: Kirk Anne
+
+ID: '727'
```
brianhlin, line 8:
```suggestion
ID: '740'
```

727 is already taken by `UNL_Howard`. We should really eliminate this ID requirement for you guys...

Objections @djw8605 @matyasselmeci ?

brianhlin (approved):
One minor fix

projects/RIT_ResearchComputing.yaml
matyasselmeci: IDs can provide rename tracking if GRACC uses them; if it doesn't, we can just key off of the name.

PR #1412, opened 2020-10-24
Added resource matchup suggestion for compare-factory-config.py (SOFTWARE-4327)

Caffretro:
## 10/24/2020
- Added resource matchup suggestion function for factory entries
- Formatted topology tree dumping
- Formatted output
- Fixed bug that ignores factory entries in yml files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ src/webapp/*.pyc
 src/config.py
 */.DS_Store
 .vscode
-bin/country-names-unify.py
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ src/webapp/*.pyc
 src/config.py
 */.DS_Store
 .vscode
+bin/country-names-unify.py
--- a/bin/compare-factory-config.py
+++ b/bin/compare-factory-config.py
@@ -135,9 +135,9 @@ def remove_readonly(func, path, _):
     func(path)
 
 
-def find_suggestion(gatekeeper, topology_DB):
+def find_suggestion(gatekeeper, topology_DB_resources):
     # return the corresponding resource if a gatekeeper can find a fqdn match
-    for resource, fqdn in topology_DB['resources']:
+    for resource, fqdn in topology_DB_resources:
         if (gatekeeper == fqdn):
             return resource
     return None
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ src/webapp/*.pyc
 src/config.py
 */.DS_Store
 .vscode
+bin/country-names-unify.py
```
brianhlin, line 9: I think you missed a commit to remove

bin/compare-factory-config.py
```diff
@@ -176,27 +192,33 @@ def run(argv):
...
-    nonmatch_all_entries = find_non_topology_matches(gfactory_DB, topology_DB)
+    nonmatch_all_entries = find_non_topology_matches(
+        gfactory_DB, topology_DB, resources)
 
     # output formatted results
+    print(f'\nOutput format: <GLIDEIN_ResourceName>:<corresponding factory entry name>:<suggestion for a resource match>\n')
     print(f'\nFactory entries that match a Topology entity other than a resource: \n')
     for x in match_nonresource_entries:
-        print(f'- {x[0]}: {x[1]}')
+        print(f'{x[0]}:{x[1]}:{x[2]}')
```
brianhlin, line 218:
```suggestion
        print(f'{x[0]},{x[1]},{x[2]}')
```

May as well make the output CSV-like if we're going to be dumping things into a spreadsheet!

bin/compare-factory-config.py
```diff
@@ -176,27 +192,33 @@ def run(argv):
...
     print(f'\nFactory entries that do not match any entity in Topology: \n')
-    for x in sorted(set(nonmatch_all_entries)):
-        print(f'- {x}')
+    for x in nonmatch_all_entries:
+        print(f'{x[0]}:{x[1]}:{x[2]}')
```
brianhlin, line 221:
```suggestion
        print(f'{x[0]},{x[1]},{x[2]}')
```

bin/compare-factory-config.py
```diff
@@ -176,27 +192,33 @@ def run(argv):
...
     match_nonresource_entries = find_non_resource_matches(
-        gfactory_DB, topology_DB)
+        gfactory_DB, topology_DB, resources)
     # Entry names corresponding to GLIDEIN_ResourceNames above
-    nonmatch_all_entries = find_non_topology_matches(gfactory_DB, topology_DB)
+    nonmatch_all_entries = find_non_topology_matches(
+        gfactory_DB, topology_DB, resources)
 
     # output formatted results
+    print(f'\nOutput format: <GLIDEIN_ResourceName>:<corresponding factory entry name>:<suggestion for a resource match>\n')
```
brianhlin, line 215: Is this the actual format? If so, could you switch the first and second columns?

brianhlin (approved):
Some minor nitpicks but otherwise LGTM

bin/compare-factory-config.py
Caffretro: Oh I was using columns for delimiters and parsing them locally, but yeah this is better

bin/compare-factory-config.py
Caffretro: that's a mistake here, output was correct and description is wrong

brianhlin:
Thanks for the changes! Could you populate the spreadsheet with the output as soon when you get the chance?

PR #1431, opened 2020-11-03
Drop CHTC_STASHCACHE_ORIGIN from IceCube's AllowedOrigins list

matyasselmeci:
per FD #65990; IceCube now has their own origin.


brianhlin (approved):
LGTM

PR #1432, opened 2020-11-04
Handling errors at parsing malformed factory entries (SOFTWARE-4336)

Caffretro:
- Added error handling code for troubleshooting malformed/incomplete factory entries
- Updated topology query so that inactive entries are also in consideration

### DIFF
--- a/bin/compare-factory-config.py
+++ b/bin/compare-factory-config.py
@@ -132,7 +132,6 @@ def get_gfactory_data(gfactory_DB, filename):
                     # and problematic entry so reviews can be requested
                     print('Error: ', filename, 'is not validly structured:')
                     print(f"Entry name: {entry_name}, content: {config}\n")
-                    continue
                 except Exception as e:
                     print('Error:', e)
                     continue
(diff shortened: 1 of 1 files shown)


bin/compare-factory-config.py
```diff
@@ -116,11 +116,25 @@ def get_gfactory_data(gfactory_DB, filename):
...
+                    # skip malformed entries
+                    # Since there exists some entries in gfactory/autoconf that 
+                    # work without the key or value we want, skip them
+                    continue
+                except TypeError as e:
+                    # as some yml files might be incomplete, print out the filename
+                    # and problematic entry so reviews can be requested
+                    print('Error: ', filename, 'is not validly structured:')
+                    print(f"Entry name: {entry_name}, content: {config}\n")
+                    continue
```
brianhlin, line 135:
```suggestion
```

`continue` is unnecessary here since you're at the bottom of the loop anyway

brianhlin (approved):
Pre-approving because apart from one nitpick it LGTM

PR #1433, opened 2020-11-04
update Derek's DN

edquist:
https://support.opensciencegrid.org/a/tickets/65978

I know @brianhlin said it didn't really matter, but derek requested it, so here it is!


brianhlin (approved):
:roll_eyes:

PR #1435, opened 2020-11-05
Send security emails from the OSG security list (SOFTWARE-4349)

brianhlin:


matyasselmeci:
You should set the Reply-To as well (unless you want people to reply to the security list).

brianhlin:
@mtstanfield what would you like the To field to be set to for security announcements? Previously this was `help@`. And do you want a different Reply-To?

mtstanfield:
setting them both to security@ should be fine, replies directly to that list are what we want.

bin/osg-notify
```diff
@@ -71,14 +77,14 @@ def generateFullMessage(subject, to, from_name, message, sign=True, keyid=None):
         msg.attach(signmsg)
         msg['Subject'] = subject
         msg['Bcc'] = ", ".join(to)
-        msg['To'] = '"Open Science Grid Helpdesk" <help@opensciencegrid.org>'
-        msg['From'] = '"%s" <help@opensciencegrid.org>' % from_name
+        msg['To'] = '"%s" <%s>' % (to_name, from_addr)
```
matyasselmeci, line 80:
```suggestion
        msg['To'] = '"%s" <%s>' % (to_name, to_addr)
```
and then add `to_addr = from_addr` earlier in the code. I know we're reusing the address but anybody looking at it would think it's a bug...

bin/osg-notify
```diff
@@ -71,14 +77,14 @@ def generateFullMessage(subject, to, from_name, message, sign=True, keyid=None):
...
+        msg['From'] = '"%s" <%s>' % (from_name, from_addr)
         return msg.as_string(unixfrom=True)
     else:
         basemsg['Subject'] = subject
         basemsg['Bcc'] = ", ".join(to)
-        basemsg['To'] = '"Open Science Grid Helpdesk" <help@opensciencegrid.org>'
-        basemsg['From'] = '"%s" <help@opensciencegrid.org>' % from_name
+        msg['To'] = '"%s" <%s>' % (to_name, from_addr)
```
matyasselmeci, line 86:
```suggestion
        msg['To'] = '"%s" <%s>' % (to_name, to_addr)
```

matyasselmeci (approved):
Looks good except for a style change.

PR #1441, opened 2020-11-09
Add GitHub Actions (SOFTWARE-4342)

brianhlin:
This PR doesn't get rid of Travis yet since the webhook currently depends on Travis and I didn't want to touch that yet.

Success here: https://github.com/brianhlin/topology/actions

### DIFF
--- a/src/tests/test_verify_schema.sh
+++ b/src/tests/test_verify_schema.sh
@@ -1,6 +1,10 @@
 #!/bin/bash -e
 
-PYTHONPATH="$PYTHONPATH:$TRAVIS_BUILD_DIR/src"
+# Use the Travis build dir if specified, otherwise assume that the
+# repo is unpacked in the current working dir
+REPO_ROOT_DIR=${TRAVIS_BUILD_DIR:-.}
+
+PYTHONPATH="$PYTHONPATH:$REPO_ROOT_DIR/src"
 export PYTHONPATH
 
 function verify_xml {
@@ -8,7 +12,7 @@ function verify_xml {
     xml="$1"
     type="$2"
     echo "Validating $type XML schema..."
-    xmllint --noout --schema "$TRAVIS_BUILD_DIR/src/schema/$type.xsd" $xml
+    xmllint --noout --schema "$REPO_ROOT_DIR/src/schema/$type.xsd" $xml
 }
 
 if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
@@ -29,17 +33,17 @@ for DATA_TYPE in miscproject vosummary rgsummary; do
 
     case $DATA_TYPE in
         miscproject)
-            YAML_DIR="$TRAVIS_BUILD_DIR/projects"
+            YAML_DIR="$REPO_ROOT_DIR/projects"
             READER=src/webapp/project_reader.py
             READER_ARGS="$YAML_DIR $CONVERTED_XML"
             ;;
         rgsummary)
-            YAML_DIR="$TRAVIS_BUILD_DIR/topology"
+            YAML_DIR="$REPO_ROOT_DIR/topology"
             READER=src/webapp/rg_reader.py
             READER_ARGS="$YAML_DIR $CONVERTED_XML /tmp/rgdowntime.xml"
             ;;
         vosummary)
-            YAML_DIR="$TRAVIS_BUILD_DIR/virtual-organizations"
+            YAML_DIR="$REPO_ROOT_DIR/virtual-organizations"
             READER=src/webapp/vo_reader.py
             READER_ARGS="$YAML_DIR $CONVERTED_XML"
             ;;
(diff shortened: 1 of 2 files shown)


src/tests/test_verify_schema.sh
```diff
@@ -1,23 +1,44 @@
...
-    openssl aes-256-cbc -K $encrypted_457175ef53a3_key -iv $encrypted_457175ef53a3_iv -in src/tests/contacts.enc -out contacts -d
+    touch contacts
     chmod 600 contacts
+
+    if [[ $GH_EVENT == 'push' ]]; then
+        echo "$CONTACT_DB_KEY" > contacts
+    else
+        openssl aes-256-cbc \
+                -K $encrypted_457175ef53a3_key \
+                -iv $encrypted_457175ef53a3_iv \
```
matyasselmeci, line 36: These variables are created by Travis (for the secrets). Does GHA do the same thing?

brianhlin: Not quite. For GHA, you can store the secrets in GitHub and access them directly (https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets):

- In the jobs and steps
- In scripts as long as you specify the env in the step (https://github.com/opensciencegrid/topology/pull/1441/files#diff-c4078cc3c976edd304cd31c0b08976d09c6af057bad110660e18ec1b856d8db1R23 and https://github.com/opensciencegrid/topology/pull/1441/files#diff-051eaf796f103706c605e10ddc25ecc7f628058cd6c52c7f362ec90d4ca537f1R32)

matyasselmeci: I see. Also, I just noticed that this line only runs for travis builds so I think you're good.

matyasselmeci (approved):

src/tests/test_verify_schema.sh
```diff
@@ -1,23 +1,44 @@
...
     type="$2"
     echo "Validating $type XML schema..."
-    xmllint --noout --schema "$TRAVIS_BUILD_DIR/src/schema/$type.xsd" $xml
+    xmllint --noout --schema "$REPO_ROOT_DIR/src/schema/$type.xsd" $xml
 }
 
-if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+if [[ $TRAVIS_PULL_REQUEST == "false" ||
+      $GH_EVENT == 'push' &&
+      $GITHUB_REPOSITORY == 'opensciencegrid/topology' ]]; then
```
edquist, line 20: When you mix `||` and `&&`, it's technically unambiguous (it's left-to-right since they have equal operator precedence), but it's a little less obvious which grouping you intended when i read it.

I'd suggest making it explicit, if this is the grouping you intended:

```suggestion
if [[ $TRAVIS_PULL_REQUEST == "false" || $GH_EVENT == 'push' ]] &&
   [[ $GITHUB_REPOSITORY == 'opensciencegrid/topology' ]]; then
```

(Or if you prefer, you can add parens inside the `[[ ... ]]`)

edquist: (I'd also normally point out that `[[ $x == "string" ]]` is can equivalently be written `[[ $x = string ]]`, but that's pretty minor.)

src/tests/test_verify_schema.sh
```diff
@@ -48,7 +69,9 @@ for DATA_TYPE in miscproject vosummary rgsummary; do
     # Resource group and VO readers should use the contact info if we have
     # access to the SSH keys for the contacts repo
     if [[ $DATA_TYPE == 'vosummary' ]] || [[ $DATA_TYPE == 'rgsummary' ]]; then
-        if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+        if [[ $TRAVIS_PULL_REQUEST == "false" ||
+              $GH_EVENT == 'push' &&
+              $GITHUB_REPOSITORY == 'opensciencegrid/topology' ]]; then
```
edquist, line 74: Same here, if this is what you meant

```suggestion
        if [[ $TRAVIS_PULL_REQUEST == "false" || $GH_EVENT == 'push' ]] &&
           [[ $GITHUB_REPOSITORY == 'opensciencegrid/topology' ]]; then
```

src/tests/test_verify_schema.sh
```diff
@@ -1,23 +1,44 @@
...
 }
 
-if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+if [[ $TRAVIS_PULL_REQUEST == "false" ||
+      $GH_EVENT == 'push' &&
+      $GITHUB_REPOSITORY == 'opensciencegrid/topology' ]]; then
+    # Ensure that the .ssh dir exists
+    mkdir ~/.ssh
+    chmod 0700 ~/.ssh
     cat >> ~/.ssh/known_hosts <<EOF
```
edquist, line 24: not sure what the default umask is in this context - if it's `0022` you should be fine, but if it's `0002` i think you need to ensure `~/.ssh/known_hosts` is not group-writable.

edquist (commented):
see comments

src/tests/test_verify_schema.sh
brianhlin: Good call. Fixed.

PR #1442, opened 2020-11-11
Make sure to update the Ubuntu package list before install

brianhlin:


.github/workflows/validate-data.yml
```diff
@@ -12,7 +12,9 @@ jobs:
         with:
           python-version: 3.6
       - name: Install packages
-        run: sudo apt-get install libxml2-utils apache2-dev libldap2-dev libsasl2-dev libyaml-dev
+        run: |
+          sudo apt-get update
```
matyasselmeci, line 16:
```suggestion
          sudo apt-get update && \
```
or is this already run with set -e?

matyasselmeci (approved):
LGTM; one comment, feel free to implement or not.

.github/workflows/validate-data.yml
brianhlin: Yup, all GHA runs are! https://github.com/opensciencegrid/topology/runs/1385808985?check_suite_focus=true#step:4:3

brianhlin:
Travis failed due to a transient (?) issue with contacting LIGO's LDAP. This doesn't seem to be the case for GHA so I wonder if LIGO is throttling Travis. Overriding CI and merging.

PR #1447, opened 2020-11-12
Update BNL-ATLAS_downtime.yaml

xzhao87:
facility-wide network intervention


matyasselmeci:
Downtime for BNL-SDCC resources needed to be in a separate yaml file. I created it in #1448 so you don't need to do anything.

PR #1452, opened 2020-11-16
remove check for 'Travis CI' app_name (SOFTWARE-4342)

edquist:
and update other logging messages.

As long as there is only one other check suite (the GHA-based one that replaces travis), i think we're fine without having to name it explicitly; though if there are other checks we may want to name just the one we care about for the webhook.


brianhlin (approved):
LGTM

PR #1461, opened 2020-11-17
Update BNL-Belle-II.yaml

xzhao87:


matyasselmeci:
Hi Xin,
Looks like our schema also requires values for `KSI2KMin` and `KSI2KMax`. Could you add those? You can use `0` if you haven't run the tests for those.
Thanks,
-Mat

xzhao87:
OK Mat, I added the two fields. Let me know if there is anything still missing. Thanks, Xin

matyasselmeci (approved):
Looks good.

PR #1464, opened 2020-11-17
Handling incorrect cert path or password in bin/osg-notify (SOFTWARE-3914)

Caffretro:
## 11/17/2020
- Added response to invalid certificate/key path
- Handling raised exception when incorrect password is received

### DIFF
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -125,6 +125,12 @@ def parseargs():
 
     args = oparser.parse_args()
 
+    if not os.path.exists(args.cert):
+        oparser.error("Certificate file path is invalid")
+    
+    if not os.path.exists(args.key):
+        oparser.error("Certificate key file path is invalid")
+
     if args.oim_recipients == 'vos' and args.owner_vo:
         oparser.error("--oim-owner-filter and --oim-recipients=vos options are conflicting")
     if args.name_filter and args.fqdn_filter:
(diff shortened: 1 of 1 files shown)


bin/osg-notify
```diff
@@ -185,7 +191,11 @@ def main():
         if args.fqdn_filter:
             results = topology_utils.get_resource_contacts_by_fqdn(args)
         else:
-            results = topology_utils.get_resource_contacts(args)
+            try:
+                # throws IOError for incorrect password
+                results = topology_utils.get_resource_contacts(args)
+            except IOError:
+                print("ERROR: incorrect password.")
```
brianhlin, line 198: Since we were seeing something like:

```
IOError: \[Errno 22] Invalid argument
```

You should check that the `errno` of the exception is 22, then `sys.exit` with your error message (we don't want the script to continue). If it's another errno, then you should raise the exception

matyasselmeci: Do you have a traceback for this exception? I don't see any mention of IOError in topology_utils; if something in topology_utils is raising (or propagating) an IOError, then that should be at least documented *in topology_utils*.

Although what I would do is catch the IOError in topology_utils, then make an "IncorrectPassword" exception and raise that instead.

brianhlin: Traceback here: https://opensciencegrid.atlassian.net/browse/SOFTWARE-3914?focusedCommentId=356097

brianhlin:
Closing PR that doesn't merge cleanly anymore

PR #1466, opened 2020-11-18
Package topology application into a docker container

jthiltges:
Initial draft at packaging topology into a container. Feedback would be helpful! The topology app looks OK from limited testing. I have not tested the webhook.

Changes to note:
- Add an `IGNORE_SECRET_PERMS` config flag to ignore ownership and permissions issues in containers
- Correct an issue where the mappings directory wasn't being passed into the loading function


Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN \
+    yum update -y && \
+    yum install -y \
+      gcc \
+      git \
+      libyaml-devel \
+      mod_ssl \
+      python3-devel \
+      python3-ldap \
+      python3-pip \
```
matyasselmeci, line 13:
```suggestion
      python3-pip && \
```

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    yum install -y \
```
matyasselmeci, line 14:
```suggestion
    && yum install -y \
```

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+      gridsite \
+      osg-ca-certs \
+      python3-mod_wsgi \
+    && yum clean all && rm -rf /var/cache/yum/*
+
+WORKDIR /app
+
+# Install application dependencies
+COPY requirements-rootless.txt src/ ./
+RUN pip3 install --no-cache-dir -r requirements-rootless.txt
```
matyasselmeci, line 26:
```suggestion
RUN pip3 install --no-cache-dir -r requirements-apache.txt
```

matyasselmeci: although it looks like you're already installing mod_wsgi and python3-ldap so I could be wrong about this. 
You should check the `/stashcache/authfile` endpoint to make sure the results contain a bunch of lines like
```
u 47146b14.0 /user/ligo rl
```

docker/config-webhook.example.py
```diff
@@ -0,0 +1,16 @@
...
+
+WEBHOOK_STATE_DIR   = "/data/webhook/state"
+WEBHOOK_DATA_DIR    = "/data/webhook/topology.git"
+WEBHOOK_DATA_REPO   = "https://github.com/opensciencegrid/topology"
+WEBHOOK_DATA_BRANCH = "master"
+WEBHOOK_SECRET_KEY  = "/etc/opt/topology/github_webhook_secret"
+WEBHOOK_GH_API_USER = "osg-bot"
+WEBHOOK_GH_API_TOKEN= "/etc/opt/topology/github_access_token"
+
+INSTANCE_NAME = "ITB webhook instance" # leave blank for production
```
edquist, line 14: I guess this means we're calling this ITB for now?

src/app.py
```diff
@@ -41,7 +41,10 @@ def _verify_config(cfg):
         else:
             st = os.stat(ssh_key)
             if st.st_uid != os.getuid() or (st.st_mode & 0o7777) not in (0o700, 0o600, 0o400):
-                raise PermissionError(ssh_key)
+                if cfg["IGNORE_SECRET_PERMS"]:
+                    app.logger.info("Ignoring permissions/ownership issues on " + ssh_key)
```
edquist, line 45: What's the deal with ignoring the ssh_key perms?  Won't ssh just error out later anyway if they are bad?

src/webapp/mappings.py
```diff
@@ -34,6 +34,6 @@ def get_nsfscience(indir: str, strict: bool) -> Dict:
     return nsfscience
 
 
-def get_mappings(self, indir="../mappings", strict=False):
+def get_mappings(indir="../mappings", strict=False):
```
edquist, line 37: Yikes. Good catch.

edquist (commented):
I didn't see how the secrets get passed in.  Maybe that'll happen at the k8s config?

Dockerfile
jthiltges: Good catch on the missing separator. I'll get that fixed!

Dockerfile
jthiltges: LDAP did seem to be working with the RPM version. I'll switch over to `requirements-apache.txt` and pip3 for installing the ldap and mod-wsgi modules. May as well keep the requirements together where possible, I guess!

src/app.py
jthiltges: I've been testing with docker and docker-compose so far (not yet running in k8s). When I bind mount the SSH key into the container, it inherits the uid of the host filesystem. If I set the permissions to 600, Apache can't read the file.

You could probably set the host filesystem owner to match the apache uid. This is a little messy. And it would probably still leave an issue with k8s. The docs for k8s secrets indicate you can set permissions, but not ownership.

edquist: Ah, i see.  So you are setting it to be world readable  Well whatever works.  I was just worried that if the above permissions check failed, then ssh would also complain.

For instance if i chmod a key file to 644:

```
$ ssh-add ./github_id_dsa
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0644 for './github_id_dsa' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
```

docker/config-webhook.example.py
jthiltges: I left the ITB instance name there as a placeholder. As you said, the secrets and config files would come from the k8s/flux config. (Or from the local filesystem if you want to use docker for development.)

src/app.py
jthiltges: In testing, it cloned the bitbucket repo and displayed contacts OK. Perhaps `ssh -i keyfile` isn't as strict?

edquist: Ok, sounds like it works then, which is the main thing!

matyasselmeci (approved):

matyasselmeci:
This looks good to me; do you have any more changes you'd like to get in, or can I merge (once the CI passes)?

jthiltges:
@matyasselmeci Up to you! We can wait until topology is up and running (and tested!) on Tiger, and handle the issues in this PR. Or just merge and open new PRs to handle them.

At the moment, I'm at a holding point, waiting for traefik config changes on Tiger to continue.

matyasselmeci:
I'll merge and we can use a new PR if you run into issues.

PR #1473, opened 2020-11-20
Add resource group UCL-INGRID with StashCache Origin resource

AndresTanasijczuk:
UCL-INGRID will host a StashCache Origin for data of the Virgo gravitational waves experiment. Currently this data are in the UNL StashCache Origin, which hosts data from both LIGO and Virgo.

The UNL Caching Infrastructure yaml file lists only LIGO as Allowed VO. I have put LIGO and Virgo. Should I only put LIGO?

I am not sure if SupportCenter should be "Self Supported" (as I put) or "Community Support Center".

### DIFF
--- a/topology/Universite catholique de Louvain/UCL-INGRID/UCL-INGRID.yaml
+++ b/topology/Universite catholique de Louvain/UCL-INGRID/UCL-INGRID.yaml
@@ -1,0 +1,27 @@
+Production: true
+SupportCenter: Self Supported
+GroupDescription: StashCache Origin in Louvain
+GroupID: 1097
+Resources:
+  UCL-Virgo-StashCache-Origin:
+    Active: true
+    Description: Virgo StashCache Origin
+    ID: 1108
+    ContactLists:
+      Administrative Contact:
+        Primary:
+          Name: Andres Tanasijczuk
+          ID: 7142b387956966f21dfcedf7fa4eef637c15837c 
+        Secondary:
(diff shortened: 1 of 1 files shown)


matyasselmeci:
"Self Supported" is fine.
AllowedVOs specifically refers to one of the VOs listed in the [virtual-organizations directory](https://github.com/opensciencegrid/topology/tree/master/virtual-organizations), where there is no "Virgo.yaml" file, only "LIGO.yaml". By putting LIGO into your AllowedVOs list, you are restricting your origin to only serve files in the directories listed here: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/LIGO.yaml#L82-L85.

Most origins just put "ANY" in their AllowedVOs list; since the VO also has to list your origin in its AllowedOrigins list, there isn't much of a use case for restricting your origin further.

AndresTanasijczuk:
Thanks Matyas.
I changed AllowedVOs to only contain "ANY".

matyasselmeci:
Resource names need to be globally unique so you should have UCL in the name somewhere too, for example `UCL-Virgo-StashCache-Origin`.
Otherwise, everything looks good.

AndresTanasijczuk:
Ok, I changed the resouce name to UCL-Virgo-StashCache-Origin as you suggest.

matyasselmeci (approved):
Looks good. Thank you!

PR #1477, opened 2020-11-23
Ignore user's environment in next_*_id scripts since environment variables may affect grep output

matyasselmeci:


edquist (commented):
> since environment variables may affect grep output

From our discussion today, it seems that this change set protects against users who ill-advisedly put

```
export GREP_OPTIONS=--color=always
```

in their `~/.bashrc`, with older versions of grep, that don't yet break altogether if this evn var is set.

Part of me wants just to declare that running these scripts with `GREP_OPTIONS=--color=always` is not a supported use case.  Because i'm moody, i guess.

But if we want to handle it gracefully, rather than helping the users fix their error - personally, i think i'd prefer to see
```
unset GREP_OPTIONS
```
instead of running things through `env -i`, which somehow feels heavy-handed to me.

Plus `env -i` removes PATH, which cripples a user who has egrep somewhere other than the standard location.  (That, you might say, is about as unlikely as a user who exports `GREP_OPTIONS=--color=always` -- but, i might say, it's not nearly as bad of an idea.)

Anyway i guess don't actually care that much either way, but them's mi thoughts.

matyasselmeci:
Well, the user doesn't really have any way of knowing that `GREP_OPTIONS=--color=always` is not something you should do...
I didn't realize `env -i` only searched for executables in standard locations -- the manpage is not very explicit on that (depends on how you interpret "start with"). I figured it found the executable in the current PATH but then started it with an empty environment. So I learned something about that, too.

Anyway, I'll be nice and just unset GREP_OPTIONS.

PR #1496, opened 2020-12-07
yaml spec requires a colon to be followed by a space for a mapping

matyasselmeci:
https://yaml.org/spec/1.2/spec.html#:%20mapping%20value//


djw8605 (approved):
LGTM

PR #1499, opened 2020-12-08
register missing LSU QB2 cluster

jdost321:
This resource does not have much life left on it since it is still rhel6 but in case it gets still used I'm registering it.


topology/Louisiana State University/LSU QB2/LSU_QB2.yaml
```diff
@@ -0,0 +1,29 @@
+Production: true
+SupportCenter: Self Supported
+GroupID: 500
```
brianhlin, line 3:
```suggestion
GroupID: 1098
```

brianhlin (approved):
The resource group ID needs to be updated but otherwise LGTM

PR #1515, opened 2020-12-14
T2_US_Wisconsin unscheduled outage of HDFS file access 2020-12-14

cvuosalo:
Unscheduled downtime for HDFS access to Wisconsin files until about 21:00 UTC. Other services are not affected. Jobs that don't require local HDFS files are not affected.


edquist:
(Just to note it here - this downtime PR was successfully merged; the second confused notice from osg-bot was due to a temporary misconfiguration on topology-itb)

PR #1518, opened 2020-12-15
remove duplicate cedar fqdn

edquist:
note the whole cedar resource group is disabled; ComputeCanada - Cedar
is the new one and it has the same fqdn

https://support.opensciencegrid.org/a/tickets/66458


brianhlin:
It's surprising that this actually passes CI. Could you instead remove the entire RG + downtime? This is an erroneous entry, anyway.

brianhlin (approved):
LGTM

PR #1530, opened 2021-01-07
Adding CC* sites view

djw8605:
Waiting on:

- [x] Approval from Janet


djw8605:
Can I get a courtesy review before merging.

src/templates/iframe.html.j2
```diff
@@ -239,14 +248,22 @@ function init(scrollwheel) {
             }
         }
 
+        var marker = new google.maps.Marker({
+            position: new google.maps.LatLng(sites[site][0], sites[site][1]),
+            map: map,
+            zIndex: resourceCounter,
+            icon: icon,
```
matyasselmeci, line 255: Is "icon" defined at this point?

matyasselmeci (approved):
One potential problem (see comment) but otherwise looks good so pre-approving.

src/templates/iframe.html.j2
djw8605: Good catch.  For some reason, it didn't complain.  Fixing.

PR #1534, opened 2021-01-08
Change my.opensciencegrid.org to topology.opensciencegrid.org

jthiltges:


bbockelm:
Yes!  No more "my.osg.org"!

brianhlin (approved):
LGTM. FWIW, we should verify that our data consumers (including GRACC @djw8605) are all using `topology.opensciencegrid.org` before removing any DNS entries

djw8605:
GRACC is using my.opensciencegrid.org... who maintains that...

matyasselmeci:
Some WLCG folks are likely to be using my.opensciencegrid.org: https://ggus.eu/?mode=ticket_info&ticket_id=148636

PR #1536, opened 2021-01-11
Update email for Andrew Melo

PerilousApricot:


matyasselmeci:
Hi Andrew,
Your ID doesn't change when your email address changes -- we only use the email has as a starting value for the ID.
We have a separate (private) contacts database where we put the actual email address. Let me know what the new email address is that you want to use, and I'll update the contacts DB.

PerilousApricot:
Ah, OK, I misunderstood. Let's drop this PR then -- I put the info in
freshdesk, ticket # 66566

PR #1544, opened 2021-01-14
Add the origin's own DN to its Authfile  (SOFTWARE-4403)

matyasselmeci:
Tested locally.

### DIFF
--- a/src/stashcache.py
+++ b/src/stashcache.py
@@ -393,7 +393,7 @@ def _origin_is_allowed(origin_hostname, vo_name, stashcache_data, resource_group
     return origin_resource.name in allowed_origins
 
 
-def _get_allowed_caches(vo_name, stashcache_data, resource_groups, suppress_errors=True):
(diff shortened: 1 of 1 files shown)


djw8605 (approved):
LGTM, thanks for this change.

src/stashcache.py
```diff
@@ -393,7 +393,7 @@ def _origin_is_allowed(origin_hostname, vo_name, stashcache_data, resource_group
     return origin_resource.name in allowed_origins
 
 
-def _get_allowed_caches(vo_name, stashcache_data, resource_groups, suppress_errors=True):
+def _get_allowed_caches(vo_name, stashcache_data, resource_groups, suppress_errors=True) -> List[Resource]:
```
edquist, line 396: Trying to pull a fast one there i see...

:)

edquist (approved):

matyasselmeci:
Deployed. @djw8605 , note that the UCL origin doesn't have its DN in topology yet so they will need to do that before this is useful to them.

PR #1580, opened 2021-01-29
register LSU - Deep Bayou Hosted CE

jdost321:
Creating with active set to false, until we're ready to push to production

### DIFF
--- a/topology/Louisiana State University/LSU - Deep Bayou/LSU-Deep_Bayou.yaml
+++ b/topology/Louisiana State University/LSU - Deep Bayou/LSU-Deep_Bayou.yaml
@@ -1,0 +1,32 @@
+Production: true
+SupportCenter: Self Supported
+GroupID: 1105
+GroupDescription: This is the LSU - Deep Bayou cluster
+Resources:
+  LSU-Deep_Bayou:
+    Active: false
+    ID : 1122
+    Description: The Hosted CE serving LSU - Deep Bayou
+    ContactLists:
+      Administrative Contact:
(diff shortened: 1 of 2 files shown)


topology/Louisiana State University/LSU - Deep Bayou/LSU-Deep_Bayou.yaml
```diff
@@ -0,0 +1,32 @@
+Production: true
+SupportCenter: Self Supported
+GroupID: 1105
+GroupDescription: This is the LSU - Deep Bayou cluster
+Resources:
+  LSU-Deep_Bayou:
```
brianhlin, line 6:
```suggestion
  LSU-DB-CE1:
```

Maybe something like this to more closely match the first bit of the FQDN?

brianhlin (approved):
One comment on the resource name but otherwise LGTM

topology/Louisiana State University/LSU - Deep Bayou/LSU-Deep_Bayou.yaml
jdost321: Hmm I think I like this suggestion. @jeffreynpeterson what are your thoughts? We can update the naming convention document if you agree.

jeffreynpeterson: matching the FQDN is good, we will still just have to make the call as to when to shorten or not the name for that depending on site.

PR #1597, opened 2021-02-04
add new service type for VO pilot containers (SOFTWARE-4370)

edquist:
'Execution Endpoint', that is.


brianhlin (approved):
LGTM. Could you submit another PR to update the services for all the resources that your container probe has created accounting records? NDSU and CUmulus backfill are two off the top of my head

PR #1598, opened 2021-02-05
Topology support for VO scitokens credentials

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-4453

### DIFF
--- a/virtual-organizations/GLOW.yaml
+++ b/virtual-organizations/GLOW.yaml
@@ -31,7 +31,7 @@ Contacts:
 Credentials:
   TokenIssuers:
     - URL: https://scitokens.org/chtc
-      UnixUser: glow
+      DefaultUnixUser: glow
 Disable: false
 FieldsOfScience:
   PrimaryFields:
--- a/virtual-organizations/GLOW.yaml
+++ b/virtual-organizations/GLOW.yaml
@@ -30,9 +30,8 @@ Contacts:
     Name: Miron Livny
 Credentials:
   TokenIssuers:
-    - https://scitokens.org/chtc
-  VomsServers:
-    - voms1.chtc.wisc.edu
+    - URL: https://scitokens.org/chtc
+      UnixUser: glow
 Disable: false
 FieldsOfScience:
   PrimaryFields:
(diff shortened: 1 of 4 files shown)


virtual-organizations/GLOW.yaml
```diff
@@ -28,6 +28,10 @@ Contacts:
   VO Manager:
   - ID: 48e1c2f26dc3479a6cf9b2de7c79d654ac27b1d1
     Name: Miron Livny
```
brianhlin, line 30: Separately, should this be @lmichael107 ?

virtual-organizations/GLOW.yaml
```diff
@@ -28,6 +28,10 @@ Contacts:
...
+Credentials:
+  TokenIssuers:
+    - URL: https://scitokens.org/chtc
```
brianhlin, line 33: Is this actually going to be the GLOW issuer?

brianhlin (commented):
Maybe we should go with `DefaultUnixUser` or `DefaultUser`? Some other minor questions

virtual-organizations/GLOW.yaml
matyasselmeci: 🤷🏻 It's what we used for StashCache below.

virtual-organizations/GLOW.yaml
matyasselmeci: I have no idea what these fields even mean...

brianhlin: We mainly use contact lists across VOs/sites to determine who to contact for various emails. Site contacts are increasingly being used to authorize token requests (i.e., CE and OS registries)

brianhlin:
> Maybe we should go with DefaultUnixUser or DefaultUser?

@matyasselmeci bikeshedding question from above

matyasselmeci:
> > Maybe we should go with DefaultUnixUser or DefaultUser?
> 
> @matyasselmeci bikeshedding question from above

I definitely want to keep Unix in there -- "User" can mean so many things. 50-50 on Default; do we have to remind people here that they can change it?

brianhlin:
> do we have to remind people here that they can change it?

I certainly don't think it hurts. It's similar to how we call the VO mapping file `voms-mapfile-default`

matyasselmeci:
> > do we have to remind people here that they can change it?
> 
> I certainly don't think it hurts. It's similar to how we call the VO mapping file `voms-mapfile-default`

OK, made the change. Anything else?

brianhlin (approved):
LGTM

PR #1602, opened 2021-02-08
Create SITE.yaml

biozit:
Adding AggieGrid cluster for State University Of Campinas


biozit:
Please, should I create the FACILITY.yaml?

matyasselmeci:
Yes, please create a FACILITY.yaml file, using the ID 10183.
Also for the directory name, please use the full name of the university instead of an abbreviation, so "University of Campanas" or "State University of Campanas".

biozit:
I think I create the right file.

matyasselmeci:
I merged your changes in #1605. You should be set now.

PR #1607, opened 2021-02-09
TokenIssuers fixups

matyasselmeci:
- Fixup the XML schema -- if there are going to be TokenIssuers, expect at least one TokenIssuer
- Add Credentials section to VO template yaml file


brianhlin (approved):
LGTM

PR #1611, opened 2021-02-11
Convenience script to run most of the next_*_id scripts and display their output in a table

matyasselmeci:
Often I need an ID both for a resource group and a resource; sometimes I need that
plus a new site and facility.  This is less annoying than running those scripts one
at a time.

I added Project and VO for completeness; I omitted Downtime (that's time based, not
sequence based) and Service Center (we're avoiding those).

### DIFF
--- a/bin/next_ids
+++ b/bin/next_ids
@@ -1,0 +1,9 @@
+#!/bin/bash
+progdir=$(dirname "$0")
+printf "%-15s: %6d\n" \
+"Project" "$("$progdir/next_project_id")" \
+"VO" "$("$progdir/next_virtual_organization_id")" \
+"Facility" "$("$progdir/next_facility_id")" \
+"Site" "$("$progdir/next_site_id")" \
+"Resource Group" "$("$progdir/next_resource_group_id")" \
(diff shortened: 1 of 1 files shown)


bin/next_ids
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+progdir=$(dirname "$0")
+printf "%-15s: %6d\n" \
+"Project" "$("$progdir/next_project_id")" \
```
edquist, line 4: Looks fine how you have it, but _if_ you wanna make things a bit less noisy with quotes you could do something like

```
cd "$(dirname "$0")"
```

and then

```
Project  "$(./next_project_id)" \
...
```
etc.

(And really even the quotes around the `"$(...)"` are a bit pedantic given that we control the scripts, which are all careful to output exactly one positive integer token to stdout.)

edquist (approved):

bin/next_ids
matyasselmeci: Sure

edquist:
Glorious

PR #1622, opened 2021-02-15
Beocat self-hosted Hosted CE is a proof-of-concept for GP-ARGO

brianhlin:
@kylehutson @Mansalu if I understood correctly, the new Beocat self-hosted Hosted-CE is part of GP-ARGO and therefore it is CC*. Is that correct?


kylehutson:
This is the testing prototype for GP-ARGO. We will be getting actual
GP-ARGO hardware in a few months.

brianhlin:
Ah ok, nevermind then!

PR #1631, opened 2021-02-17
Add downtime for SWT2_CPB_SE due to loss of cooling in the machine room

mdsosebee:


brianhlin (approved):
Verified offline. Merging.

PR #1634, opened 2021-02-17
Updating FNAL_Wilson resource with new CE fqdn

Farrukh-Aftab:


brianhlin (approved):
Hi @Farrukh-Aftab somehow we all managed to miss this, apologies. LGTM

PR #1648, opened 2021-03-01
correcting project name to match OSG Connect infrastructure

lmichael107:


brianhlin:
Hey @lmichael107, I see that `TNTech_ITS.yaml` still exists. Should we be removing the old project, too?

lmichael107:
Yes. What's weird is that my local (laptop) repo does NOT have the TNTech_ITS.yaml project anymore, as I used 'git mv' to rename it. I've not deleted it from https://github.com/lmichael107/topology, which I hope fixes things.

brianhlin (approved):
LGTM

PR #1660, opened 2021-03-10
Fix service name for CU backfill containers

brianhlin:
Was marked as a CE since we didn't have a dedicated service type yet


edquist (commented):
LGTM

edquist (approved):

PR #1664, opened 2021-03-11
Update AGLT2.yaml

wuwj6269:
add sl-um-es3.slateci.io:30113 as the second squid server


topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -33,6 +33,41 @@ Resources:
         Description: Generic squid service
     VOOwnership:
       ATLAS: 90
+AGLT2-squid:
```
brianhlin, line 36:
```suggestion
AGLT2-squid-2:
```

Choosing a new resource name

topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -33,6 +33,41 @@ Resources:
...
+          Name: Wenjing Wu
+        Secondary:
+          ID: fe2ecaf4ad4a0923d1be605ad55a8ca71f67eb66
+          Name: Philippe Laurens
+        Tertiary:
+          ID: a833f94ad3840fe779702409c6b778d79aaed76f
+          Name: Shawn McKee
+    Description: This is the slate SQUID at AGLT2
+    FQDN: sl-um-es3.slateci.io
+    ID: 678
```
brianhlin, line 61:
```suggestion
    ID: 1129
```

Updating the resource ID

topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -33,6 +33,41 @@ Resources:
...
+    Services:
+      Squid:
+        Description: Generic squid service
+        Details:
+          uri_override: sl-um-es3.slateci.io:30113
+          Monitored: false
```
brianhlin, line 67:
```suggestion
          Monitored: true
```

Enable WLCG monitoring

topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -33,6 +33,41 @@ Resources:
...
+    Services:
```
brianhlin, line 68:
```suggestion
```

Remove extra `Services:` section

topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -33,6 +33,41 @@ Resources:
...
+          ID: a833f94ad3840fe779702409c6b778d79aaed76f
+          Name: Shawn McKee
+    Description: This is the slate SQUID at AGLT2
+    FQDN: sl-um-es3.slateci.io
+    ID: 678
+    Services:
+      Squid:
+        Description: Generic squid service
+        Details:
+          uri_override: sl-um-es3.slateci.io:30113
```
brianhlin, line 66: I don't think you can actually specify a port, does it work this way @DrDaveD ?

brianhlin (approved):
@wuwj6269 should be ok once you apply the suggestions. Waiting on a response from Dave about specifying the port

topology/University of Michigan/AGLT2/AGLT2.yaml
wuwj6269: Hi Brian. I just copied the example from mwt2, because this is using a different port than the default one.

brianhlin:
Hi @wuwj6269 it looks like the spacing is bad. Could you make sure the indentation matches your existing squid entry?

topology/University of Michigan/AGLT2/AGLT2.yaml
DrDaveD: No, the monitoring doesn't support that.  If we have to use another port, it can be done by an exception on the squid monitoring machine.  However I encourage you to convince kubernetes to use the standard port if possible instead.  I managed to do it in the [OSG kubernetes frontier-squid config](https://github.com/opensciencegrid/tiger-osg-config/tree/master/manifests/base/frontier-squid).

DrDaveD: This shows the 3 places port 3401 is configured: https://github.com/opensciencegrid/tiger-osg-config/search?q=3401

brianhlin: @wuwj6269 is 30113 the monitoring port or the port that jobs should be accessing your squid with?

LincolnBryant: @DrDaveD I just want to note that the way you're currently exposing the monitoring port is by using an entire IP address for the monitoring, via `LoadBalancer`:
```
  type: LoadBalancer
  selector:
    app: frontier-squid
  ports:
  - port: 3401
    protocol: UDP
    targetPort: 3401
    name: squid-monitor```

What we are doing on some nodes is to use `hostPort` to expose the monitoring port, e.g.:

https://github.com/slateci/slate-catalog/blob/master/stable/osg-frontier-squid/osg-frontier-squid/templates/deployment.yaml#L60-L65

This as I understand it somewhat short circuits the K8S networking. I _think_ you can run the default port on two different nodes, as long as the service is pinned to a particular node.

DrDaveD: @LincolnBryant That certainly looks simpler, I'll have to try that.  I assume it has to be paired with the [corresponding part](https://github.com/slateci/slate-catalog/blob/master/stable/osg-frontier-squid/osg-frontier-squid/templates/service.yaml#L30) in the Service, but unlike the way I did it it can be in the same Service.

@brianhlin That's a good point about the monitoring port vs the squid service port.  I was only thinking about monitoring because that's what the topology entry is used for.  The squid service port to use is configured in the applications, and the WLCG Web Proxy Auto Discovery gets that mainly from CMS and ATLAS configurations and from Shoal, and a little bit from exceptions on the squid monitor machine.

PR #1676, opened 2021-03-15
Add StashCache for Swinburne University (2)

edquist:
This is based on #1659, with missing fields added for SUT-STASHCACHE


brianhlin (approved):
This LGTM but I don't believe they'll be able to cache anything until the cache is added here: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/LIGO.yaml#L93. @zvada can you coordinate with your LIGO contacts to ensure that they want to allow Swinburne to cache LIGO data?

PR #1682, opened 2021-03-17
UCSD HPC Rename

brianhlin:
Per https://support.opensciencegrid.org/a/tickets/66930. We should hold off on merging this until we redeploy CEs and update the factory


brianhlin:
Superseded by #1783

PR #1696, opened 2021-03-22
Explain why ldap module might be missing in stashcache.py

matyasselmeci:


src/stashcache.py
```diff
@@ -2,7 +2,17 @@
...
-import ldap
+import sys
+try:
+    # TODO: Rewrite this module to use ldap3 which gets installed in rootless installs too.
+    import ldap
+except ModuleNotFoundError:
+    print("""\
+*** ldap module unavailable; this comes from "python-ldap" which is only in requirements-apache.txt
+    since it requires the openldap header files (openldap-devel on EL) to install.
+    If the openldap header files are available, pip install python-ldap to make this error go away.
```
edquist, line 13:
```
<------------------------------- HANDY RULER -------------------------------->
```

```suggestion
*** ldap module unavailable; this comes from "python-ldap" which is only in 
    requirements-apache.txt since it requires the openldap header files 
    (openldap-devel on EL) to install.  If the openldap header files are 
    available, pip install python-ldap to make this error go away.
```

edquist (approved):

PR #1701, opened 2021-03-23
Decommissioning the cluster

mbatgt:
Georgia Tech is in the process of vacating the old data center and this cluster will need to be decommissioned in favor of the new cluster that's being built and configured. The access to the old cluster was disabled on 3/23.

### DIFF
--- a/topology/Georgia Institute of Technology/Georgia Tech/Georgia Tech PACE.yaml
+++ b/topology/Georgia Institute of Technology/Georgia Tech/Georgia Tech PACE.yaml
@@ -1,9 +1,9 @@
 GroupDescription: Resources in the Georgia Tech PACE cluster, to be used by LIGO.
 GroupID: 423
-Production: true
+Production: false
 Resources:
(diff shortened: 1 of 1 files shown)


topology/Georgia Institute of Technology/Georgia Tech/Georgia Tech PACE.yaml
```diff
@@ -1,9 +1,9 @@
 GroupDescription: Resources in the Georgia Tech PACE cluster, to be used by LIGO.
 GroupID: 423
-Production: true
+Production: false
```
matyasselmeci, line 3:
```suggestion
Production: true
```
"Production" refers to the grid group; "true" means it's part of the production (default) grid, "false" means it's part of the ITB (testing) grid. So changing this is not necessary.

matyasselmeci (approved):

mbatgt:
Thanks for approving the changes and your clarification for the "production" field!

matyasselmeci:
Sure!

PR #1721, opened 2021-03-29
Make explicit that the 'Production' field is for choosing between the Production pool and the ITB pool

matyasselmeci:


brianhlin (approved):
LGTM!

PR #1728, opened 2021-04-02
Added new project: ACE_NIAID

rynge:


matyasselmeci:
I think our naming convention is that we do the organization first, then the PI, so this would be `NIAID_Hurt`, no?

matyasselmeci (approved):
Discussed over Slack; this is a special case.

PR #1732, opened 2021-04-06
Replicate belle.kek.jp

DrDaveD:
This adds replication of the belle.kek.jp cvmfs repository.  BNL has already been replicating it for a long time.

@zvada fyi


DrDaveD:
There is also a corresponding [config-repo PR](https://github.com/cvmfs-contrib/config-repo/pull/65) for Marian to add the client configuration.

I set UseOASIS to false; there is no /cvmfs/oasis.opensciencegrid.org/belle.

matyasselmeci (approved):

matyasselmeci:
This is fine from a topology point of view (that is, the value of UseOASIS does not affect the display of Managers or OASISRepoURLs).

PR #1741, opened 2021-04-08
ResourceAllocations (ITB)

matyasselmeci:


src/webapp/project_reader.py
```diff
@@ -46,34 +47,50 @@ def get_resource_allocation(data):
...
+            data['Sponsor']['VirtualOrganization'] = OrderedDict([("ID", ID), ("Name", name)])
+
+        if 'ResourceAllocation' in data:
+            data['ResourceAllocation'] = get_resource_allocation(data)
+    except Exception as e:
+        log.error("%r adding project %s", e, file)
+        log.error("Data:\n%s", pprint.pformat(data))
+        raise
+    project.update(data)
+    return project
```
edquist, line 72: You get a star for trying to out some of this stuff out into a function.

It's kinda too bad that there is still the try/except scaffolding around the core of the function, and then again in its callsite in `get_projects`.  I am tempted to say the try/except error logging here could just be left in the caller, but maybe then you couldn't log the contents of the file so easily, which i suppose is your point.  In which case you have my sympathy.

edquist (approved):
Go for it ... Connect Four!

src/webapp/project_reader.py
matyasselmeci: Thanks :-)

PR #1743, opened 2021-04-08
Convert scripts to Python 3

matyasselmeci:


bin/osg-notify
```diff
@@ -203,8 +197,8 @@ def main():
                     emails.add(contact['Email'])
         recipients.update(emails)
 
-    with open(args.message, 'r') as fp:
-        contents = fp.read()
+    with open(args.message, 'rb') as fp:
+        contents = fp.read().decode('utf-8', errors='replace')
```
edquist, line 201: The point of opening binary and then decoding is to handle utf-8 encoding errors, i take it?

bin/osg-topology
```diff
@@ -178,14 +172,16 @@ def main():
...
         sys.exit("Failed to read email file: %s" % exc)
+    except UnicodeDecodeError as exc:
+        sys.exit("Invalid unicode in email file: %s" % exc)
 
-    commands = {"list_resource_contacts": list_resource_contacts,
-                "list_vo_contacts": list_vo_contacts,
-                "list_contacts": list_contacts,
+    commands = {"list-resource-contacts": list_resource_contacts,
+                "list-vo-contacts": list_vo_contacts,
+                "list-contacts": list_contacts,
```
edquist, line 184: does this change (underscores to dashes) break anything?  Or is it just the user CLI?

bin/osg-topology
```diff
@@ -168,7 +160,9 @@ def main():
     oparser = get_parser()
 
     args = oparser.parse_args()
-    if getattr(args, "filter") is not None:
+    if not args.which:
+        oparser.error(f"No command specified; see {sys.argv[0]} --help for usage")
+    if getattr(args, "filter", None) is not None:
```
edquist, line 165: Is the `None` default necessary?  I am wondering if `oparser.parse_args()` won't automatically set the `filter` attribute to `None` by default.

Not that it hurts anything to provide the default, though.

edquist (approved):
A few small questions, otherwise it looks good.  Nothing too crazy.

bin/osg-notify
matyasselmeci: Exactly.

bin/osg-topology
matyasselmeci: It will not. I think I can add `default=None` to the `add_argument()` calls above to get the same behavior though.

bin/osg-topology
matyasselmeci: It's necessary because of the
```
subparsers = oparser.add_subparsers(dest="which")
```
line. This causes argparse to put the name of the subparser into `args.which`. The previous code manually added a default value for `args.which` using the underscore version. It's not user-visible.

PR #1747, opened 2021-04-13
Support for Resource Allocations in projects

matyasselmeci:
SOFTWARE-4563

Same as #1741 but for master. Do not merge until I've updated the production topology image to 1.5.0.


matyasselmeci:
1.5.0 has already been deployed on topology-itb and works just fine.

matyasselmeci:
Deployed in production.

PR #1769, opened 2021-04-21
Add --regex argument to get-scitokens-mapfile to print output in regex format; also use SCITOKENS instead of SCITOKEN

matyasselmeci:
Without:
```
# OSG
SCITOKENS "https://scitokens.org/osg-connect" osg
# GLOW
SCITOKENS "https://scitokens.org/chtc" glow
```

With:
```
# OSG
SCITOKENS /https\:\/\/scitokens\.org\/osg\-connect/ osg
# GLOW
SCITOKENS /https\:\/\/scitokens\.org\/chtc/ glow
```

### DIFF
--- a/bin/get-scitokens-mapfile
+++ b/bin/get-scitokens-mapfile
@@ -50,7 +50,7 @@ def main(argv):
             url = token_issuer.get("URL")
             if url:
                 if args.regex:
-                    url = f'/{re.escape(url)}/'
+                    url = f'/^{re.escape(url)},/'
                 else:
                     url = f'"{url}"'
             unix_user = token_issuer.get("DefaultUnixUser")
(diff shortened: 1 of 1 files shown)


bin/get-scitokens-mapfile
```diff
@@ -43,12 +49,15 @@ def main(argv):
         for token_issuer in vo_data["Credentials"]["TokenIssuers"]:
             url = token_issuer.get("URL")
             if url:
-                url = f'"{url}"'
+                if args.regex:
+                    url = f'/{re.escape(url)}/'
```
brianhlin, line 53:
```suggestion
                    url = f'/^{re.escape(url)},/'
```

@edquist and I verified that the regex acts more like `re.search` instead of `re.find` so we should:

1. Lock the regex to the front of the string
2. Ensure that there isn't any funny business at the end of the URL: the authenticated identity of a SciToken will have the form `<TOKEN ISSUER>,<TOKEN SUBJECT>` so we can use `,` as the URL terminator

brianhlin (approved):
One change but otherwise LGTM

PR #1789, opened 2021-05-03
Update Lehigh - Hawk.yaml

ultramathman:
Changing status to Active/Production after successful runs from OSG Factory Operations.


brianhlin (approved):
LGTM. Thanks for the PR!

PR #1795, opened 2021-05-06
new Resource Allocation schema (ITB)

matyasselmeci:
- Update TG-CHE200122's allocations to match the new schema
- Update ResourceAllocations schema
- Update verify_projects.py for new schema
- Update project_reader.py with the new schema
- enumerate() returns the index first
- Combine TG-CHE200122.yaml's allocations
- Make Type first


projects/TG-CHE200122.yaml
```diff
@@ -26,13 +26,13 @@ Sponsor:
   CampusGrid:
     Name: OSG-XSEDE
 
-ResourceAllocation:
-  XRAC:
+ResourceAllocations:
+  - Type: XRAC
     AllowedSchedds:
       - CHTC-XD-SUBMIT
     ResourceGroups:
```
brianhlin, line 33: Maybe I missed the bikeshedding window on this but I find `ResourceGroups` confusing here alongside `AllowedSchedds`. I _think_ that this is differentiating between allowed access points and allowed execution points and I don't think `ResourceGroups` makes sense for that: presumably `CHTC-XD-SUBMIT` also must belong to its own resource group, no?

src/tests/verify_projects.py
```diff
@@ -55,36 +55,36 @@ def validate_project_file(self, project_filename: str) -> List[str]:
...
-                        )
+        if not is_null(project, "ResourceAllocations", "ResourceAllocation"):
+            for resource_allocation in project["ResourceAllocations"]["ResourceAllocation"]:
+
+                # Check 1
+                if not is_null(resource_allocation, "ResourceGroups", "ResourceGroup"):
+                    project_rgs = resource_allocation["ResourceGroups"]["ResourceGroup"]
+                    project_rg_names = [x["GroupName"] for x in project_rgs]
+
+                    for rg in project_rg_names:
```
brianhlin, line 66: Couldn't you speed this up with a set comparison?

src/webapp/project_reader.py
```diff
@@ -62,8 +60,9 @@ def get_one_project(file: str, campus_grid_ids: Dict, vos_data: VOsData) -> Dict
             ID = vos_data.vos[name]['ID']
             data['Sponsor']['VirtualOrganization'] = OrderedDict([("ID", ID), ("Name", name)])
 
-        if 'ResourceAllocation' in data:
-            data['ResourceAllocation'] = get_resource_allocation(data)
+        if 'ResourceAllocations' in data:
+            resource_allocations = [get_resource_allocation(ra, idx) for idx, ra in enumerate(data['ResourceAllocations'])]
+            data['ResourceAllocations'] = {"ResourceAllocation": resource_allocations}
```
brianhlin, line 65: As I mentioned on Slack, this data structure confuses me (same thing with the resource groups above). Is the extra dict layer (`ResourceAllocation`/`ResourceGroup`) for the XML generation?

brianhlin (commented):
Am I too late to get in on that good bikeshedding action? Code itself looks generally fine, though

src/webapp/project_reader.py
matyasselmeci: Yep it's for the XML. The library that generates XML takes a dict element with a list value and turns it into a series of tags where the name of the tag is the dict key, so the Python data structure that looks like
```
{
 "ul": {
   "li": ["1", "2", "3"]
 }
}
```
turns into
```
<ul>
 <li>1</li>
 <li>2</li>
 <li>3</li>
</ul>
```

projects/TG-CHE200122.yaml
matyasselmeci: I think we still have time for a bit of bikeshedding. `AllowedSchedds` are topology resources that have a "Submit Node" service; `ResourceGroups` are topology resource groups that contain the hosted CEs in front of the HPC clusters. So yeah the names are confusing.

How about `SubmitResources` and `ExecuteResourceGroups` instead? We can bikeshed over Slack if you don't like those.

src/tests/verify_projects.py
matyasselmeci: Sure can.

projects/TG-CHE200122.yaml
brianhlin: Yeah, I like that better

matyasselmeci: Cool. @bbockelm, any objections?

PR #1799, opened 2021-05-10
Removing lucille from WLCG Accounting monitoring

djw8605:
In FD#67238, Horst reports that the CE lutgw1.lunet.edu is down, and may come back up this summer as a new, or resurrected CE.  I am turning off the WLCG accounting monitoring, as the CE hasn't reported to GRACC in several months, at least.


brianhlin (approved):
LGTM

PR #1811, opened 2021-05-12
Accept zero-length downtime

matyasselmeci:
We don't know if our downstream consumers accept downtime deletions so have the tests accept zero-length downtimes as a way of effectively deleting it.


brianhlin (approved):
LGTM

PR #1823, opened 2021-05-17
update SDSC-PRP-CE1 resource name

jdost321:
Part of the ongoing naming cleanup activities here
https://opensciencegrid.freshdesk.com/a/tickets/66930


brianhlin:
@jdost321 do you want to do the resource + site renames in one fell swoop?

jdost321:
interesting I didn't intend those to get into this PR i guess because i added them to the same branch before it was committed? Let's keep them separate just in case we have to revert the site rename?

brianhlin (approved):
Jeff and I discussed offline that although unintended, this is actually exactly what we planned on doing anyway

PR #1828, opened 2021-05-19
swap Doug Benjamin in for John Hover for BNL USATLAS

edquist:
https://support.opensciencegrid.org/a/tickets/67235

>>> I noticed references to John Hover. He has left BNL a while ago and those reference should be removed.
>>
>> I see John Hover listed for various support center contacts for SDCC and USATLAS.  Do you want to replace him for all of these?
>
> For the US ATLAS at BNL yes. I will check with colleagues to see if I should be the replacement 
> For the rest.


brianhlin (approved):
Sure, though I don't think we really user support center these days

PR #1834, opened 2021-05-20
add new osg-bot status message for CI action_required (SOFTWARE-4627)

edquist:
in these cases apparently the first-time contributor must be authorized to run the GHA check suite workflow


src/webapp/webhook_status_messages.py
```diff
@@ -114,6 +114,19 @@
     "-- OSG-BOT  :disappointed:"
 )
 
+ci_action_required = (
+    "Greetings @{sender},\n"
+    "\n"
+    "Thank you for your pull request!\n"
+    "\n"
+    "Your downtime update looks clean; but OSG Staff must authorize"
```
brianhlin, line 122: I don't think we can claim cleanliness of the downtime update if the GHA won't run?

edquist: It's actually the same language used [just above](https://github.com/opensciencegrid/topology/blob/0b477ad46dd87dcd4b6704a4e40ac585fe56ea30/src/webapp/webhook_status_messages.py#L106-L115) in this file for `ci_falure` ("Your downtime update looks clean, but our CI checks failed;").

The "downtime update looks clean" in both cases refers to the webhook app's [automerge checks](https://github.com/opensciencegrid/topology/blob/master/src/webapp/automerge_check.py#L142-L148), which detect that the PR is a downtime update that satisfies all the automerge eligibility conditions, apart from the CI checks.

So, osg-bot is saying, "this looks like a downtime update, and it passes all my own checks for automerge eligibility, except that I also require a successful CI run."

But feel free to suggest alternate language that you think would be more clear.

brianhlin: I'd strike it entirely from both messages:

> Thanks for the PR but OSG Staff must authorize...

and

> Thanks for the PR but our CI checks failed...

Separately, for this case, is it possible to extract messages from the CI?

edquist: Ok; so, one point of the messages (at least as much for us staff as for the PR requester) was to acknowledge that the PR is a downtime PR and to note that, apart from the CI checks, all the downtime automerge eligibility requirements have been met.  I wonder if we can find a succinct way to express that also that is not confusing.

>Separately, for this case, is it possible to extract messages from the CI?

In the new case, i think the point is that the CI was not triggered, right?

And in the case of the ci failure, i suppose it would be possible to grab all the logs and dump them, though finding the error seems like a Hard Problem (tm), and clicking the red X usually leads to the answer.

brianhlin: > apart from the CI checks, all the downtime automerge eligibility requirements have been met

If they haven't been met, `osg-bot` tells the user exactly what the problem is, no? So if we get one of these CI failure messages (either CI didn't run or CI actually fails), don't we know that the pre-reqs have been satisfied?

> In the new case, i think the point is that the CI was not triggered, right?
> 
> And in the case of the ci failure, i suppose it would be possible to grab all the logs and dump them, though finding the error seems like a Hard Problem (tm), and clicking the red X usually leads to the answer.

Yeah, I'm talking about the CI failure case. I want to save the user the trouble of going to the actual CI, especially because they can be very noisy. Finding the error shouldn't be very hard if our tests provide sensible summaries. Something like:

> The XML validation failed, click the "actions" tab for details

edquist: > If they haven't been met, `osg-bot` tells the user exactly what the problem is, no? So if we get one of these CI failure messages (either CI didn't run or CI actually fails), don't we know that the pre-reqs have been satisfied?

Yeah that's a fair point.  If there are any downtime files affected, osg-bot will tell us specifically about any impediments to automerge eligibility.  So, like you are saying, we would know that there are none upon hearing just about the CI failure / failure to run ... if we remember that that's how the logic works (without being reminded about it in the message).

> Yeah, I'm talking about the CI failure case. I want to save the user the trouble of going to the actual CI, especially because they can be very noisy. Finding the error shouldn't be very hard if our tests provide sensible summaries.

I see ... so i took a quick glance at the docs and found 

https://docs.github.com/en/rest/reference/checks#get-a-check-run

which would point us to, in this case,

https://api.github.com/repos/opensciencegrid/topology/check-runs/2623296167
https://api.github.com/repos/opensciencegrid/topology/check-runs/2623296167/annotations

which doesn't seem to give the details we'd care about for tracking things down.  Maybe there is an api for getting more details about the check run, but i have not found it.

Anyway we can make a ticket for this if you want, but i have no expectations for what we will have.

> > The XML validation failed, click the "actions" tab for details

o_O

Seems _way_ easier to just click the red X by the failed commit short-hash.  When you click [Details](https://github.com/opensciencegrid/topology/runs/2623296167) It opens up the section with the failure and jumps to the line with the Error.  

To me, clicking on the [Actions](https://github.com/opensciencegrid/topology/actions) tab seems a lot less straightforward...

brianhlin: Yeah, I meant the "Checks" tab in the PR, which I believe should do the same thing as clicking the red X in the failure mode. 

As for getting "summary" email from the CI, looking at the API ref, it has this in the example:

```
  "output": {
    "title": "Mighty Readme report",
    "summary": "There are 0 failures, 2 warnings, and 1 notice.",
    "text": "You may have some misspelled words on lines 2 and 4. You also may want to add a section in your README about how to install your app.",
    "annotations_count": 2,
    "annotations_url": "https://api.github.com/repos/github/hello-world/check-runs/4/annotations"
```

Which leads me to believe that we just need to figure out how to annotate our CI output so that we can pick it up.

brianhlin:
For this PR, I think I'd just like to see rewording of the message.  "Your...update looks clean but..." is contradictory contradictory: what is "clean" and why can't you just merge a "clean" PR?

edquist:
> For this PR, I think I'd just like to see rewording of the message.

You got it - done!

>"Your...update looks clean but..." is contradictory contradictory:

:disappointed: 

>what is "clean"

I tried to explain that above:

> The "downtime update looks clean" in both cases refers to the webhook app's [automerge checks](https://github.com/opensciencegrid/topology/blob/master/src/webapp/automerge_check.py#L142-L148), which detect that the PR is a downtime update that satisfies all the automerge eligibility conditions, apart from the CI checks.
> 
> So, osg-bot is saying, "this looks like a downtime update, and it passes all my own checks for automerge eligibility, except that I also require a successful CI run."

The point is that it is "clean" with respect to automerge eligibility requirements; mainly, the PR:

- Has only modified downtime files

- Was submitted by a github user registered with a contact (in our contact repo),
  which is associated with the all the resources corresponding to the changes made in the downtime file(s)

...

> and why can't you just merge a "clean" PR?

Another case of a "clean but won't merge" is if the PR head is not up-to-date with master.

src/webapp/webhook_status_messages.py
edquist: I see; yeah there is probably a way to call the api to add output text for the api run.  I am not sure how to do that; currently the most specific thing i see in the output is "Validate Topology data", which is pretty generic

```
  "output": {
    "title": "Validate Topology data",
    "summary": "There are 1 failures, 0 warnings, and 0 notices.",
    "text": null,
    "annotations_count": 1,
    "annotations_url": "https://api.github.com/repos/opensciencegrid/topology/check-runs/2623296167/annotations"
  },
```

brianhlin (approved):
LGTM. My "what is clean?" question was from the POV of a user who doesn't have the background knowledge that you (or now I) have! 

EDIT: Basically, we should tell people which requirement/test failed for the PR explicitly. "Clean" is a value judgement that only makes sense in the passed-all-requirements case.

src/webapp/webhook_status_messages.py
brianhlin: Something like this maybe? https://github.com/marketplace/actions/github-checks

brianhlin:
@edquist will the webhook pick up these changes automatically once merged?

matyasselmeci:
I believe you'll need to tag a new release, which causes dockerhub to make a new image, and use that new image in the topology pods on tiger. (I can point you to the right place later)

edquist:
Ah, i had forgotten that tagging was a necessary step in the process.

PR #1860, opened 2021-06-02
Upgrade urllib3

matyasselmeci:


brianhlin (approved):
Pre-approving but I'd like to see this tested first

matyasselmeci:
This is making pip install hang. I don't understand why it doesn't affect building the docker image...

matyasselmeci:
Tested on topology-itb via the v1.5.3.1 tag. I'll upgrade production topology tomorrow.

PR #1862, opened 2021-06-03
Added new project: Fairfield_Kubasik

rynge:


matyasselmeci:
Looks like there's a "Fairfield" organization you can use -- it's used by https://github.com/opensciencegrid/topology/blob/master/projects/cms.org.fairfield.yaml

rynge:
I don't understand the request. Are you saying we do not want "University" in the name?

matyasselmeci:
We want to avoid having duplicates of the same organization with different spellings and I don't think we're able to rename existing projects, at least not if they have records in GRACC.

rynge:
Ok, will remove it then. Hang on

matyasselmeci (approved):

PR #1869, opened 2021-06-07
Fix psc bridges2

jdost321:
For this one I'm hoping just a rename of site, resource group, and resource is enough, keeping existing registrations / IDs.


brianhlin (approved):
LLGTM

PR #1886, opened 2021-06-11
The "Submit Node" service is also used for... submit nodes

matyasselmeci:


brianhlin (approved):
That's kinda miserable but a comment's is better than nothing

PR #1887, opened 2021-06-11
Register OSG Connect submit hosts login04 and login05

matyasselmeci:
@LincolnBryant , @jlstephen , do these entries look good? They didn't really fit into any of the existing ResourceGroups so I created a new one.


jlstephen:
Looks fine to me. @LincolnBryant do we want to add David Jordan as the tertiary contact?

brianhlin (approved):
LGTM and we can always add David Jordan later if that's what you want @LincolnBryant @jlstephen

PR #1910, opened 2021-06-22
Move Topology container upload to GHA (SOFTWARE-4652)

yongleyuan:
Modify `build-sw-container.yml` file:
- Add git tag regex that triggers GHA workflow
- Remove `strategy:` block
- Remove `TIMESTAMP` env var and its required step
- Change tag list to "latest,{tag_without_leading_v}"

Add it to `.github/workflows` dir.

### DIFF
--- a/.github/workflows/build-sw-container.yml
+++ b/.github/workflows/build-sw-container.yml
@@ -7,15 +7,6 @@ on:
       - /^v([0-9.]+)$/
 
 jobs:
-  make-date-tag:
-    runs-on: ubuntu-latest
-    if: startsWith(github.repository, 'opensciencegrid/')
-    outputs:
-      dtag: ${{ steps.mkdatetag.outputs.dtag }}
-    steps:
-    - name: make date tag
-      id: mkdatetag
-      run: echo "::set-output name=dtag::$(date +%Y%m%d-%H%M)"
 
   build:
     runs-on: ubuntu-latest
@@ -25,8 +16,6 @@ jobs:
 
     - name: Generate tag list
       id: generate-tag-list
-      env:
-        REPO: ${{ matrix.repo }}
       run: |
         docker_repo=${GITHUB_REPOSITORY/opensciencegrid\/docker-/opensciencegrid/}
         tag_list=latest,${{ GITHUB_REF##v }}
--- a/.github/workflows/build-sw-container.yml
+++ b/.github/workflows/build-sw-container.yml
@@ -2,7 +2,6 @@ name: Build and Push Docker image
 
 on:
   push:
-    branches: [ master ]
     tags:
       - /^v([0-9.]+)$/
 
--- a/.github/workflows/build-sw-container.yml
+++ b/.github/workflows/build-sw-container.yml
@@ -33,5 +33,4 @@ jobs:
       uses: docker/build-push-action@v2.2.0
       with:
         push: true
-        build-args: BASE_YUM_REPO=${{ matrix.repo }}
         tags: "${{ steps.generate-tag-list.outputs.taglist }}"
--- a/.github/workflows/build-sw-container.yml
+++ b/.github/workflows/build-sw-container.yml
@@ -3,7 +3,8 @@ name: Build and Push Docker image
 on:
   push:
     tags:
-      - /^v([0-9.]+)$/
+      - v[0-9]+.[0-9]+.[0-9]+
+      - v[0-9]+.[0-9]+.[0-9]+.[0-9]+
 
 jobs:
 
@@ -17,7 +18,7 @@ jobs:
       id: generate-tag-list
       run: |
         docker_repo=${GITHUB_REPOSITORY/opensciencegrid\/docker-/opensciencegrid/}
-        tag_list=$docker_repo:latest,$docker_repo:${GITHUB_REF#v }
+        tag_list=$docker_repo:latest,$docker_repo:${GITHUB_REF#v}
         echo "::set-output name=taglist::$tag_list"
 
     - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/build-sw-container.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  make-date-tag:
+    runs-on: ubuntu-latest
+    if: startsWith(github.repository, 'opensciencegrid/')
+    outputs:
+      dtag: ${{ steps.mkdatetag.outputs.dtag }}
+    steps:
+    - name: make date tag
+      id: mkdatetag
+      run: echo "::set-output name=dtag::$(date +%Y%m%d-%H%M)"
```
brianhlin, line 18:
```suggestion
```

We don't use the timestamp anywhere anymore so we can drop this job

.github/workflows/build-sw-container.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Generate tag list
+      id: generate-tag-list
+      env:
+        REPO: ${{ matrix.repo }}
+      run: |
+        docker_repo=${GITHUB_REPOSITORY/opensciencegrid\/docker-/opensciencegrid/}
+        tag_list=latest,${{ GITHUB_REF##v }}
```
brianhlin, line 32:
```suggestion
        tag_list=$docker_repo:latest,$docker_repo:${GITHUB_REF#v }
```

1. I forgot to mention that each tag has the form `CONTAINER REPO:CONTAINER TAG`
2. In bash, you can reference vars with `$VAR` or `${VAR}`
3. We can slightly shorten the parameter expansion since we just need to remove the single character. From `man bash`:

```
       ${parameter#word}
       ${parameter##word}
              Remove  matching  prefix  pattern.   The word is expanded to produce a pattern just as in pathname expansion, and matched against the expanded value of parameter using the
              rules described under Pattern Matching below.  If the pattern matches the beginning of the value of parameter, then the result of the expansion is the  expanded  value  of
              parameter with the shortest matching pattern (the ``#'' case) or the longest matching pattern (the ``##'' case) deleted.  If parameter is @ or *, the pattern removal oper‐
              ation is applied to each positional parameter in turn, and the expansion is the resultant list.  If parameter is an array variable subscripted with @ or *, the pattern re‐
              moval operation is applied to each member of the array in turn, and the expansion is the resultant list.
```

.github/workflows/build-sw-container.yml
```diff
@@ -0,0 +1,49 @@
...
+  build:
+    runs-on: ubuntu-latest
+    if: startsWith(github.repository, 'opensciencegrid/')
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Generate tag list
+      id: generate-tag-list
+      env:
+        REPO: ${{ matrix.repo }}
```
brianhlin, line 29:
```suggestion
```

We got rid of the matrix strategy block so we need to remove this, too

.github/workflows/build-sw-container.yml
```diff
@@ -0,0 +1,49 @@
+name: Build and Push Docker image
+
+on:
+  push:
+    branches: [ master ]
```
brianhlin, line 5:
```suggestion
```

Looking at the documentation, it looks like branch and tag restriction lists are or'ed, not and'ed together https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

.github/workflows/build-sw-container.yml
```diff
@@ -0,0 +1,49 @@
...
+    tags:
+      - /^v([0-9.]+)$/
```
brianhlin, line 7: It looks like regular expressions aren't supported so you'll need to modify this using https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet to make sure that `v1.2.3`-style tags trigger the run. Note that any part of the version number can have multiple digits

brianhlin (changes requested):
@yongleyuan could you reword your commit to match the ticket description of https://opensciencegrid.atlassian.net/browse/SOFTWARE-4652 (keeping the ticket ID in parens at the end)? You'll want to use `git rebase -i` and `git push -f` to do that. Then go ahead and address the other comments

.github/workflows/build-sw-container.yml
yongleyuan: Must the tag have two dots? For example, are v1.2 and v2 valid?

yongleyuan: > It looks like regular expressions aren't supported so you'll need to modify this using https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet to make sure that `v1.2.3`-style tags trigger the run. Note that any part of the version number can have multiple digits

Must the tag have two dots? For example, are v1.2 and v2 valid?

brianhlin: We use something approaching [semantic versioning](https://semver.org/) so there will always be 3 numbers each separated by 1 dot so `v1.2` and `v2` are invalid in this case.

yongleyuan: Will `v[0-9]+.[0-9]+.[0-9]+` work?

brianhlin: Looks right to me!

matyasselmeci: No, we've occasionally used 4 numbers. You could try `v[0-9.]+` or `v[0-9]+.[0-9]+.[0-9]+*` -- both are a little too permissive but I don't see a way to get it exactly right.

brianhlin: Since `tags` is a list we could do something like:

```
tags:
  - v[0-9]+.[0-9]+.[0-9]+
  - v[0-9]+.[0-9]+.[0-9]+.[0-9]+
```

matyasselmeci: Sure, that works.

brianhlin (approved):
LGTM!

PR #1912, opened 2021-06-24
Register the CHTC LHCb CE

brianhlin:
We want to start accepting OSG and GLOW pilots


matyasselmeci (approved):
Does LHCb still use it?

brianhlin:
Yup, they do

PR #1924, opened 2021-07-02
Add workflow_dispatch

mtru32:
Adds the ability to trigger new workflow runs from the UI.


brianhlin:
Hey @mtru32 this is my mistake but I don't actually want this change for this repo in particular. Generally, though, this change looks good so you can continue with the other repos from the ticket. Also, don't forget to include `(SOFTWARE-XXXX)` in the PR title!

PR #1928, opened 2021-07-07
Update brown-cms.yaml

gavila1-brown:
Change FQDN for resource back to brux11.hep.brown.edu


matyasselmeci (approved):
No CE records in GRACC in the last two years, so this is fine.

PR #1930, opened 2021-07-07
Update TACC-Stampede2.yaml

tanyalevshina:
added WLCGInformation block (ApelNormalFactor) for TACC-Stampede2

### DIFF
--- a/topology/University of Texas at Austin/Texas Advanced Computing Center/TACC-Stampede2.yaml
+++ b/topology/University of Texas at Austin/Texas Advanced Computing Center/TACC-Stampede2.yaml
@@ -37,6 +37,10 @@ Resources:
       ATLAS: 50
     WLCGInformation:
       APELNormalFactor: 18
+      InteropAccounting: false
+      InteropMonitoring: false
+      KSI2KMax: 0
+      KSI2KMin: 0
   TACC-Stampede2:
     ID: 1082
     Active: false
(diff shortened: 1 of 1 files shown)


djw8605:
Brian, could you confirm that the WLCG is at the right indention level. It doesn’t look right at first glance. 

-Derek

topology/University of Texas at Austin/Texas Advanced Computing Center/TACC-Stampede2.yaml
```diff
@@ -35,6 +35,8 @@ Resources:
     VOOwnership:
       CMS: 50
       ATLAS: 50
+  WLCGInformation:
+      APELNormalFactor: 18
```
matyasselmeci, line 39:
```suggestion
    WLCGInformation:
        APELNormalFactor: 18
```

matyasselmeci:
Good catch!

matyasselmeci:
Looks like the schema won't validate unless you specify a few more WLCGInformation attributes:
- InteropMonitoring
- InteropAccounting
- KSI2KMin
- KSI2KMax

(the first two are bools and the last two are floats; the other WLCGInformation attributes are strings so when topology fills them in with an empty string, that fits the schema. This should probably be fixed in topology.)

matyasselmeci (approved):

PR #1940, opened 2021-07-13
Break out cilogon ldap stuff into separate module (SOFTWARE-4528)

edquist:
inspired by SOFTWARE-4528, which uses cilogon info for purposes
unrelated to oasis managers

Also, move safe_dict_get into common.  (Currently it's only used in
oasis_managers, but it is a generic helper function with perhaps
broader applicability.)

### DIFF
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -1,0 +1,44 @@
+import ldap3
+
+
+def get_contact_cilogon_id_map(global_data):
+    """ return contacts dict, limited to users with a CILogonID """
+    contacts = global_data.get_contacts_data().users_by_id;
+    return { k: v for k, v in contacts.items() if v.cilogon_id is not None }
+
+
+# cilogon ldap query constants
+_ldap_url = "ldaps://ldap.cilogon.org"
+_username = "uid=readonly_user,ou=system,o=OSG,o=CO,dc=cilogon,dc=org"
+_basedn   = "o=OSG,o=CO,dc=cilogon,dc=org"
+
+
+def get_cilogon_ldap_id_map(ldappass):
+    """ return dict of cilogon ldap data for each CILogonID, with the
+        structure: {CILogonID: { "dn": dn, "data": data }, ...} """
+    server = ldap3.Server(_ldap_url)
+    conn = ldap3.Connection(server, _username, ldappass)
+    if not conn.bind():
+        return None  # connection failure
+    conn.search(_basedn, '(voPersonID=*)', attributes=['*'])
+    result_data = [ (e.entry_dn, e.entry_attributes_as_dict)
+                    for e in conn.entries ]
+    conn.unbind()
+    return {
+        voPersonID: { "dn": dn, "data": data }
+        for dn, data in result_data
+        if "voPersonID" in data
+        for voPersonID  in data["voPersonID"]
+    }
+
+
+def cilogon_id_map_to_ssh_keys(m):
+    """ convert id map (as returned by get_cilogon_ldap_id_map) to a dict with
+        structure: {CILogonID: [sshPublicKey, ...], ...} for each id that has
+        ssh public keys defined """
+    return {
+        k: v['data']['sshPublicKey']
+        for k, v in m.items()
+        if 'sshPublicKey' in v['data']
+    }
+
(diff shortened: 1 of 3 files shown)


src/webapp/cilogon_ldap.py
```diff
@@ -0,0 +1,44 @@
...
+def get_cilogon_ldap_id_map(ldappass):
+    """ return dict of cilogon ldap data for each CILogonID, with the
+        structure: {CILogonID: { "dn": dn, "data": data }, ...} """
+    server = ldap3.Server(_ldap_url)
+    conn = ldap3.Connection(server, _username, ldappass)
+    if not conn.bind():
+        return None  # connection failure
+    conn.search(_basedn, '(voPersonID=*)', attributes=['*'])
+    result_data = [ (e.entry_dn, e.entry_attributes_as_dict)
+                    for e in conn.entries ]
```
brianhlin, line 25: Is the `entry_dn` really a separate attr from the rest of `entry_attributes_as_dict`? What sort of info are we grabbing from `entry_attributes_as_dict`?

src/webapp/cilogon_ldap.py
```diff
@@ -0,0 +1,44 @@
+import ldap3
+
+
+def get_contact_cilogon_id_map(global_data):
+    """ return contacts dict, limited to users with a CILogonID """
+    contacts = global_data.get_contacts_data().users_by_id;
+    return { k: v for k, v in contacts.items() if v.cilogon_id is not None }
```
brianhlin, line 7: I think this is a function that we want to drop in the future, i.e. we want to grab contact info directly from CILogon and drop the contact db

brianhlin (approved):
Simple reorganization LGTM but I've got a question about the structure of the CILogon ID map

src/webapp/cilogon_ldap.py
edquist: Yeah so, `conn.entries` here is a list of `ldap3.abstract.entry.Entry` objects returned by the `conn.search(...)` call.

Each of these entries has a dn and an "attributes", which can be accessed as a dict.  For instance calling `e.entry_to_json()` on a single entry results in

```
{
    "attributes": {
        "cn": [
            "Brian Bockelman"
        ],
        "eduPersonOrcid": [
            "0000-0003-2981-3809"
        ],
        "eduPersonPrincipalName": [
            "bockelman@wisc.edu"
        ],
        "givenName": [
            "Brian"
        ],
        "isMemberOf": [
            "CO:members:all",
            "CO:members:active",
            "CO:admins",
            "hub-admins",
            "CO:COU:SOTERIA:members:active",
            "CO:COU:SOTERIA:members:all",
            "soteria-brian-bockelman"
        ],
        "mail": [
            "BBOCKELMAN@MORGRIDGE.ORG"
        ],
        "objectClass": [
            "person",
            "organizationalPerson",
            "inetOrgPerson",
            "eduPerson",
            "eduMember",
            "voPerson"
        ],
        "sn": [
            "Bockelman"
        ],
        "uid": [
            "http://cilogon.org/serverA/users/9265706"
        ],
        "voPersonID": [
            "OSG1000001"
        ]
    },
    "dn": "voPersonID=OSG1000001,ou=people,o=OSG,o=CO,dc=cilogon,dc=org"
}
```

...

> What sort of info are we grabbing from `entry_attributes_as_dict`?

This function returns a map of `voPersonID: { "dn": dn, "data": data }` for each entry, where `data` is the attributes dict.  (The voPersonID is the CILogonID.)

I suspect i included the "dn" from the entry here because it was not otherwise present in the "attributes" data.  Though as far as i can tell we don't actually use the returned dn here.

Separately (and maybe the only part that's currently used), `cilogon_id_map_to_ssh_keys` takes that returned map and grabs the `sshPublicKey` attr, transforming it into a map of `voPersonID: [sshPublicKey, ...]`.  This is used in oasis_managers to get the ssh public keys for vo managers.

src/webapp/cilogon_ldap.py
edquist: Sure and come to think of it maybe it doesn't really belong in the new cilogon_ldap module anyway. It just returns a subset of the global contacts_data users_by_id map, for users with a cilogon_id.

And yeah it will have to go away when the contact db does.

brianhlin: I think it's fine to move it here for now

src/webapp/cilogon_ldap.py
brianhlin: Ah ok, that makes sense. I was curious about why the DN was separate but this seems like a fine way to organize things

PR #1945, opened 2021-07-14
Update UMN-CMS.yaml

sull0153:
Retiring cache01 cache02.spa.umn.edu and replacing with new servers cse-app-squid-01.cse.umn.edu and cse-app-squid-02.cse.umn.edu


brianhlin (approved):
LGTM!

PR #1947, opened 2021-07-15
Adding Downtime Generator for Entire Sites (SOFTWARE-3526)

CannonLock:


brianhlin:
Closed in favor of #1947

matyasselmeci:
I am reopening this so I can review it; #1948 has way too many unrelated commits

src/webapp/forms.py
```diff
@@ -8,6 +8,94 @@
...
+    def get_start_datetime(self):
+        return datetime.datetime.combine(self.start_date.data, self.start_time.data)
+
+    def get_end_datetime(self):
+        return datetime.datetime.combine(self.end_date.data, self.end_time.data)
+
+    def get_yaml(self, resources, services) -> str:
+        yaml = ""
+        for resource, service in zip(resources, services):
+            yaml += models.get_downtime_yaml(
```
matyasselmeci, line 85: The downtime entry for each _resource_ must have a unique ID -- I don't see any code for dealing with that.

I suggest:
- Modify `models.get_downtime_yaml()` to take an ID as an optional parameter
- Call `models._dtid()` before the loop to get the initial ID
- Pass the ID to `get_downtime_yaml()`
- Increment the ID with each iteration of the loop

matyasselmeci (changes requested):
Overall not bad but there are several changes required:
- This is a downtime generator for a ResourceGroup, not a Site so you will have to rename everything reflect that. Sorry :-(
- There is no code for handling ID uniqueness (see my comment for a suggested strategy)
- Some minor fixups and changes for clarity

matyasselmeci (approved):
You should rename the class as well to `GenerateResourceGroupDowntimeForm` but otherwise it's good.

PR #1951, opened 2021-07-15
create cached object for cilogon ldap data (SOFTWARE-4528)

edquist:
for consistency, the contacts_reader.ContactsData class is entirely re-used for the cilogon/comanage data from ldap, which is accomplished here by transforming the relevant parts to match the structure from contacts.yaml.

still WIP at the moment but flaming criticism is welcome, i think.

(and yes at the moment plumbing needs to be added to read the ldap passwd from disk)

### DIFF
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -42,3 +42,25 @@ def cilogon_id_map_to_ssh_keys(m):
         if 'sshPublicKey' in v['data']
     }
 
+
+def _entry2cinfo(entry):
+    ci = {'DNs': [entry['dn']]}
+    emails = entry['data'].get('mail')
+    if emails:
+        ci['PrimaryEmail'] = emails[0].lower()
+        if len(emails) >= 2:
+            ci['SecondaryEmail'] = emails[1].lower()
+    else:
+        ci['PrimaryEmail'] = None
+    return ci
+
+
+def cilogon_id_map_to_yaml_data(m):
+    return {
+        id_: {'CILogonID'          : id_,
+              'FullName'           : entry['data']['cn'],
+              'ContactInformation' : _entry2cinfo(entry)}
+        for id_, entry in m.items()
+    }
+
+
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -65,3 +65,10 @@ def cilogon_id_map_to_yaml_data(m):
     }
 
 
+def merge_yaml_data(yaml_data_main, yaml_data_secondary):
+    yd = dict(yaml_data_main)
+    for id_, contact in yaml_data_secondary.items():
+        if id_ not in yd:
+            yd[id_] = contact
+    return yd
+
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -51,7 +51,7 @@ def _entry2cinfo(entry):
         if len(emails) >= 2:
             ci['SecondaryEmail'] = emails[1].lower()
     else:
-        ci['PrimaryEmail'] = None
+        return None
     return ci
 
 
@@ -59,8 +59,9 @@ def cilogon_id_map_to_yaml_data(m):
     return {
         id_: {'CILogonID'          : id_,
               'FullName'           : entry['data']['cn'],
-              'ContactInformation' : _entry2cinfo(entry)}
+              'ContactInformation' : cinfo}
         for id_, entry in m.items()
+        for cinfo in [_entry2cinfo(entry)] if cinfo
     }
 
 
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -61,7 +61,7 @@ def cilogon_id_map_to_yaml_data(m):
         cinfo = _entry2cinfo(entry)
         if cinfo:
             data[id_] = {'CILogonID'          : id_,
-                         'FullName'           : entry['data']['cn'],
+                         'FullName'           : entry['data']['cn'][0],
                          'ContactInformation' : cinfo}
     return data
 
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -56,13 +56,14 @@ def _entry2cinfo(entry):
 
 
 def cilogon_id_map_to_yaml_data(m):
-    return {
-        id_: {'CILogonID'          : id_,
-              'FullName'           : entry['data']['cn'],
-              'ContactInformation' : cinfo}
-        for id_, entry in m.items()
-        for cinfo in [_entry2cinfo(entry)] if cinfo
-    }
+    data = {}
+    for id_, entry in m.items()
+        cinfo = _entry2cinfo(entry)
+        if cinfo:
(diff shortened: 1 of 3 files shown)


src/webapp/cilogon_ldap.py
```diff
@@ -42,3 +42,25 @@ def cilogon_id_map_to_ssh_keys(m):
...
 
+
+def _entry2cinfo(entry):
+    ci = {'DNs': [entry['dn']]}
+    emails = entry['data'].get('mail')
+    if emails:
+        ci['PrimaryEmail'] = emails[0].lower()
+        if len(emails) >= 2:
+            ci['SecondaryEmail'] = emails[1].lower()
+    else:
```
matyasselmeci, line 53: At least one email address is required; if you don't find one, put a warning in the log and skip the person.

matyasselmeci: note that you need an email address to register with COManage in the first place so if this branch gets hit, something is wrong...

edquist: Yes well, there happen to be two entries (both for Oksana Shadura) that do not have email addresses at the moment:

```
  "OSG1000041": {
    "data": {
      "cn": [
        "Oksana Shadura"
      ],
      "givenName": [
        "Oksana"
      ],
      "isMemberOf": [
        "CO:members:all",
        "CO:COU:OpenData Facility:members:all",
        "CO:members:active",
        "CO:COU:OpenData Facility:members:active",
        "opendata-facility-sponsors"
      ],
      "objectClass": [
        "person",
        "organizationalPerson",
        "inetOrgPerson",
        "eduMember",
        "voPerson"
      ],
      "sn": [
        "Shadura"
      ],
      "voPersonID": [
        "OSG1000041"
      ]
    },
    "dn": "voPersonID=OSG1000041,ou=people,o=OSG,o=CO,dc=cilogon,dc=org"
  },
```

```
  "OSG1000044": {
    "data": {
      "cn": [
        "Oksana Shadura"
      ],
      "givenName": [
        "Oksana"
      ],
      "isMemberOf": [
        "CO:members:all",
        "CO:COU:OpenData Facility:members:all",
        "CO:members:active",
        "CO:COU:OpenData Facility:members:active"
      ],
      "objectClass": [
        "person",
        "organizationalPerson",
        "inetOrgPerson",
        "eduMember",
        "voPerson"
      ],
      "sn": [
        "Shadura"
      ],
      "voPersonID": [
        "OSG1000044"
      ]
    },
    "dn": "voPersonID=OSG1000044,ou=people,o=OSG,o=CO,dc=cilogon,dc=org"
  }
```

edquist: > At least one email address is required; if you don't find one, put a warning in the log and skip the person.

Sounds good!

edquist:
@matyasselmeci, this is mostly ready to review; i've added logic to combine the two ContactsData sources (contact db and comanage ldap) into a third ContactsData (by merging the yaml_data's), and the merged ContactsData is now the one returned by get_contacts_data.  I _think_ this is effectively what we want, just i've reused the ContactsData class rather than making two separate parallel ones for the comanage and merged ones.

I realize now that i have not yet added the pluming for passing ldappass to the call to get_cilogon_ldap_id_map in models.py ([here](https://github.com/edquist/topology/blob/SOFTWARE-4528.breakout-cilogon-ldap2/src/webapp/models.py#L157)), which is why i am not unmarking this as a draft.  The cilogon ldap password _filename_ lives in global_data.cilogon_ldap_passfile, though i don't think we store the password itself in any variable.

Anyway that's where things are at at the moment.  Feel free to look things over for issues, etc.

src/webapp/models.py
```diff
@@ -146,6 +148,41 @@ def get_contacts_data(self) -> ContactsData:
 
         return self.contacts_data.data
 
+    def get_comanage_data(self) -> ContactsData:
```
matyasselmeci, line 151: Like with get_contact_db_data, there should be a config knob to skip this and return an empty ContactsData.

src/webapp/cilogon_ldap.py
```diff
@@ -42,3 +42,33 @@ def cilogon_id_map_to_ssh_keys(m):
...
+    return ci
+
+
+def cilogon_id_map_to_yaml_data(m):
+    return {
+        id_: {'CILogonID'          : id_,
+              'FullName'           : entry['data']['cn'],
+              'ContactInformation' : cinfo}
+        for id_, entry in m.items()
+        for cinfo in [_entry2cinfo(entry)] if cinfo
```
matyasselmeci, line 64: wait so is `[_entry2cinfo(entry)]` always going to be a 1-element list, with the element being either a dict or a None?

brianhlin:
Should we make a `CILOGON_LDAP_USER` config knob? I can imagine that'd be something that we want to configure, say, for pointing at the test COManage registry

src/webapp/cilogon_ldap.py
edquist: Yeah, exactly.  The point of 392fed4 was to exclude the whole contact if the entry lacks an email address.  It seemed like the most straightforward way to add that filtering to the dict comprehension (without a redundant `_entry2cinfo(entry)` call).  But feel free to suggest if you have something you like better in mind.

matyasselmeci: If you have to pull trickery like that in order to get the dict comprehension to work, then it's time to stop using a dict comprehension and do it the old fashioned way.
```
    ret = {}
    for id_, entry in m.items():
        cinfo = _entry2cinfo(entry)
        if not cinfo:
            continue
        ret[id_] = {'CILogonID'          : id,
                    'FullName'           : entry['data']['cn'],
                    'ContactInformation' : cinfo}
    return ret
```

edquist: Eh sure if that's how you'd like it.

For me, adding an extra `for var in [some_expression()]` line is kind of my standard idiom to assign an intermediate value to a var in a list/dict comprehension / generator expression.  It just feels more pythonic to me than building up an object in a regular for-loop.

(As a piece of trivial knowledge, apparently list comprehensions have been part of python since the year 2000.  I can't even comprehend how they did the special effects in The Matrix without that feature.)

edquist:
>`CILOGON_LDAP_USER`

yes and another for the ldap url

src/webapp/models.py
edquist: Updated - now if any of the `CILOGON_LDAP_*` vars are None, an empty ContactsData is returned.

(`CILOGON_LDAP_URL` & `_USER` have their the expected values as defaults; though `_PASSFILE` defaults to None)

src/webapp/cilogon_ldap.py
edquist: Updated to the Old Fashioned Burger

src/webapp/cilogon_ldap.py
edquist: contacts with no email are omitted now

edquist:
@matyasselmeci - i've added ldap password plumbing and i think i've addressed your comments.  Can you take another peek?

And if you'd like i can rebase out the "fixies".

matyasselmeci (approved):
LGTM

edquist:
Everything looks good in prod after the last fix.

PR #1953, opened 2021-07-16
Topology Cacher (SOFTWARE-4704)

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-4704


matyasselmeci:
OK, done making changes; this is ready for review.

matyasselmeci:
Added an RPM -- scratch build at https://koji.opensciencegrid.org/koji/taskinfo?taskID=343319

src/topology_cacher.py
```diff
@@ -0,0 +1,583 @@
...
+
+    args = parser.parse_args(argv[1:])
+    log.setLevel(
+        between(10 * (args.quiet - args.verbose), logging.DEBUG, logging.CRITICAL)
+    )
+
+    try:
+        os.makedirs(args.outdir, exist_ok=True)
+    except OSError as e:
+        pass  # ¯\_(ツ)_/¯
```
edquist, line 539: if you want to safe characters

```suggestion
        "¯\_(ツ)_/¯"
```

rpm/topology.spec
```diff
@@ -1,32 +1,58 @@
...
+%{python_sitelib}/topology_cacher.py*
+%{python_sitelib}/__pycache__/topology_cacher*
+%config(noreplace) /etc/cron.d/topology-cacher.cron
 
 
 %changelog
+* Wed Aug 04 2021 Mátyás Selmeci <matyas@cs.wisc.edu> 1.7.0
+- Add topology-cacher (SOFTWARE-4704)
+
 * Thu Mar 18 2021 Mátyás Selmeci <matyas@cs.wisc.edu> 1.4.6-1
```
edquist, line 56: that's quite the version bump.

src/topology_cacher.py
```diff
@@ -0,0 +1,583 @@
...
+        return self.SERVICE_ID_SCHEDD in self.service_ids
+
+    def to_dict(self) -> dict:
+        return dict(
+            group_name=self.group_name,
+            name=self.name,
+            fqdn=self.fqdn,
+            service_ids=self.service_ids,
+            tags=self.tags,
+        )
```
edquist, line 185: is there a reason why you implement `to_dict` here, rather than using the `_asdict` method that comes with `namedtuple` and as far as i can tell accomplishes the same thing?

src/topology_cacher.py
```diff
@@ -0,0 +1,583 @@
...
+            for eResource in eResourceGroup.findall("./Resources/Resource"):
+                resource_name = safe_element_text(eResource.find("./Name"))
+                fqdn = safe_element_text(eResource.find("./FQDN"))
+                service_ids = list(
+                    filter(
+                        None,
+                        [
+                            safe_element_text(svc)
+                            for svc in eResource.findall("./Services/Service/ID")
+                        ],
```
edquist, line 234: idk, but,

```suggestion
                        map(safe_element_text, eResource.findall("./Services/Service/ID"))
```

src/topology_cacher.py
```diff
@@ -0,0 +1,583 @@
...
+                    continue
+                tags = list(
+                    filter(
+                        None,
+                        [
+                            safe_element_text(tag)
+                            for tag in eResource.findall("./Tags/Tag")
+                        ],
+                    )
+                )
```
edquist, line 248: since this is a pattern,
```
def findall_nonempty(elt, path):
    return list(filter(None, map(safe_element_text, elt.findall(path))))
```

and then

```suggestion
                tags = findall_nonempty(eResource, "./Tags/Tag")
```

edquist (commented):
Some comments so far, still have more to review

rpm/topology.spec
matyasselmeci: Yeah this RPM is weird. It only contains topology-utils (and now this) but has the same version as the topology webapp (which we don't actually put into an RPM).

src/topology_cacher.py
matyasselmeci: Did not know about that. Thanks, I was wondering why I had to write that myself...

src/topology_cacher.py
```diff
@@ -0,0 +1,567 @@
...
+
+            for eResource in eResourceGroup.findall("./Resources/Resource"):
+                resource_name = safe_element_text(eResource.find("./Name"))
+                fqdn = safe_element_text(eResource.find("./FQDN"))
+                service_ids = list(
+                    filter(
+                        None,
+                        map(safe_element_text, eResource.findall("./Services/Service/ID"))
+                    )
+                )
```
edquist, line 224: erp, sorry i had meant for you to use findall_nonempty here also:

```suggestion
                service_ids = findall_nonempty(eResource, "./Services/Service/ID")
```

... i mean, it is nice, isn't it?

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+                        )
+                        continue
+
+                    allocation["submit_resources"].append(
+                        {
+                            "fqdn": resinfo.fqdn,
+                            "group_name": resinfo.group_name,
+                            "name": resinfo.name,
+                        }
+                    )
```
edquist, line 385: i guess you don't want to include the whole resinfo dict here

edquist (approved):
I left some comments but feel free to interpret them at your own discretion.

The fact that it works for you as-is is probably good enough to go on (for me).

Though i imaging you might find it satisfying to break up some of the big functions (like `get_project_resource_allocations`).

topology-cacher.cron
```diff
@@ -0,0 +1,2 @@
+SHELL=/bin/bash
+0,15,30,45 * * * * root sleep $(( $RANDOM \% 180 )) && /usr/bin/python3 -m topology_cacher
```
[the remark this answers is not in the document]
matyasselmeci: Yeah this is meant to run in containers which don't have systemd.
Maybe one day someone will come up with a daemon that runs systemd timers without needing the rest of systemd. Until then, back to stone knives and bearskins :-)

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+    path = ""
+    try:
+        for filename, contents in [
+            ("miscproject.xml", data.projects),
+            ("rgsummary.xml", data.resources),
+        ]:
+            path = os.path.join(args.outdir, filename)
+            with open(path, "w", encoding="utf-8") as fh:
+                fh.write(elem2str(contents))
+                log.info("Wrote %s", path)
```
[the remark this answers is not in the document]
matyasselmeci: Yep. I don't want to save the downloaded XML if it's broken/unparseable.

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+    parser.add_argument(
+        "--topology",
+        metavar="URL",
+        default=TOPOLOGY,
+        help="Base URL of the Topology service. [%(default)s]",
+    )
+
+    args = parser.parse_args(argv[1:])
+    log.setLevel(
+        between(10 * (args.quiet - args.verbose), logging.DEBUG, logging.CRITICAL)
```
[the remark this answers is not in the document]
matyasselmeci: Derp. Fixed.

src/topology_cacher.py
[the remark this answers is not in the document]
edquist: Ah!  Now that's thinking outside the box!

topology-cacher.cron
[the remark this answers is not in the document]
edquist: > stone knives and bearskins

Um, rock on?  :)

src/topology_cacher.py
matyasselmeci: I don't, not that much :-)

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+        self.resinfo_by_name = {}
+        self.resinfo_by_fqdn = {}
+        self.projects = self.get_projects()
+        self.resources = self.get_resources()
+
+    def get_projects(self) -> ET.Element:
+        return self._get_data("/miscproject/xml")
+
+    def get_resources(self) -> ET.Element:
+        data = self._get_data("/rgsummary/xml")
```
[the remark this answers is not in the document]
matyasselmeci: Yep, you're right. Fixed.

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+                    (eExecuteResourceGroup_list, "ExecuteResourceGroups"),
+                    (eSubmitResource_list, "SubmitResources"),
+                ]:
+                    if not var:
+                        log.warning(
+                            "ResourceAllocation has a missing or empty %s: %s",
+                            name,
+                            elem2str(eResourceAllocation),
+                        )
+                        continue
```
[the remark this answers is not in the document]
matyasselmeci: Crap, I wanted to `continue` the outer for loop. `next LABEL` was nice in Perl.

src/topology_cacher.py
```diff
@@ -0,0 +1,562 @@
...
+                            "group_name": group_name,
+                            "local_allocation_id": local_allocation_id,
+                        }
+                    )
+
+                # Done with this allocation
+                allocations.append(allocation)
+
+        # Done with all projects
+        return ret
```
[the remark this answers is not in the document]
matyasselmeci: Yeah good point. When I have a little more time I'll try and break this function up.

I think some of the info in the docstrings would be better off in the README.md file so I'll do that as well.

PR #1955, opened 2021-07-20
add downtime for MWT2_CE_UIUC due to UIUC planned maintenance

mollycule:


LincolnBryant:
I see that Jess isn't associated with this resource - we'll work on fixing in that in the future. In the meantime 👍  from me.

brianhlin (approved):
LGTM

PR #1962, opened 2021-07-22
Add downtime for Purdue-Geddes due to power maintenance

smithnp:


matyasselmeci:
Hi Nick,
The only person listed as a contact for Purdue-Geddes is Erik Gough. Could you please send a request to be added to help@opensciencegrid.org, and CC Erik on the email?

PR #1969, opened 2021-07-26
Convert the StashCache module to ldap3

brianhlin:
In preparation for adding LIGO LDAP credentials

I did a diff of the output between the old version of `_generate_ligo_dns` vs the new one and the list was exactly the same. Still working on a full local test.

### DIFF
--- a/src/stashcache.py
+++ b/src/stashcache.py
@@ -49,9 +49,9 @@ def _generate_ligo_dns() -> List[str]:
     query = "(&(isMemberOf=Communities:LSCVirgoLIGOGroupMembers)(gridX509subject=*))"
     try:
         server = ldap3.Server(_ligo_ldap_url, connect_timeout=10)
-        conn = ldap3.Connection(server, receive_timeout=10)
+        conn = ldap3.Connection(server, raise_exceptions=True, receive_timeout=10)
         conn.bind()
-    except ldap3.core.exceptions.LDAPExceptionError:
+    except ldap3.core.exceptions.LDAPException:
         log.exception("Failed to connect to the LIGO LDAP")
         return []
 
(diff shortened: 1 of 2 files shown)


matyasselmeci:
Looks good to me though the only lines of that code that were written by me were the ones that set the timeout :-P

You might want to ask @edquist for his opinion since he has more LDAP experience.

brianhlin:
@matyasselmeci unfortunately he's out this week so your review will have to do since we need to get this sorted out for OPS-122 by Sunday

matyasselmeci:
I was Slacking with him this morning so I completely forgot that he's out. OK, I'll take a look.

matyasselmeci:
Looks good to me so far.

src/stashcache.py
```diff
@@ -52,22 +42,31 @@ class NotRegistered(Exception):
...
+        conn.bind()
+    except ldap3.core.exceptions.LDAPExceptionError:
+        log.exception("Failed to connect to the LIGO LDAP")
+        return []
+
+    try:
+        conn.search(_ligo_ldap_branch,
+                    query,
+                    search_scope='LEVEL',
+                    attributes=['gridX509subject'])
```
edquist, line 62: The only thing i'll mention here is that i remember _not_ getting back the stuff i wanted when i explicitly listed the attributes i wanted, which led me just to request `attributes=['*']` instead (which gave more than i cared about, but at least not less than i cared about).

I would suggest trying it out both ways and just confirm you are not missing anything you care about with the explicit `attributes` list.

edquist: Here, i mean
https://github.com/opensciencegrid/topology/blob/master/src/webapp/cilogon_ldap.py#L23

brianhlin: This worked for me --maybe the issue you ran into was with requesting multiple attrs?

brianhlin:
Topology ITB is running a container built from this branch so I think this is ready to go @matyasselmeci

brianhlin:
Force pushed to rebase on `master`

PR #1971, opened 2021-07-28
Update UFlorida-HPC.yaml to change the xrootd endpoint

bockjoo:
Redirector is changed from cmsio5.rc.ufl.edu to cmsio7.rc.ufl.edu due to the down cmsio5.


brianhlin (approved):
LGTM

PR #1975, opened 2021-08-04
Add push (SOFTWARE-4718) and add context (SOFTWARE-4724)

mtru32:


.github/workflows/build-sw-container.yml
```diff
@@ -16,10 +16,21 @@ jobs:
...
+        tag_list=()
+        for registry in hub.opensciencegrid.org docker.io; do
+          for image_tag in "$REPO" "$REPO-$TIMESTAMP"; do
+            tag_list+=("$registry/$docker_repo":"$image_tag")
+          done
+        done
+        # This causes the tag_list array to be comma-separated below,
+        # which is required for build-push-action
+        IFS=,
+        echo "::set-output name=taglist::${tag_list[*]}"
```
brianhlin, line 33: This list of tags is very different from what we had previously (`opensciencegrid/topology:latest` and `opensciencegrid/topology:X.Y.Z` where `X.Y.Z` is a version number)

brianhlin (changes requested):
Hey @mtru32 this repo pushes a non-standard tag list so you'll need to revisit this PR

brianhlin:
@mtru32 could you revert the last commit? I don't think I actually want to add the ability to build a container at the push of a button. I also don't think it'll play nicely with the `GITHUB_REF`-based tag

brianhlin (approved):
LGTM

PR #1987, opened 2021-08-13
OSG projects for Christina's 2021 Virtual School mentees

ChristinaLK:


projects/CSM_BeEST.yaml
```diff
@@ -0,0 +1,12 @@
+Name: CSM_BeEST
```
matyasselmeci, line 1: In the past you've used "Mines_" as the prefix for Colorado School of Mines projects.

matyasselmeci:
Odd that osg-bot didn't say anything about "University of Edinburgh" and "University of Sheffield" but they're both legit new organizations so that's fine; I'll add them to the mappings file after I merge this.

Is it too late to change `CSM_BeEST` to `Mines_BeEST`?

ChristinaLK:
CSM_BeAST is the name of the project on the OSG account management side, so I think it has to match, right?

matyasselmeci:
I don't know; you'll have to ask them. (Would that be Mats?)

matyasselmeci (approved):

PR #2003, opened 2021-08-17
Add LIGO LDAP credentials (OPS-122)

brianhlin:
I went with this to avoid touching too much code. Not sure if you guys prefer reading the passfile in `models.py`

### DIFF
--- a/src/stashcache.py
+++ b/src/stashcache.py
@@ -7,6 +7,8 @@ import ldap3
 import asn1
 import hashlib
 
+from webapp.common import readfile
+from webapp.models import GlobalData
 from webapp.topology import Resource, ResourceGroup
 from webapp.vos_data import VOsData
 
@@ -30,7 +32,6 @@ __oid_map = {
 
 __dn_split_re = re.compile("/([A-Za-z]+)=")
 
-_ligo_ldap_url    = "ldaps://ldap.ligo.org"
 _ligo_ldap_branch = "ou=people,dc=ligo,dc=org"
 
 class DataError(Exception):
@@ -40,7 +41,7 @@ class NotRegistered(Exception):
     """Raised when the FQDN is not registered at all"""
 
 
-def _generate_ligo_dns() -> List[str]:
+def _generate_ligo_dns(ldapurl: str, ldapuser: str, ldappass: str) -> List[str]:
     """
     Query the LIGO LDAP server for all grid DNs in the IGWN collab.
 
@@ -48,8 +49,8 @@ def _generate_ligo_dns() -> List[str]:
     """
     query = "(&(isMemberOf=Communities:LSCVirgoLIGOGroupMembers)(gridX509subject=*))"
     try:
-        server = ldap3.Server(_ligo_ldap_url, connect_timeout=10)
-        conn = ldap3.Connection(server, raise_exceptions=True, receive_timeout=10)
+        server = ldap3.Server(ldapurl, connect_timeout=10)
+        conn = ldap3.Connection(server, user=ldapuser, password=ldappass, raise_exceptions=True, receive_timeout=10)
         conn.bind()
     except ldap3.core.exceptions.LDAPException:
         log.exception("Failed to connect to the LIGO LDAP")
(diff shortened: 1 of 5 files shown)


src/stashcache.py
```diff
@@ -178,19 +179,24 @@ def _cache_is_allowed(resource, vo_name, stashcache_data, public, suppress_error
     return ret
 
 
-def generate_cache_authfile(vo_data: VOsData, resource_groups: List[ResourceGroup], fqdn=None, legacy=True, suppress_errors=True) -> str:
+def generate_cache_authfile(global_data: GlobalData,
```
edquist, line 182: Do you need to give similar treatment to `generate_public_cache_authfile` also, to handle the `if public_only` block here?

https://github.com/opensciencegrid/topology/blob/e2d34ad34579f974c5467dc93617d8bd299fe30a/src/app.py#L253-L256

brianhlin: Yup! I'll get right on it

brianhlin:
Force pushed to fix a few issues and rebase on master

brianhlin:
Ignore the base branch swaps. I just wanted to test out the new GitHub functionality to change base branches and @edquist had questions about changes in wording :)

src/tests/verify_authfile.sh
```diff
@@ -2,4 +2,4 @@
 
 cd "$(dirname "$0")/../../"
 export PYTHONPATH=src/:$PYTHONPATH
-exec ./bin/osg-authfile > /dev/null
+exec ./bin/osg-authfile > /dev/null 2>&1
```
edquist, line 5: So, you want to hide a traceback when there are errors?

brianhlin: Yeah, AIUI, these tracebacks would be logged but since we're running the functions manually, they get dumped to `stderr`. The functions themselves handle these particular exceptions and so they exit zero, we just want to throw away their out (I forget why, @matyasselmeci ?)  and err because GitHub Actions apparently look for these tracebacks and "annotate" the test run with an error but otherwise lets the test pass because the script exist zero. See https://github.com/brianhlin/topology/actions/runs/1144133515. Exceptions that we aren't catching should still cause the script to exit non-zero.

**tl;dr** yeah, we want to throw these tracebacks away since we catch them otherwise they just lead to confusing GHA results

edquist: Well i admit i am confused.

But let's see.  IIUC:

- there are uncaught exceptions
- these produce tracebacks on stderr
- GHA uses stderr output to annotate results
- these GHA runs still pass because (for some reason the script) still exits `0`
- we throw out stderr entirely to avoid the confusing GHA annotations with such "passing" jobs
- uncaught exceptions should still cause the script to exit non-zero

So, for this guy,

> ```Error:     raise LDAPPasswordIsMandatoryError('password is mandatory in simple bind')```

is this an exception we catch & handle, but print the traceback anyway, or is it an uncaught exception that _should_ cause the script to exit non-zero, but doesn't?

brianhlin:
> is this an exception we catch & handle, but print the traceback anyway, or is it an uncaught exception that should cause the script to exit non-zero, but doesn't?

The former since `LDAPPasswordIsMandatoryError` is a child of `LDAPException`. It's being dumped to stderr because we use `log.exception()` without a logfile destination in the case that these functions are being called directly.

> these GHA runs still pass because (for some reason the script) still exits 0

We ignore these errors because we still can produce a partial stash cache/origin authfile without the info from LDAP. So if we encounter this, we just want to log and move on.

edquist:
Ok, thank you for the explanation.

I guess the only downside then is you also lose the stderr from any uncaught exceptions.  They will still fail (exit non-zero) as you said, but they'd be harder to diagnose without the traceback.

matyasselmeci:
> We ignore these errors because we still can produce a partial stash cache/origin authfile without the info from LDAP.

We may want to revisit this idea at some point -- I don't like the thought of service authfiles going back and forth between "LIGO" and "no LIGO" versions in case their LDAP server is flakey.

How about having separate endpoints for the LIGO and the not-LIGO parts of the config and having the authfile-update timer combine the two? We already have code for combining authfiles here: https://github.com/opensciencegrid/xcache/pull/117/files

brianhlin:
I think that's a good idea but a separate endpoint + combining authfiles would still have you oscillating between LIGO/no LIGO in the case of flaky LDAP. The only way around that I can think of is keeping a LIGO authfile cache around and using that if LDAP queries fail.

Can we handle that in a different ticket/PR, though?

matyasselmeci:
Oh yeah totally a different PR.

It wouldn't go back and forth -- like what it does with the Authfile, authfile-update will keep its previous copy of Authfile-ligo if downloading from the LIGO endpoint fails. (So yeah it's sorta like a cache, except client-side.)

src/tests/verify_authfile.sh
brianhlin: Oops, I broke off the comments into the main PR thread. See that for the explanation

edquist (approved):
Godspeed

brianhlin:
Force pushed to rebase on master

PR #2010, opened 2021-08-18
Take project Name from filename if missing; warn if the Name and the filename do not match

matyasselmeci:
SOFTWARE-4755

### DIFF
--- a/src/webapp/project_reader.py
+++ b/src/webapp/project_reader.py
@@ -65,6 +65,14 @@ def get_one_project(file: str, campus_grid_ids: Dict, vos_data: VOsData) -> Dict
             data['ResourceAllocations'] = {"ResourceAllocation": resource_allocations}
         if 'ID' not in data:
             del project['ID']
+
+        name_from_filename = os.path.basename(file)[:-5]  # strip '.yaml'
+        if not is_null(data, 'Name'):
+            if data['Name'] != name_from_filename:
+                log.warning("%s: 'Name' %r does not match filename" % (file, data['Name']))
+        else:
+            data['Name'] = name_from_filename
+
     except Exception as e:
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Also, these are the warnings it found (edited for brevity):
```
projects/UCAnschutz_Langner.yaml: 'Name' 'UCAnshutz_Langner' does not match filename
projects/GSU_Maimon.yaml: 'Name' 'GSU_Wang' does not match filename
projects/TNTTech_ITS.yaml: 'Name' 'TNTech_ITS' does not match filename
projects/PRP_Defanti.yaml: 'Name' 'PRP' does not match filename
projects/NMSU_staff.yaml: 'Name' 'NMSU_Staff' does not match filename
```

brianhlin:
@lmichael107 @ChristinaLK @rynge @eharstad any idea which ones are the right project names from @matyasselmeci's comment above?

src/webapp/project_reader.py
```diff
@@ -65,6 +65,14 @@ def get_one_project(file: str, campus_grid_ids: Dict, vos_data: VOsData) -> Dict
             data['ResourceAllocations'] = {"ResourceAllocation": resource_allocations}
         if 'ID' not in data:
             del project['ID']
+
+        name_from_filename = os.path.basename(file)[:-5]  # strip '.yaml'
```
brianhlin, line 69: What about `.yml`?

brianhlin (approved):
One comment but otherwise LGTM

src/webapp/project_reader.py
matyasselmeci: We don't read .yml files; this goes for the VO and resource readers as well.

PR #2016, opened 2021-08-20
Various project template changes

matyasselmeci:
- Drop "Name"
- Make "Description" a YAML multi-line string (avoids syntax errors from `:` characters
- Add a template ResourceAllocations section

### DIFF
--- a/template-project.yaml
+++ b/template-project.yaml
@@ -30,12 +30,12 @@ Sponsor:
 ### Contact the Software Team for assistance in filling in these values.
 # ResourceAllocations:
 
-##### Type is how the allocation was obtained: "XRAC" for allocations though
+##### Type is how the allocation was obtained: "XRAC" for allocations through
 ##### XSEDE or "Other" for allocations directly through the site.
 #   - Type: Other
 
 ####### SubmitResources is a list of Submit Nodes (Access Points) that users of
-####### project can submit from.  There must be at least one; each line must
+####### the project can submit from.  There must be at least one; each line must
 ####### refer to a Topology Resource.
 #     SubmitResources:
 #       - UChicago_OSGConnect_login04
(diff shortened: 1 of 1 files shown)


matyasselmeci:
The ResourceAllocations section is nasty enough that I added the line 'please ask Software for help' since I don't expect most facilitators to dig through `/topology` to do Resource/ResourceGroup lookups. We should consider creating some sort of mapping they can use instead.

matyasselmeci:
by mapping I mean like an endpoint for a human readable "submit node FQDN" <-> "submit node Resource Name", and also one for "HPC Site"<->"ResourceGroup"

where "HPC Site" is the shortened name of a site that we (will) tell people to put in their +Desired_Allocations list

template-project.yaml
```diff
@@ -21,3 +22,32 @@ Sponsor:
...
   #   Name: <VO>
+
+
+### Uncomment the ResourceAllocations block if this project can make use of
+### one or more XRAC (XSEDE) or other HPC allocations.
+###
+### Contact the Software Team for assistance in filling in these values.
+# ResourceAllocations:
+
+##### Type is how the allocation was obtained: "XRAC" for allocations though
```
brianhlin, line 33:
```suggestion
##### Type is how the allocation was obtained: "XRAC" for allocations through
```

template-project.yaml
```diff
@@ -21,3 +22,32 @@ Sponsor:
...
+##### XSEDE or "Other" for allocations directly through the site.
+#   - Type: Other
+
+####### SubmitResources is a list of Submit Nodes (Access Points) that users of
+####### project can submit from.  There must be at least one; each line must
```
brianhlin, line 38:
```suggestion
####### the project can submit from.  There must be at least one; each line must
```

brianhlin (approved):
LGTM

PR #2017, opened 2021-08-20
Add robot DNs to cache authfiles (SOFTWARE-4767)

brianhlin:
Completely untested but this should work


djw8605 (approved):
All looks good to me.  this is still draft though?

PR #2018, opened 2021-08-21
Drop Name field from projects where it matches the filename

matyasselmeci:


brianhlin (approved):
Sounds good to me -- can't say that I looked very closely at all 839 projects, though

PR #2028, opened 2021-08-27
Add the /ospool/PROTECTED namespace which will be served out of the new authenticated origin being set up at CHTC

matyasselmeci:


bbockelm (approved):
I don't particularly care for `/ospool` myself but, given the lack of a better idea, I'm going to go ahead and approve.  We can always change it easily up until the second a user starts writing scripts.

PR #2034, opened 2021-09-01
More OASIS cleanup (SOFTWARE-4279)

brianhlin:
This PR includes

1. Removal of OASIS managers that did not register in COManage by the deadline
2. Setting `UseOasis: false` for VOs with empty dris and without any managers (or the only remaining managers were BrianB and Derek)

There are a bunch of VOs with `UseOASIS: true` with empty dirs that I left around because they've got a COManage-registered manager


DrDaveD (approved):
LGTM

PR #2060, opened 2021-09-28
update AMNH-Mendel resource name and FQDN

jdost321:


djw8605:
I don't know about this change.  It may have ramifications in GRACC.

jdost321:
could have sworn I've changed host names and resource names without issue in the past.. Let me just disable it and make an ew one so it looks like a new resource.

matyasselmeci:
You need to disable and make a new one when a resource has records in GRACC in the past two years. [Kibana link](https://gracc.opensciencegrid.org/kibana/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2y,to:now))&_a=(columns:!(),filters:!(),index:'373839d0-634f-11eb-a078-6f5db37402b1',interval:auto,query:(language:kuery,query:'ProbeName:*%5C:hosted-ce39.opensciencegrid.org'),sort:!(!(EndTime,desc))))

(BTW Kibana says I'm not authorized to make short URLs.)

topology/American Museum of Natural History/AMNH/AMNH.yaml
```diff
@@ -288,12 +288,49 @@ Resources:
...
+    Tags:
+      - CC*
+
   AMNH-Mendel-CE1:
     Active: true
     Description: Third Hosted CE for the AMNH
     # If you have an up-to-date local git clone, fill ID with the output from `bin/next_resource_id`
     # Otherwise, leave it blank and we will fill in the appropriate value for you.
-    ID: 1111
+    ID:
```
matyasselmeci, line 333:
```suggestion
    ID: 1221
```

matyasselmeci (approved):

PR #2072, opened 2021-10-05
Added new project: UConn_Zhu

rynge:


projects/UConn_Zhu.yaml
```diff
@@ -0,0 +1,10 @@
+Description: GERS Laboratory working on big earth observation
+  data and environmental change
```
brianhlin, line 2:
```suggestion
Description: >
  GERS Laboratory working on big earth observation
  data and environmental change
```

Use YAML line continuations

brianhlin (approved):
One suggestion, otherwise LGTM

PR #2089, opened 2021-10-11
Add new PS servers iut2-net1[1-6]

luehring:
Added 6 more PS servers which are 3 physical servers using 2 NICs for PS. The 1G NIC does the latency tests and the 40G NIC does the bandwidth tests. I received 1228 from next_resource_id and assumed that the other 5 servers would get 1229-1233.


luehring:
Sorry Matyas... I think that I most likely have the same problem with this pull request as the last one.  I don't think that I have worked out the correct steps yet.

matyasselmeci:
Hi Fred, this looks good. I had to modify the IDs because I merged another PR first but that's not your fault.

PR #2104, opened 2021-10-21
* swap around FQDN and FQDNaliases until they are right. [rtj]

rjones30:


rjones30:
Brian, ok I added this update. Should a similar clause also be added to the
UConn-OSG.yaml?
-Richard J.

PR #2111, opened 2021-10-29
Distribute osg-notify as a container (SOFTWARE-4856)

RyanEweSeng:


Dockerfile.client
```diff
@@ -0,0 +1,17 @@
...
+                   epel-release \
+                   yum-utils
+
+RUN yum -y install python3 \
+                   python3-requests \
+                   python3-gnupg
+
+COPY topology/bin/osg-notify               /usr/local/bin
+COPY topology/src/net_name_addr_utils.py   /usr/lib/python3.6/site-packages
+COPY topology/src/topology_utils.py        /usr/lib/python3.6/site-packages
```
matyasselmeci, line 17: You should specify that the ENTRYPOINT is /usr/local/bin/osg-notify, so that is the command that will be run when people run the container. Use ENTRYPOINT (specifically, the 'exec form' of it) instead of CMD because people will need to pass arguments to the container.

https://docs.docker.com/engine/reference/builder/#entrypoint

matyasselmeci (changes requested):
See my comment about the ENTRYPOINT; also, I don't see where the workflow specifies that it's Dockerfile.client that should be used instead of the default Dockerfile.

Dockerfile.client
RyanEweSeng: How would I go about testing to make sure it does all of these correctly? I have made some changes but would like to test more before pushing.

matyasselmeci: My mistake, I didn't see that line. Usually ENTRYPOINT lines are near the end of the file.

.github/workflows/build-client-container.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Generate tag list
+      id: generate-tag-list
+      env:
+        TIMESTAMP: ${{ needs.make-date-tag.outputs.dtag }}
+      run: |
+        docker_repo=opensciencegrid/osg-notify
```
brianhlin, line 40:
```suggestion
        docker_repo=opensciencegrid/topology-notify
```

brianhlin (changes requested):
A few minor changes to make but mostly LGTM

brianhlin (approved):
LGTM now

PR #2132, opened 2021-11-15
Handle incorrect cert path or password in osg-notify (SOFTWARE-3914)

RyanEweSeng:


brianhlin (changes requested):
@RyanEweSeng this will resolve the bugs as outlined in the ticket but I'd like to see this handled slightly differently to align with best coding practices. Namely, you should raise exceptions in libraries instead of printing, allowing the caller to catch the exception and do the error handling. To do this, you should add an `AuthError` exception class based on `RuntimeError` (see https://github.com/opensciencegrid/osg-test/blob/master/osgtest/library/osgunittest.py#L32-L37 for an example) and raise `AuthError` with a message

Then in `osg-notify`, you'll catch this exception, perform any retries/checks that you need to, and print the error. We catch and print here, sometimes just printing the same message as passed along by the exception as to not scare off the user with stacktraces.

brianhlin:
@RyanEweSeng: @matyasselmeci pointed out a slightly better exception layout:

```
class Error(Exception):
    pass

class AuthError(Error):
    pass
```

With this, a caller could catch all `topology_utils` exceptions with `except toplology_utils.Error as exc`

src/topology_utils.py
```diff
@@ -223,17 +233,10 @@ def get_contacts(args, urltype, roottype):
...
-                response = session.get(url)
-            except OSError as exc:
-                attempts -= 1
-                if attempts > 0: print("Incorrect password, please try again")
-        if response is None:
-            print("Too many incorrect password attempts, exiting")
-            exit()
+        try:
+            response = session.get(url)
+        except OSError as exc:
```
matyasselmeci, line 238: There are _many_ reasons why an OSError might happen while getting a webpage, not all of which mean incorrect password. You should experiment to find out what specific error you get when you're using an incorrect password, and catch that specific error. I suspect it will be one of the others in this list: https://docs.python-requests.org/en/latest/api/#exceptions

(Wait until topology.opensciencegrid.org is back up before experimenting.)

bin/osg-notify
```diff
@@ -176,7 +176,20 @@ def main():
...
     recipients = set(args.recipients.split())
     if args.oim_recipients and 'vos' in args.oim_recipients:
-        results = topology_utils.get_vo_contacts(args)
+        attempts = 3
+        while attempts > 0:
+            try:
+                results = topology_utils.get_vo_contacts(args)
+            except topology_utils.InvalidPathError as exc:
+                print(exc)
+                exit()
```
matyasselmeci, line 185:
```suggestion
                exit(1)
```
You should use a non-zero exit code to indicate that an error happened.

bin/osg-notify
```diff
@@ -176,7 +176,20 @@ def main():
...
+            except topology_utils.IncorrectPasswordError as exc:
+                attempts -= 1
+                if attempts == 0:
+                    print("Too many incorrect password attempts, exiting")
+                    exit()
```
matyasselmeci, line 190:
```suggestion
                    exit(1)
```

bin/osg-notify
```diff
@@ -185,10 +198,23 @@ def main():
...
+        attempts = 3
+        while attempts > 0:
+            try:
+                if args.fqdn_filter:
+                    results = topology_utils.get_resource_contacts_by_fqdn(args)
+                else:
+                    results = topology_utils.get_resource_contacts(args)
+            except topology_utils.InvalidPathError as exc:
+                print(exc)
+                exit()
```
matyasselmeci, line 210:
```suggestion
                exit(1)
```

bin/osg-notify
```diff
@@ -185,10 +198,23 @@ def main():
...
+            except topology_utils.IncorrectPasswordError as exc:
+                attempts -= 1
+                if attempts == 0:
+                    print("Too many incorrect password attempts, exiting")
+                    exit()
```
matyasselmeci, line 215:
```suggestion
                    exit(1)
```

matyasselmeci (dismissed):
I left a few comments -- you should take a look (once topology.opensciencegrid.org is back up).

src/topology_utils.py
RyanEweSeng: Hey Mat, I experimented with the exceptions but the stacktrace keeps showing that OSError is the exception being thrown and the requests library raises ProtocolError and ConnectionError. I tried catching those exceptions but it doesn't seem to work.

bin/osg-notify
```diff
@@ -185,10 +198,23 @@ def main():
...
+        attempts = 3
+        while attempts > 0:
+            try:
+                if args.fqdn_filter:
+                    results = topology_utils.get_resource_contacts_by_fqdn(args)
+                else:
+                    results = topology_utils.get_resource_contacts(args)
+            except topology_utils.InvalidPathError as exc:
+                print(exc)
+                exit(1)
```
brianhlin, line 210:
```suggestion
                exit(str(exc))
```

This can be simplified. For example:

```
blin@blin-latitude:~$ python
Python 3.9.7 (default, Aug 30 2021, 00:00:00) 
[GCC 10.3.1 20210422 (Red Hat 10.3.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> exit('asdfasdf')
asdfasdf
blin@blin-latitude:~$ echo $?
1
```

bin/osg-notify
```diff
@@ -185,10 +198,23 @@ def main():
...
+            except topology_utils.IncorrectPasswordError as exc:
+                attempts -= 1
+                if attempts == 0:
+                    print("Too many incorrect password attempts, exiting")
+                    exit(1)
```
brianhlin, line 215:
```suggestion
                    exit("Too many incorrect password attempts, exiting")
```

brianhlin (approved):
Looks good! Two minor suggestions but I think we can get this merged in shortly

PR #2139, opened 2021-11-16
Deal with space between `FQAN:` and the FQAN

matyasselmeci:
Make sure that in auth lists for stash namespaces, 
```
- FQAN: /foobar
```
and
```
- FQAN:/foobar
```
both work.

(tested locally)


brianhlin (approved):
LGTM

PR #2142, opened 2021-11-16
new project from university of arkansas

ChristinaLK:
Feel free to edit the university name! I don't know if the Fayetteville location is relevant or not or if it should just be University of Arkansas.


matyasselmeci:
Their website just says "University of Arkansas"; by comparison, wisc.edu specifically says University of Wisconsin-Madison. So I think we can drop the Fayetteville location.

matyasselmeci (approved):

PR #2152, opened 2021-11-19
Updates to UConn-HPC and UConn-OSG site services

rjones30:


matyasselmeci (approved):

matyasselmeci:
This looks good to me; checks are failing because the topology service is down due to a cooling outage. I will rerun the checks once the service is back up.

PR #2169, opened 2021-12-01
Fix tag list in generated XML (SOFTWARE-4919)

brianhlin:
Verified this works with data from #2167


matyasselmeci (approved):
yep, that's what it should be

PR #2178, opened 2021-12-08
updating UCSDT2

ddavila0:


brianhlin:
@ddavila0 do you still want this to be a draft?

djw8605:
APEL numbers look good to me.

ddavila0:
I see that a validation has failed, seems that I need to get rid of all the Downtime records of UCSDT2-SE3 ? or should I instead do NOT remove the UCSDT2-SE3 resource and just mark it as Active=False ?

brianhlin:
@ddavila0 I'd do the latter unless there's some other goal you have in removing the entire resource

brianhlin (approved):
Looks fine to me but I can't say anything about its correctness. If you notice any issues @jdost321 , we can submit another PR to fix them up

PR #2181, opened 2021-12-12
Update LIGO.yaml

stuartthebruce:
Make top level /zenodo directory public

### DIFF
--- a/virtual-organizations/LIGO.yaml
+++ b/virtual-organizations/LIGO.yaml
@@ -82,6 +82,8 @@ DataFederations:
             Base Path: /user/ligo
       /gwdata:
         - PUBLIC
+      /zenodo:
+        - PUBLIC
     AllowedOrigins:
       - CIT_LIGO_ORIGIN
       - LIGO-StashCache-Origin


djw8605:
I will discuss this with the group.  Zenodo is not quite "unique" to your experiment.  What will the directory be used for?

stuartthebruce:
> I will discuss this with the group. Zenodo is not quite "unique" to your experiment. What will the directory be used for?

It is a mirror of all LIGO public data on Zenodo and the plan is to move the current gwdata/zeonodo directory up one level.

P.S. Too bad that each VO or Origin server does not have its own namespace. Is that on the roadmap for a future enhancement?

djw8605:
We tend to put each VO under it's own namespace.  For example, gluex is under `/gluex`, hcc under `/hcc`.  But, it is a global namespace, so we try police it the best we can.

stuartthebruce:
I am still confused, why is the next directory after /cvmfs/gwosc.osgstorage.org a global namespace that OSG needs to police? Doesn't `gwosc` was already scope this to Gravitational Wave Open Science? Note, I am asking to understand and we can indeed move `zenodo` down under `gwdata` or give it a more specific name. However, this is perhaps also an opportune time time consider the namespace as both LIGO and Virgo start operating their own Origin servers for their proprietary instrument data (before those data becomes public under `gwosc`).

@djw8605 do you think we can sort this out via github/slack, or should we put this on the agenda for a standing zoom call (or setup a short dedicated zoom call)?

PR #2183, opened 2021-12-13
Add downtime for SLATE_US_NMSU_AGGIE_GRID

nlvw:
The nodes hosting the OSG CE may be affected during Discovery's planned downtime so connection to AggieGrid could be partially affected during this time.


brianhlin (approved):
LGTM

PR #2213, opened 2022-01-06
NERSC-Perlmutter resource group (FD #68597)

matyasselmeci:


matyasselmeci:
Steve says it's intentional for both to have the same FQDN.

bbockelm:
NACK.  Resources should have a unique FQDN.  Existing duplicates are mistakes we should have eliminated by now and I don't want to dig the hole deeper.

retzkek:
Proposed FQDNs:

```
NERSC-Perlmutter-CPU: perlmutter-cpu.hepcloud.fnal.gov
NERSC-Perlmutter-GPU: perlmutter-gpu.hepcloud.fnal.gov 
```

matyasselmeci:
Thanks Kevin, I used those new FQDNs, listed the real ones in FQDNAliases and added a comment to explain. @bbockelm does it look better now?

brianhlin (approved):
LGTM, though I'm not 100% clear how the accounting works such that the separate resources are required

PR #2215, opened 2022-01-06
Update Cori artificial reporting FQDNs

retzkek:
Similar to what's being done for Perlmutter in #2213, this allows Haswell and KNL resources to be accounted separately.


matyasselmeci (approved):
LGTM. I didn't find any records >1 year in Kibana.

PR #2216, opened 2022-01-06
Remove Marian from the contact list of LIGO-StashCache-Origin as part of offboarding

matyasselmeci:


brianhlin (approved):
Thanks, not sure how I missed this one!

PR #2219, opened 2022-01-12
Update uconn HPC fqdn alias

djw8605:
GRACC isn't able to map osgce.storrs.hpc.uconn.edu to a facility because it is only in the FQDN alias.  Swap the fqdn alias and FQDN so GRACC can map it correctly.  The gratia probe ProbeName is set to "htcondor-ce:osgce.storrs.hpc.uconn.edu"


matyasselmeci:
You're changing both their Squid and their CE... is their Squid's FQDN causing issues too, or it this just for consistency?

djw8605:
Oh, there is no reason to change the squid, I was just overly aggressive on changing the FQDNs.

matyasselmeci:
Doesn't the Squid FQDN have to match what WLCG monitoring is seeing?

matyasselmeci (approved):

PR #2225, opened 2022-01-19
Add RDS_OSDF_CACHE Resource

DannySaba:
Adding an OSDF cache resource.

PR #2229, opened 2022-01-20
Fix the indentation of the RDS auth cache

brianhlin:
Fixes #2225


brianhlin:
Oops, meant to actually base it off of the other PR

PR #2232, opened 2022-01-20
Update compare-factory-config to exit non-zero on mismatch; remove usage of git module (SOFTWARE-4953)

RyanEweSeng:


bin/compare-factory-config.py
```diff
@@ -192,19 +190,18 @@ def find_non_topology_matches(gfactory_DB, topology_DB, resources):
...
     get_topology_data(topology_DB)
-    # cloning gfactory repository to a temporary directory
-    temp_dir = tempfile.mkdtemp()
-    git.Repo.clone_from(
-        'https://github.com/opensciencegrid/osg-gfactory',
-        to_path=temp_dir
-    )
+    # cloning user input osg-gfactory repository to a temporary directory
+    if len(argv) != 2:
+        print('Error: Invalid number of arguments\nUsage: compare-factory-config.py <GIT_REPO>')
```
brianhlin, line 201:
```suggestion
        print('Error: Invalid number of arguments\nUsage: compare-factory-config.py <FACTORY CONFIG GIT REPO DIR>')
```

`GIT_REPO` is a little ambiguous (it could be a URL like you pointed out on Slack)

bin/compare-factory-config.py
```diff
@@ -192,19 +190,18 @@ def find_non_topology_matches(gfactory_DB, topology_DB, resources):
...
+        exit(2)
+    temp_dir = argv[1]
```
brianhlin, line 203: This var should be renamed as it's provided by the dir and we shouldn't be treating it as temporary!

bin/compare-factory-config.py
```diff
@@ -237,6 +234,9 @@ def run(argv):
 
     shutil.rmtree(temp_dir, onerror=remove_readonly)  # file cleanup
```
brianhlin, line 235: Where treating the user's input as `temp_dir` bites us! I bet that you'd be pretty mad if a program ate all of the date that you provided to it.

brianhlin (changes requested):
There's one issue and I've got a few suggestions but otherwise this looks solid

PR #2246, opened 2022-01-24
Rename purdue-hammer-ce to purdue-hammer

djw8605:
Payload records are showing as site Purdue-Hammer.  This will rename the resource name.

This will cause a discontinuity 1 year ago with the ResourceName in GRACC, but we don't visualize ResourceName anywhere.  Resource name is also not used in WLCG reporting, only ResourceGroup.

Pilot (Batch) records match the ProbeName to the FQDN in topology, so this will not affect any matching there.


djw8605:
Downtime updated.

djw8605:
Could you take another look at this pull request.

PR #2248, opened 2022-01-26
Add FNAL VO issuers and subjects (SOFTWARE-4969)

brianhlin:
FYI @DrDaveD @jteheran, this info is used as the source for the `osg-scitokens-mapfile` package


DrDaveD (approved):
LGTM

PR #2251, opened 2022-01-27
Fix up contact ranks for GLOW:

matyasselmeci:
- always have a Primary
- don't have two contacts with the same rank


brianhlin (approved):
Good catch

PR #2263, opened 2022-02-04
Decommission AggieGrid Resource

nlvw:
The Aggie Grid cluster is being decommissioned so it needs to be removed, or disabled, in OSG.


brianhlin (approved):
LGTM, thanks!

PR #2282, opened 2022-02-15
Update DN

DannySaba:
Added quotes to DN.


matyasselmeci:
I think you'll need to change the outer quotes to single quotes.

matyasselmeci (approved):

PR #2294, opened 2022-02-21
Add SciToken issuer to Gluex VO

jasoncpatton:
Will use https://zeus.phys.uconn.edu as the issuer. Do not merge until https://zeus.phys.uconn.edu/.well-known/openid-configuration exists.


brianhlin (approved):
LGTM. I'm not sure why we shouldn't just merge this now: I think the only reason we would need to avoid merging this would be if we're not set on the FQDN. Is that the case @jasoncpatton ?

jasoncpatton:
> LGTM. I'm not sure why we shouldn't just merge this now: I think the only reason we would need to avoid merging this would be if we're not set on the FQDN. Is that the case @jasoncpatton ?

I think we're good there. Feel free to merge!

PR #2308, opened 2022-02-24
unifying host name

biozit:
change the name of osg.chic.nrp.internet2.edu to osg-chicago-stashcache.nrp.internet2.edu


brianhlin:
Hi @biozit, could you elaborate on what is being unified here?

biozit:
The Chicago cache has an alias on the DNS: 
host osg-chicago-stashcache.nrp.internet2.edu
osg-chicago-stashcache.nrp.internet2.edu is an alias for osg.chic.nrp.internet2.edu.
osg.chic.nrp.internet2.edu has address 163.253.70.2

The host on FQDN on this file is different from the CN and is not the same on the POD Hostname causing a problem to creating the auth file on the StashCache. 

CACHE_FQDN=${CACHE_FQDN:-$(hostname -f)} -> osg-chicago-stashcache.nrp.internet2.edu

Furthermore, I am trying to change the name on the Checkmk, using the same FQDN for all OSG.

biozit:
Hi Brain, please, do you need more information?

brianhlin:
Sounds reasonable!

FWIW, though, you should be able to specify the `CACHE_FQDN` env var in your pod and the authfile generator will use its value to do the Topology lookup.

PR #2314, opened 2022-02-25
* fix the typo in uri-override for the UConn-HPC_stash_origin service…

rjones30:
… [rtj]


brianhlin (approved):
LGTM but I'm not entirely sure if this makes much of a difference

PR #2321, opened 2022-02-28
Software 5051

CannonLock:
Attaches a CORS header so that I can request from webpage.

Restructures the return to comply with @brianhlin suggestion. 
- To do this I added some functionality to a function, but I checked and that same function is not used anywhere else in topology


CannonLock:
@matyasselmeci I can't request a review but would you mind taking a peak at this sometime?

matyasselmeci (approved):

matyasselmeci:
> To do this I added some functionality to a function, but I checked and that same function is not used anywhere else in topology

I'm betting that function was used when I was doing the initial XML->YAML conversion to import data from OIM (the system that topology replaced)

PR #2325, opened 2022-03-01
Update AMNH.yaml

ssgituser:
Copied and pasted duplicate ID for AMNH-Mendel-AP1. Please provide appropriate ID for resource


matyasselmeci (approved):

matyasselmeci:
Thanks for letting us know. The checks really should have caught that...

ssgituser:
No problem. Only noticed when I could find resource to download the token so that I could configure the new access point

PR #2343, opened 2022-03-07
* add new resource UConn-OSG_GWMS_Collector for gremlin.phys.uconn.ed…

rjones30:
…u [rtj]


rjones30:
There is a new resource being added in this PR, needs a new ID to be assigned.

matyasselmeci (approved):

PR #2351, opened 2022-03-10
Add a new resource group for the Jupyter service at CHTC

brianaydemir:


brianhlin:
I would also consider naming the group `CHTC-Jupyter` (or `CHTC-Jupyter-ITB` if it's for testing) then use the more specific name in the resource. Who knows, maybe we stand up some other OSG services related to JupyterHub!

brianaydemir:
I've renamed the group to the `CHTC-Jupyter`, and the resource to `User-EP`.

Offline, we discussed having these user EPs join the OSPool, even during testing. A `START` expression should prevent jobs from running on these EPs for all but the intended Jupyter users.

matyasselmeci (approved):

PR #2353, opened 2022-03-14
Add Resource Page (SOFTWARE-5086)

CannonLock:


djw8605:
If the search field is empty, shouldn't it show all resources?

src/static/js/pages/resources.js
```diff
@@ -0,0 +1,77 @@
...
+            return gridjs.html(node.outerHTML)
+        },
+        html: data => {
+            let node = lod_to_ul(data["Services"]["Service"], "Name")
+            return node
+        }
+    }, {
+        id: "ContactLists",
+        name: "Contact Information",
+        hidden: true, // Danger, there is a reason I don't use this
```
matyasselmeci, line 47: What's the reason?

matyasselmeci (approved):

PR #2355, opened 2022-03-14
Registered GATech ISyE submit host (FD 68302)

rynge:


topology/Georgia Institute of Technology/Georgia Tech/Georgia Tech ISyE.yaml
```diff
@@ -0,0 +1,26 @@
+GroupDescription: Resources in the Georgia Tech PACE cluster, to be used by LIGO.
```
brianhlin, line 1: I think this may describe Memo's group and resources at GT?

brianhlin (commented):
I think we should throw this into a different site than Memo's group. Maybe something like `GT-ISyE`?

topology/Georgia Institute of Technology/Georgia Tech/Georgia Tech ISyE.yaml
rynge: I fixed the group description. Do we want a site per department?

rynge:
I moved the resource out to a separate site. Let me know if it looks good now.

PR #2358, opened 2022-03-17
Update Lancium.yaml

cushingw:
Added squid service entry for Lancium Compute

### DIFF
--- a/topology/Lancium/Lancium/Lancium.yaml
+++ b/topology/Lancium/Lancium/Lancium.yaml
@@ -21,7 +21,9 @@ Resources:
         Primary:
           ID: 91af5365110c5601977381f3ba7d8372a333cf67
           Name: Andrew Grimshaw
-    FQDN: ospool-backfill.lancium.com
+    FQDN: portal.lancium.com
     Services:
       Execution Endpoint:
         Description: Lancium backfill containers
+      Squid:
+        Description: Generic squid service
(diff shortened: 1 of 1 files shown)


brianhlin:
Hi @cushingw, thanks for the PR! For the squid registration, could you register it as a separate resource and leave the backfill registration as is?

See https://github.com/opensciencegrid/topology/blob/85fbd0f0f60a1cc39076db5fecf7fad29a91f157/topology/University%20of%20Wisconsin/CHTC/CHTC.yaml#L294-L310 for an example. Don't worry too much about getting all the fields right, I can drop in suggestions as necessary

DrDaveD (dismissed):

DrDaveD:
@brianhlin it doesn't need to be separate, does it?

brianhlin:
Not strictly but I'd really prefer it if it were: the semantics are fuzzy here with the backfill containers but they're certainly not on the same host as the Squid service

DrDaveD:
I guess they're both behind the same NAT, but not exactly the same resource.  I agree that the resource name doesn't really make sense for the squid.

cushingw:
Is that better?

DrDaveD (dismissed):
The CI check failed because there is no value for the resource "ID:".  A new value can be found by rebasing on the most recent master and  running bin/next_resource_id.  Currently it is 1270.

topology/Lancium/Lancium/Lancium.yaml
```diff
@@ -25,3 +25,24 @@ Resources:
     Services:
       Execution Endpoint:
         Description: Lancium backfill containers
+  Lancium-LMF-Squid:
+    Active: true
+    Description: Squid service located at Lancium LMF datacenter
+    ID:
```
brianhlin, line 31:
```suggestion
    ID: 1270
```

PR #2360, opened 2022-03-18
Software-5093

CannonLock:
Dumps output of bin/get-scitokens-mapfile --regex at a text endpoint


src/webapp/common.py
```diff
@@ -302,6 +302,22 @@ def readfile(path, logger):
...
+
+    escaped_string = re.escape(pattern)
+
+    if sys.version_info < (3, 7):
+        return escaped_string
+
+    unescaped_characters = ['!', '"', '%', "'", ',', '/', ':', ';', '<', '=', '>', '@', "`"]
+    for unescaped_character in unescaped_characters:
+
+        escaped_string = re.sub(unescaped_character, f"\\{unescaped_character}", escaped_string)
```
matyasselmeci, line 316:
```suggestion
    unescaped_characters = '!"%\',/:;<=>@`'
    escaped_string = re.sub(f"([{unescaped_characters}])", r'\\\1', escaped_string)
```
You can put the unescaped characters in a character class `[...]` in a group `(...)` and backreference it `\1` in order to avoid the loop.

CannonLock: Ah nice catch, looks much more succinct!

matyasselmeci (approved):
Looks good.

matyasselmeci:
I can merge this after you've resolved the conflicts.

PR #2367, opened 2022-03-21
Software 5095

CannonLock:
Shows the collaboration table data layouts and the search history that contains previous searches and data indices.

https://user-images.githubusercontent.com/49032265/159372286-7edcdc0a-fb88-4e4c-866d-6ffdb03943c1.mov

Fast demo due to the 10mb limit.


matyasselmeci (approved):
Looks nice!

PR #2373, opened 2022-03-23
Convert to Local Datetime with UTC Offset ( SOFTWARE-3527 )

CannonLock:
Converts the date + UTC time fields into UTC offset + datetime field.

Quick Demo in different timezones:
https://user-images.githubusercontent.com/49032265/159735855-c6d580ff-f0d3-407a-ab16-ec05852d2d38.mov


matyasselmeci (approved):
Nice!!

PR #2375, opened 2022-03-24
Modify T3_KR_KISTI contact person information 

geonmo:
I knew that GGUS tickets are continuously being delivered to those who were previously members. 

So, I create a PR to change this.


brianhlin (approved):
LGTM

PR #2381, opened 2022-03-28
Read-only display for stash-cache/stash-origin authfiles and scitokens.conf (Software-5097)

CannonLock:
Creates a page to display the results of the /origin and /cache endpoints for each resource. 

Search bar is powered by all resource data and can be searched like the resource table.

A json endpoint was added at /resource/resource-files that provides a dictionary of form {\<resourcename>: {<file_name>: <file_string>, **}, ** }.

Singular test added to confirm that this cumulative file endpoint has files consistent to the ones found at the individual endpoints.


CannonLock:
https://user-images.githubusercontent.com/49032265/160487480-53b0ef52-508a-4cea-8aad-5c1d72abff90.mov

matyasselmeci (approved):
I like this. LGTM.

src/webapp/topology.py
```diff
@@ -75,6 +75,74 @@ def __init__(self, name: str, yaml_data: Dict, common_data: CommonData):
         self.fqdn = self.data["FQDN"]
         self.id = self.data["ID"]
 
+    def get_stashcache_files(self, global_data, legacy):
+        """Gets a resources Cache files as a dictionary"""
+
+        import src.stashcache as stashcache
```
matyasselmeci, line 81:
```suggestion
        import stashcache
```

PR #2385, opened 2022-04-04
Install pkill

matyasselmeci:


jthiltges (approved):
LGTM (And sigh at stream changes)

PR #2391, opened 2022-04-06
Removing CE gk04 per Mark

djw8605:
Per ticket FD 69358


brianhlin (commented):
Should we just set `Active: false`?

djw8605:
Yeah, that would probably work as well.

matyasselmeci (approved):

PR #2398, opened 2022-04-12
Remove LIGO DN when not Allowed ( SOFTWARE-5029 )

CannonLock:
Checks for 'ANY' or 'LIGO' adding the ligo dn hashes. Additionally implements tests to confirm these cases and the alternative.


src/tests/test_stashcache.py
```diff
@@ -0,0 +1,46 @@
...
+from app import app, global_data
+import stashcache
+
+@pytest.fixture
+def client():
+    with app.test_client() as client:
+        yield client
+
+
+class TestStachcache:
```
matyasselmeci, line 20:
```suggestion
class TestStashcache:
```

matyasselmeci:
Seems reasonable; can you take a look at the test failure, though?

matyasselmeci (changes requested):
One suggestion; also, can you bring this up to date with master and check if there are test failures?

matyasselmeci (approved):

PR #2399, opened 2022-04-12
Register 2 new CMS resources for UMD:

matyasselmeci:
hepcms-ce2.umd.edu and hepcms-se2.umd.edu

Note that hepcms-ce2 is running both a CE and an XRootD door.

[FD 69353](https://support.opensciencegrid.org/a/tickets/69353)


brianhlin (approved):
LGTM, though I'd imagine that they'll need to provide a `WLCGInformation` section for their CE

PR #2405, opened 2022-04-13
forbid duplicate FQDNs for XRootD services (SOFTWARE-5065)

edquist:
See SOFTWARE-5065 for the deets.

Currently this will catch and block one such bad XRootD service:

```
*** FQDNs must be unique for XRootD services
Duplicate FQDN 'iut2-slate.mwt2.org' used for XRootD services:
 - 'XRootD cache server' on 'MWT2_IU_STASHCACHE' (Indiana University/MWT2 ATLAS IU/IUCachingInfrastructure.yaml)
282 Resource Group files processed.
1 error(s) encountered.
```


matyasselmeci:
Would it be hard to print both if there is a duplicate?

edquist:
>Would it be hard to print both if there is a duplicate?

Sure we can do that, not hard.

Note that only the one is printed above because the other resource(s) with the same FQDN apparently do not have any xrootd services.  Currently i only explicitly list here the resources for the duplicate FQDNs that have xrootd services.

Also relatedly we could re-enable the warnings for all duplicate FQDNs.  Not sure if that's something desirable while we are at this.

matyasselmeci:
> Note that only the one is printed above because the other resource(s) with the same FQDN apparently do not have any xrootd services. Currently i only explicitly list here the resources for the duplicate FQDNs that have xrootd services.

That makes sense but the reason this check is here is because we do lookups by FQDN in stashcache.py, and if there are multiple resources, then we take the first resource, _not_ the first resource with an xrootd service. (We could change that if we really had to but if our options are "complicate our code" vs "fix our data", guess which one I'd pick :-)  ).

> Also relatedly we could re-enable the warnings for all duplicate FQDNs. Not sure if that's something desirable while we are at this.

There are some duplicate FQDNs that we don't know how to fix :-(

edquist:
Arright @matyasselmeci, I now have it list all resources for duplicate FQDNs (where at least one of those resources has an XRootD service), and list all XRootD services for those resources.

How's this look for ya?

```
*** FQDNs must be unique for XRootD services
Duplicate FQDN 'iut2-slate.mwt2.org' used for XRootD services:
 - MWT2_IU_STASHCACHE (Indiana University/MWT2 ATLAS IU/IUCachingInfrastructure.yaml)
   - XRootD cache server
 - MWT2_SLATE_IU (University of Chicago/MWT2 ATLAS UC/MWT2.yaml)
282 Resource Group files processed.
1 error(s) encountered.
```

matyasselmeci:
Looks great!

matyasselmeci (approved):
LGTM

matyasselmeci:
I approved it; since it does break the tests^W^W^Wdo its job, we should fix our data before merging.

brianhlin:
@edquist could you fixup the conflicts and rebase on master? Go ahead and reach out to our friends at MWT2 and ask about the two caches: AFAIK, there aren't any SLATE caches so I'm guessing the latter cache can be removed

edquist:
Conflicts resolved (had to rename test 17->20)

@brianhlin, any idea why there is no log for the failed check?

https://github.com/opensciencegrid/topology/runs/8137972965?check_suite_focus=true

brianhlin:
The old tests ran so long ago that they were removed but it looks like it ran again with your latest commit:

```
*** FQDNs must be unique for XRootD services
Duplicate FQDN 'iut2-slate.mwt2.org' used for XRootD services:
 - MWT2_IU_STASHCACHE (Indiana University/MWT2 ATLAS IU/IUCachingInfrastructure.yaml)
   - XRootD cache server
 - MWT2_SLATE_IU (University of Chicago/MWT2 ATLAS UC/MWT2.yaml)
311 Resource Group files processed.
```

edquist:
@brianhlin, I've deleted the `IUCachingInfrastructure` resource group, since this was the only resource in it.  Lmk if that is not what you wanted.

brianhlin (approved):
LGTM. @edquist can you rebase on `master` and update this PR? Then it's good to be merged form my perspective

PR #2422, opened 2022-04-27
Add /virgo/virgo namespace to LIGO VO

jasoncpatton:
If I understand correctly, this should add the "virgo:/virgo/virgo" VOMS subgroup to `cvmfs-x509-helper`.


jasoncpatton:
Re-doing this per Slack conversation with @brianhlin

PR #2429, opened 2022-05-03
Register FIU-PATH EP

matyasselmeci:
I copied the site address from the "FIU HPC" site.


brianhlin (commented):
@bbockelm do we actually want PATh FIU resources to be accounted as UW hours?

brianhlin (approved):
> do we actually want PATh FIU resources to be accounted as UW hours?

BrianB says these resources are actually owned by UW so this organization makes sense

PR #2444, opened 2022-05-11
Register UNL PATh Facility backfill containers

brianhlin:
Note: no ITB resources as it's not clear how we'll do accounting/ITB jobs in the PATh facility and @djw8605 has not been a fan of EPs running production work being registered as ITB

Also, I differentiated the resource/resource group names for my own sanity


topology/University of Wisconsin/UNL-PATH/UNL-PATH.yaml
```diff
@@ -0,0 +1,29 @@
...
+          Name: Matyas Selmeci
+        Tertiary:
+          ID: 66ee5cfb622a7343dac85dee42815d1f4fbc2d85
+          Name: Brian Bockelman
+      Security Contact:
+        Primary:
+          ID: 46a55ac4815b2b8c00ff283549f413113b45d628
+          Name: Aaron Moate
+    Description: PATh facility EP located at FIU.
+    FQDN: fiu-svc-1.facility.path-cc.io
```
djw8605, line 25: This looks like the FIU execution point?

djw8605 (changes requested):

topology/University of Wisconsin/UNL-PATH/UNL-PATH.yaml
brianhlin: Oops, fixed!

matyasselmeci (approved):

PR #2449, opened 2022-05-12
New Franklin and Marshall Resources

coldbeer101:
New topology resources for Franklin and Marshall

They need IDS on the resoruces still.

Thanks!


topology/Franklin and Mashall College/FACILITY.yaml
```diff
@@ -0,0 +1 @@
+ID:
```
brianhlin, line 1:
```suggestion
ID: 10196
```

topology/Franklin and Mashall College/FANDM-ITS/SITE.yaml
```diff
@@ -0,0 +1,9 @@
+AddressLine1: 637 Colleg Ave
+City: Lancaster 
+Country: United States
+ID:
```
brianhlin, line 4:
```suggestion
ID: 10335
```

brianhlin (approved):
LGTM. We should consider adding you as a contact to some of these Hosted CEs!

PR #2463, opened 2022-05-18
UCSDT2 downtime to restart network switch

jdost321:


jdost321:
I used to just create one manually for all ces and gridtfp but this time for the first time i tried this tool https://topology.opensciencegrid.org/generate_resource_group_downtime

I'd like this to be fixed but I don' tthink i should have to be added as an admin to all those resources just to state a UCSD downtime

matyasselmeci (approved):

matyasselmeci:
We have https://opensciencegrid.atlassian.net/browse/SOFTWARE-3385 to implement site-wide contacts but as you can see it's on the back burner...

PR #2482, opened 2022-06-02
Supplement comanage data with contact db info (SOFTWARE-4734)

edquist:
`merge_yaml_data` used just to add the contact db items to the comanage contact data, as separate contacts; but we actually want to add the missing info to the comanage contacts themselves.

Per SOFTWARE-4734, we want to match on CILogonID, if possible, otherwise match on email address.  Primary or Secondary email from either source can match against either one in the other source.  (Sanity checks of the contact db data confirm for me that this is a safe thing to do.)

I've broken the work for this into a handful of smaller functions in order to help keep my sanity, though there is a growing feeling that all of these, essentially dict manipulation functions could cozily find themselves in a separate module, rather than with the proper cilogon ldap functions.  I'm open to suggestions if anyone has a strong feeling on that.

### DIFF
--- a/src/webapp/cilogon_ldap.py
+++ b/src/webapp/cilogon_ldap.py
@@ -66,10 +66,65 @@ def cilogon_id_map_to_yaml_data(m):
     return data
 
 
+def get_osgid_lookup(yaml_data):
+    osgid_lookup = {}
+    for id_, contact in yaml_data.items():
+        if 'CILogonID' in contact:
+            osgid_lookup[contact['CILogonID']] = contact
+    return osgid_lookup
+
+
+def get_email_lookup(yaml_data):
+    email_lookup = {}
+    for id_, contact in yaml_data.items():
+        ci = contact.get('ContactInformation')
+        if not ci:
+            continue
+        for Email in ('PrimaryEmail', 'SecondaryEmail'):
+            if Email in ci:
+                email_lookup[ci[Email]] = contact
+    return email_lookup
+
+
+def get_sup_contact(contact, osgid_lookup, email_lookup):
+    id_ = contact.get('CILogonID')
+    if id_ in osgid_lookup:
+        return osgid_lookup[id_]
+    ci = contact.get('ContactInformation')
+    if not ci:
+        return None
+    for Email in ('PrimaryEmail', 'SecondaryEmail'):
+        if Email in ci:
+            addr = ci[Email]
+            if addr in email_lookup:
+                return email_lookup[addr]
+    return None
+
+
+def supplement_contact_info(contact, sup_contact):
+    for k,v in sup_contact.items():
+        if k not in contact:
+            contact[k] = sup_contact
+        elif isinstance(k, dict):
+            for k2,v2 in v.items():
+                if k2 not in contact[k]:
+                    contact[k][k2] = v2
+
+
 def merge_yaml_data(yaml_data_main, yaml_data_secondary):
+    # main is comanage (cilogon), secondary is contact db
     yd = dict(yaml_data_main)
+    osgid_lookup = get_osgid_lookup(yaml_data_secondary)
+    email_lookup = get_email_lookup(yaml_data_secondary)
+
+    for contact in yd.values():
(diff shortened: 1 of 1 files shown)


matyasselmeci (approved):

src/webapp/cilogon_ldap.py
```diff
@@ -66,10 +66,65 @@ def cilogon_id_map_to_yaml_data(m):
...
+
+
+def supplement_contact_info(contact, sup_contact):
+    for k,v in sup_contact.items():
+        if k not in contact:
+            contact[k] = v
+        elif isinstance(k, dict):
+            for k2,v2 in v.items():
+                if k2 not in contact[k]:
+                    contact[k][k2] = v2
```
matyasselmeci, line 111:
```suggestion
        elif isinstance(v, dict):
            contact[k].setdefault(v)
```

src/webapp/cilogon_ldap.py
```diff
@@ -66,10 +66,65 @@ def cilogon_id_map_to_yaml_data(m):
     return data
 
 
+def get_osgid_lookup(yaml_data):
+    osgid_lookup = {}
+    for id_, contact in yaml_data.items():
+        if 'CILogonID' in contact:
+            osgid_lookup[contact['CILogonID']] = contact
```
matyasselmeci, line 73:
```suggestion
    for contact in yaml_data.values():
        if 'CILogonID' in contact:
            osgid_lookup[contact['CILogonID']] = contact
```

matyasselmeci (approved):
One bug (`isinstance(k, dict)` instead of `isinstance(v, dict)` -- I sure hope the key isn't a dict) and a few tweaks; otherwise pre-approving.

src/webapp/cilogon_ldap.py
edquist: Yeah... The first time through i wasn't sure if i'd use the id - after refactoring to its current form clearly it is unused.  I noticed this after making the PR, and thought of making the same change, but was too lazy to change it yesterday :)

src/webapp/cilogon_ldap.py
edquist: >One bug (`isinstance(k, dict)` instead of `isinstance(v, dict)` -- I sure hope the key isn't a dict)

Hey, good catch!

I will fix the `elif isinstance` line to use `v`.

I don't think the `setdefault` replacement you're suggesting is the logic i had in mind though.

This is the `elif` block, which means `k` (most likely `'ContactInformation'`) is in `contact` (ie, the one from comanage).

My intention with the next level of iteration over `v` is to fill in
```
    contact[k][k2] = sup_contact[k][k2]
```
where `contact[k]` exists, but `k2` is only in `sup_contact[k]`, not `contact[k]`.

That is, fill in any `ContactInformation` details from contact db that are not in comanage data, without replacing the whole `ContactInformation` sub-dict.  I don't think `setdefault` will do that as suggested.

edquist: @matyasselmeci, i've tweaked `supplement_contact_info` a bit more, hopefully to improve the clarity of what i am trying to do (as described in my comment above).  Going to kick this back to you to re-review, really just this one function, as i don't think the `setdefault` suggestion works as written.

FWIW, i realize it would also work to reduce the body of the `elif` block to a recursive call (`supplement_contact_info(contact[k], sup_contact[k])`), though at the moment i don't think we ever need more than one level deep.  (Really, it's just for the `ContactInformation` key.)

edquist: @brianhlin, roping you in since Mat is out today.  Mat pre-approved but one of his suggestions didn't look right to me.  The only function in question (re his suggested change) was `supplement_contact_info`, which I've made some tweaks to.  So that i don't let undue power go to my head, would you mind giving a quick once-over of this function, and see if it looks reasonable enough to you?

https://github.com/opensciencegrid/topology/pull/2482/files#diff-94a1e62c71318b0903ff568fc7252f8113799d34c2c36698204b821ada90469aR104-R110

Thanks!

brianhlin (approved):
Looks reasonable to me. Go ahead with testing!

PR #2488, opened 2022-06-07
allow ContactLists ID to match CILogonID, too (SOFTWARE-5216)

edquist:
this fixes the issue with a github username mapping to multiple
contacts (one from contact db and the other from comanage), and
also allows specifying a CILogonID (aka OSGID) in topology data
ContactLists for resources.

The set arithmetic is saying that there must be some overlap between
the contact IDs associated with the github username and the IDs
in the ContactLists registered for a given resource.


matyasselmeci (approved):
LGTM!

PR #2494, opened 2022-06-08
Create JLAB-TEST.yaml

kurtjstrosahl-jlab:
adding a test project for jlab that users looking to try out the OSG can use to get started.

### DIFF
--- a/projects/JLAB-TEST.yaml
+++ b/projects/JLAB-TEST.yaml
@@ -1,7 +1,6 @@
 Department: Physics
 Description: Jefferson Lab's test experiment
 FieldOfScience: Nuclear Physics
-ID: '851'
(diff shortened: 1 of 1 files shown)


brianhlin:
Marking as `do-not-merge` just so we can make sure that if this project runs in the OSPool that it matches relevant naming conventions

@kurtjstrosahl-jlab @jasoncpatton would users with access to this project be submitting solely in the EIC/CLAS12 pools or can the jobs flock out to the OSPool?

kurtjstrosahl-jlab:
These would be jobs for the OSPool, not dedicated resources. @ppaschos and I discussed it yesterday.

brianhlin (approved):
LGTM and we got the thumbs up on the name from the OSPool facilitators

PR #2495, opened 2022-06-08
allow case differences in contact names (SOFTWARE-5216)

edquist:
this is primarily to allow the all-caps names that are often found in comanage contact names


brianhlin (approved):
LGTM

PR #2498, opened 2022-06-08
Add WISC-PATH site and OSPool backfill containers

brianhlin:


matyasselmeci:
That one's already registered at `CHTC/PATh_facility.yaml`

brianhlin:
Ah ok, I was looking for `WISC-PATH-EP`, which is what we've reported to the GRACC. I say we get rid of the CHTC resource group in favor of this one to match the other PATH sites under UW unless @bbockelm wants these hours to be credited to CHTC

matyasselmeci:
I'm fine either way. What do you say, @bbockelm?

matyasselmeci:
@bbockelm what's your opinion on having the CHTC Site with a PATh_facility ResourceGroup (as we do currently) versus having a WISC-PATH Site with a WISC-PATH ResourceGroup?

matyasselmeci (approved):

matyasselmeci:
Merging this since, as you said, this name has already reported hours. I'll get rid of the other one in a different PR.

PR #2500, opened 2022-06-09
Replace CHTC contact IDs with CILogonIDs

edquist:
If the case-insensitive changes from #2495 / v1.18.2 work then the checks should pass this time...


edquist:
Good, checks are passing again now, so the Carl Edquist / CARL EDQUIST difference is accepted by the tests.

edquist:
Per Brian B i'm doing this for CHTC contacts also.  Because we don't have a CHTC Staff flag in the contact db, i am limiting this to the contacts that are registered with resources in the CHTC site.

edquist:
@matyasselmeci @brianhlin ping a ling ding when you getta chance

matyasselmeci (approved):
Well sed!

edquist:
>Well sed!

"sed, sed, have sed", or is simple past "sedded" ?

PR #2501, opened 2022-06-09
Add leaflet map

CannonLock:


CannonLock:
Video Demo: https://drive.google.com/file/d/1qCG44DC7INQ4brf3IXj43WSR7rB74zBq/view?usp=sharing

Same map as the previous with the below differences:

- OSPool View
- Uses Leaflet which is free vs the Google map that was entering pay territory
- Compatible with previous url parameters
- Changed the icons to svgs because I like the infinite resolution

PR #2505, opened 2022-06-09
Upgrade MarkupSafe, but cap it < 2.1 due to an incompatibility:

matyasselmeci:
https://stackoverflow.com/questions/72191560/importerror-cannot-import-name-soft-unicode-from-markupsafe

Due to a typo in the requirements files (missing "," between "1.1.1" and "<1.2"), we were running the latest version anyway, so the upgrade is safe.


brianhlin:
Unit tests, integration tests, and now typo-based tests. LGTM!

PR #2509, opened 2022-06-13
University of Alabama Tuscaloosa is a 2020 CC* awardee

brianhlin:


edquist (approved):
If @brianhlin says so, that's good enough for me!

PR #2513, opened 2022-06-14
Add unique resource ID check (SOFTWARE-5068)

edquist:


src/tests/verify_resources.py
```diff
@@ -337,19 +339,41 @@ def test_8_res_ids(rgs, rgfns):
...
+    ridres = autodict()
+    gidrgs = autodict()
 
     for rg,rgfn in zip(rgs,rgfns):
         if not isinstance(rg.get('GroupID'), int):
             print_emsg_once('ResGrpID')
             print("Resource Group missing numeric GroupID: '%s'" % rgfn)
             errors += 1
+        else:
+            gidrgs[rg['ID']] += [rgfn]
```
matyasselmeci, line 351:
```suggestion
            gidrgs[rg['GroupID']] += [rgfn]
```
Because consistency is not our strong suit.

edquist: Good catch!

PR #2519, opened 2022-06-17
Remove duplicate contact entries in /contacts page (SOFTWARE-5214)

edquist:
As in the commit message:

We add a `without_duplicates()` method to ContactsData to return a copy of the ContactsData object, with contact db entries removed, if all field values are identical (modulo case) to a corresponding comanage contact entry.

(this is meant to be used _after_ the contact db and comanage results are merged, of course)

The reason this is done in a copy, rather than removing the duplicates from the original, is so that lookups by contact id will still work.

I am open to moving the new helper functions, `_recursive_lower` and/or `_id_is_duplicate` somewhere else, if you think another module would be a better home for them.  In any case i felt like the implementation was a lot cleaner breaking it up into multiple functions this way.


matyasselmeci (approved):
LGTM

edquist:
:smiling_face_with_tear:

PR #2530, opened 2022-06-24
Move CHTC-submit6 to the PATh_facility ResourceGroup since it was repurposed as a PATh Facility AP

matyasselmeci:


brianhlin (approved):
LGTM

PR #2533, opened 2022-06-27
Update BU_ATLAS_Tier2.yaml

SaulYoussef:
Adding OSG 3.6, ipv6 gatekeeper


SaulYoussef:
Thanks so much for catching this, matyasselmeci.  Cleaning up accounting after the fact is definitely to be avoided.  I am about to commit a change as you suggested, but would appreciate it if you could have a look and approve if it looks OK.  I created a new "Resource" called "NET2v6" with ID 1308 as you suggest.  I'll keep "NET2" "Active:true" for the moment since it's still up.  - Saul

matyasselmeci:
OK, I merged your other PR so I will close this one.

PR #2539, opened 2022-06-29
Update BU_ATLAS_Tier2.yaml

SaulYoussef:
As suggested by Matyas Selmeci, I have copied the NET2 resource into "NET2v6", changing the FQDN and ID, but keeping everything else the same.  This is is meant to reflect adding a new ipv6 gatekeeper.  - Saul


matyasselmeci (approved):

matyasselmeci:
Looks good. Thanks!

PR #2544, opened 2022-07-05
move ASGC to Academia Sinica facility

edquist:
https://support.opensciencegrid.org/a/tickets/68235

Not clear if we want to add ASGC-IGWN-CE resource group here right away, or if we'll let the admin add that.


brianhlin:
Let's do it for them since it's been such a long-lived ticket. I'd also like to override their suggestion:

RG: ASGC-IGWN
Resource: ASGC-IGWN-CE

They may need to run other services, like a squid or an OSDF cache, so there's no sense in having an RG that's specific to a service type

topology/Academia Sinica/ASGC/ASGC-IGWN.yaml
```diff
@@ -0,0 +1,34 @@
+Production: true
+SupportCenter: Self Supported
+GroupDescription: This is an htcondor-ce cluster for the IGWN service on ASGC.
+GroupID: 1300
+
+Resources:
+  ASGC-IGWN-CE:
+    Active: false
+    Description: This is an htcondor-ce for ASGC-IGWN.
+    ID: 1309
```
brianhlin, line 10:
```suggestion
    ID: 1310
```

topology/Academia Sinica/ASGC/ASGC-IGWN.yaml
```diff
@@ -0,0 +1,34 @@
+Production: true
+SupportCenter: Self Supported
+GroupDescription: This is an htcondor-ce cluster for the IGWN service on ASGC.
```
brianhlin, line 3:
```suggestion
GroupDescription: These are resources for IGWN running at the Academia Sinica Grid Computing Center
```

This is phrased oddly. The HTCondor-CE dumps pilots into their local batch scheduler so it'd be more correct to say that the CE is serving the cluster. However, it's a group so they could feasibly run multiple services so highlighting the CE also isn't quite right.

IGWN is also the collaboration of LIGO/VIRGO so they're not really a service and ASGC is the computing center

brianhlin (approved):
LGTM with some suggestions so pre-approving. Feel free to rebase or however you want to bring this branch up-to-date

PR #2546, opened 2022-07-07
rename UW-IT facility

coldbeer101:
Renaming Facility Directory to show different GRACC name on customers request


brianhlin (approved):
LGTM

PR #2551, opened 2022-07-11
Add Expanse-PATH-EP resource

matyasselmeci:


brianhlin (changes requested):
This should go into https://github.com/opensciencegrid/topology/blob/master/topology/University%20of%20California%20San%20Diego/San%20Diego%20Supercomputer%20Center/SDSC-Expanse.yaml

matyasselmeci:
Moved. Why do we have the SDSC-Expanse site if it's empty and we're not putting stuff in it?

PR #2557, opened 2022-07-14
Validate AllowedVOs in xrootd origin/cache resources (SOFTWARE-4167)

edquist:
test_18_osdf_data_warnings is TODO

PR #2560, opened 2022-07-15
Add missing AllowedVOs list to the Georgia_Tech_PACE-GridFTP resource…

matyasselmeci:
… (which is both a gridftp resource and a cache)

Spotted by Carl's code in #2557


brianhlin (approved):
LGTM!

PR #2567, opened 2022-07-20
Register the HTCSS dev OSPool AP (INF-61)

brianhlin:
We're dropping this into Tiger so you guys get to be secondary/tertiary admin contacts :)


matyasselmeci (approved):
I have no objections to being responsible for the security of... whatever this is... :-P

PR #2578, opened 2022-07-26
Add downtime for Crane-CE1 due to cluster filesystem outage

jthiltges:


jthiltges:
Did I delete the branch too soon? The helpful horse assured me it was already merged, and it indeed seems merged. @edquist

edquist:
Hey @jthiltges, thanks for reporting this.  Not quite sure what osg-bot is fussing about but your PR was definitely merged; so i think you are all set and i don't think it was wrong that you deleted your branch.

From the looks of it, maybe something triggered a duplicate check suite hook (also successful), which caused a second merge attempt (after the PR was already merged), and so github complained that "Pull Request is not mergeable" ...

If we start seeing repeats of this it might be something for us to look into.

PR #2583, opened 2022-08-01
Ignore check_runs which do not validate data (SOFTWARE-5272)

edquist:
otherwise osg-bot gets jumpy and tries to merge a suitable PR for every check_suite that succeeds - and now there are two that run for each PR (data and src).  We only want it to care about the one for the data checks.

...

A bit more complicated than expected because the check_suite name does not appear in the payload for the check_suite hook, so we need to follow an api url to grab the name of each check run for each check suite.  Ah well.

### DIFF
--- a/.github/workflows/validate-data.yml
+++ b/.github/workflows/validate-data.yml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 
 jobs:
   validate-data:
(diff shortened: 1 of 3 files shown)


matyasselmeci (approved):
LGTM. Can you add a comment to the "Validate Topology Data" action to mention that the name must match what's in the webhook?

edquist:
Sure thing; added.

PR #2590, opened 2022-08-01
Build images from 4-number versions with ".itb" in them

matyasselmeci:
I would like to differentiate between images built from _master_ vs images built from _itb_.


edquist (approved):
what, you think just because i suggested it i will approve it too?  ;)

PR #2596, opened 2022-08-02
Register the origins I'll be using to test the stash-origin Helm chart

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-5258


brianhlin:
Do you need corresponding VO entries, too?

matyasselmeci:
I will, but I'll make a separate pull request for those depending on whether we merge #2558 first.

PR #2608, opened 2022-08-04
Create TG-PHY210083.yaml

showmic09:


matyasselmeci:
This looks like an XSEDE project name -- @rynge, shouldn't the sponsor be `OSG-XSEDE` instead of `OSG Connect`?

rynge:
I think OSG Connect is correct. That is where we are putting all the XSEDE allocations now.

matyasselmeci (approved):

PR #2610, opened 2022-08-05
rename CHTC-EHT-EP -> CHTC-GPU-EP (SOFTWARE-5283)

edquist:
@bbockelm, updating per your comment [here](https://github.com/opensciencegrid/tiger-osg-config/pull/492#pullrequestreview-1063487723)

>2. In the topology registration, let's call it something non-EHT related. "Tiger-GPU-EP" or "CHTC-GPU-EP" would be good starters.

Is it correct that i am also updating the FQDN here?

### DIFF
--- a/topology/University of Wisconsin/CHTC/CHTC_OSPOOL.yaml
+++ b/topology/University of Wisconsin/CHTC/CHTC_OSPOOL.yaml
@@ -64,7 +64,7 @@ Resources:
         Description: OSPool Access Point
     Tags:
       - OSPool
-  CHTC-EHT-EP:
+  CHTC-GPU-EP:
     Active: true
     ContactLists:
       Administrative Contact:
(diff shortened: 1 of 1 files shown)


edquist:
Hey @matyasselmeci, Brian x2 are out today.  Could you give this a quick review?  I am renaming the new resource per Brian B (see description), and i am also renaming the (fake) FQDN.

Only, off hand i do not remember if i will need to re-do the sealed secret in the tiger config as a result of the FQDN change?

matyasselmeci:
I don't recall off-hand. A SealedSecret creates a Secret from it so do `kubectl get secrets` to get the secret name, and do `kubectl describe secret <secret name>` (you might have to pipe the contents through `base64 -d`). If you see the hostname in the secret, you'll need to make a new secret. If you don't, you won't.

If this is being done on tiger, you'll have to change the hostname on tiger as well.

matyasselmeci (approved):

PR #2614, opened 2022-08-09
Cache LIGO DN list like we do with CILogon data

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-5276 -- this puts a copy of the DN list we obtain by querying LIGO LDAP into a CachedData object in GlobalData.

I renamed cilogon_ldap.py to ldap_data.py and added the LIGO LDAP stuff to it. Aside from a few renamings so that both the CILogon and the LIGO code use the same names, I didn't have to change much.


matyasselmeci:
There we go, tests fixed. This is ready for review.

edquist:
Cache it like it's hot

src/webapp/ldap_data.py
```diff
@@ -8,24 +8,24 @@
...
+_cilogon_basedn   = "o=OSG,o=CO,dc=cilogon,dc=org"
 
 
-def get_cilogon_ldap_id_map(ldap_url, username, ldappass):
+def get_cilogon_ldap_id_map(ldap_url, ldap_user, ldap_pass):
     """ return dict of cilogon ldap data for each CILogonID, with the
         structure: {CILogonID: { "dn": dn, "data": data }, ...} """
     server = ldap3.Server(ldap_url)
-    conn = ldap3.Connection(server, username, ldappass)
+    conn = ldap3.Connection(server, ldap_user, ldap_pass)
```
edquist: > Some naming tweaks for consistency

I think my naming convention consistency consistently consists of being consistent with my mood at the time of writing ;)

src/webapp/ldap_data.py
```diff
@@ -158,16 +158,17 @@ def get_ligo_ldap_dns(ldap_url: str, ldapuser: str, ldap_pass: str) -> List[str]
...
+            try:
+                conn.search(base_branch.format(group=group),
+                            queries[group],
+                            search_scope='SUBTREE',
+                            attributes=['gridX509subject'])
+                results += [dn for e in conn.entries for dn in e.gridX509subject]
+            except ldap3.core.exceptions.LDAPException:
+                log.exception("Failed to query LIGO LDAP for %s DNs", group)
+    finally:
+        conn.unbind()
```
edquist, line 172: i squinted at this for a while and then gave up and did a 

```git show --color-words 46e1faff```

and suddenly it all made sense.

Too bad github was not very good at highlighting the actual change here (inserting `try:` and `finally:`)

File: src/webapp/ldap_data.py [hunk 3]
```diff
@@ -165,7 +165,7 @@ def get_ligo_ldap_dns(ldap_url: str, ldapuser: str, ldap_pass: str) -> List[str]
                             queries[group],
                             search_scope='SUBTREE',
                             attributes=['gridX509subject'])
-                results += [dn for e in conn.entries for dn in e.gridX509subject]
+                results.extend(dn for e in conn.entries for dn in e.gridX509subject)
```
edquist, line 168: Do you really think this is any more efficient though?

src/webapp/models.py
```diff
@@ -204,6 +205,28 @@ def get_contacts_data(self) -> ContactsData:
...
         return self.merged_contacts_data.data
 
+    def get_ligo_dns(self) -> List[str]:
+        """
+        Get list of DNs of authorized LIGO users from their LDAP
+        """
+        if not (self.ligo_ldap_url and self.ligo_ldap_user and
+                self.ligo_ldap_passfile):
+            log.debug("LIGO_LDAP_{URL|USER|PASSFILE} not specified; "
+                      "getting empty list")
```
edquist, line 217: i would say `"returning empty list"`, but i see there is precedent elsewhere for the "getting" wording

src/stashcache.py
```diff
@@ -3,7 +3,7 @@
 
 from webapp.common import is_null, readfile, PreJSON, XROOTD_CACHE_SERVER, XROOTD_ORIGIN_SERVER
 from webapp.exceptions import DataError, ResourceNotRegistered, ResourceMissingService
-from webapp.ldap_data import get_ligo_ldap_dns
+from webapp.ldap_data import get_ligo_ldap_dn_list
```
edquist, line 6: >because `dns` looks like it's talking about DNS, not DNs

How about DsN? ;)

src/config-ci.py
```diff
@@ -10,8 +10,10 @@
 NO_GIT = True
 
 # test_verify_schema.sh might have checked this out
-if os.path.isdir("/tmp/contact"):
+if os.path.isdir("/tmp/contact") and os.path.exists("/tmp/contact/contacts.yaml"):
```
edquist, line 13:
```suggestion
if os.path.exists("/tmp/contact/contacts.yaml"):
```

seems like the `os.path.isdir("/tmp/contact")` part is redundant now that you're checking for the file itself?

edquist (approved):
LGTM.  Thank you for breaking this up into relatively small commits - made reviewing it much easier.

edquist:
> LGTM

Er, but do see comments before merging pls.

src/webapp/ldap_data.py
matyasselmeci: `+=` makes a copy whereas `.extend()` modifies in-place, but yeah in this case the difference is negligible. Now if the _appending_ were done in a loop, that would be a different story.

src/webapp/ldap_data.py
matyasselmeci: That's consistent with what I know about you ;)

src/webapp/models.py
matyasselmeci: These are debug messages so if we think "returning" is better then we can change it in multiple places -- but that's a separate PR :)

PR #2619, opened 2022-08-09
[CIT_CMS_T2] Remove absent entries and associated downtime entries

juztas:


matyasselmeci (approved):

matyasselmeci:
I've seen no GRACC records from these hosts in the last 2 years so it's OK to delete these.

PR #2620, opened 2022-08-10
Update DataFederations/StashCache data to use the new schema added in SOFTWARE-5179

matyasselmeci:
This _does_ result in some output differences:

- Authfiles for caches not in the AllowedCaches list of a Namespace no longer allow access to data from that Namespace
  (instead of just not being able to obtain it from the origin)

- Authfiles for authenticated origin instances no longer allow access to public data they may have


brianhlin (approved):
LGTM!

PR #2628, opened 2022-08-16
Use James Hongs' COManage ID (FD#70290)

brianhlin:
If someone has a COManage ID, we should always prefer it since we have
services (e.g., OS Registry) that do lookups against COManage

@edquist I thought that this would "just work" with your merging COManage and Topology IDs work


matyasselmeci (approved):

edquist:
>edquist I thought that this would "just work" with your merging COManage and Topology IDs work

yes that's my impression also.  did i mess something up here?

PR #2629, opened 2022-08-16
Fix duplicate IDs in GP-ARGO resources

matyasselmeci:


brianhlin (approved):
Sigh, we don't check for this?

matyasselmeci:
We have #2513 to check for this, it's just that we haven't merged it yet because we have too many duplicate IDs. :/

PR #2635, opened 2022-08-17
Add /osgconnect/protected namespace

matyasselmeci:


matyasselmeci:
@djw8605 / @bbockelm , do I have the base path for the issuer correct?  Are the scitoken scopes going to be relative to /osgconnect/protected?

djw8605:
I think no base path. Scopes will have /osgconnect/protected in them. Very similar to the public directory. 

-Derek

matyasselmeci:
OK, done. Can you review?

djw8605 (approved):
LGTM.  At some point, we should include a "restricted path", but I think this is good.

I don't think this will cause issues because it's the same issuer as another, but we will see.

matyasselmeci:
OK, added a 'restricted path'.

PR #2641, opened 2022-08-20
Create IFAHURP-PRP.yaml

edubergeek:
Register prp01.ifa.hawaii.edu as an OSG resource


edubergeek:
Still working on the local version so please hold off on merging

PR #2652, opened 2022-08-30
Update SITE.yaml

rob-c:
Changing ECDF site Longitude to point to Edinburgh city
(matching gocdb entry https://goc.egi.eu/portal/index.php?Page_Type=Site&id=105)


bbockelm (approved):
Oof.  I wonder how many times we flip the +/- on the longitude.

Change LGTM.

PR #2654, opened 2022-09-06
Create GPN.yaml

showmic09:

PR #2656, opened 2022-09-08
Added new project: GPN

rynge:


rynge:
GPN is a special case, and facilitation team has decided it does not need to follow the insitution/lastname convention.

brianhlin:
@rynge gotcha. @showmic09 already has a PR in #2654 where the only difference is Showmic's PR has:

```
Department: Research Computing
```

Which one's the right one?

PR #2658, opened 2022-09-09
normalize emails to lowercase for merging contacts

edquist:
As noticed while inspecting ldap results for SOFTWARE-5313, comanage apparently sometimes stores email addresses in all-caps. That ought to be OK, since cilogon_id_map_to_yaml_data() stores them in lowercase.  But the email_lookup table generated from the contacts db does not do this, and sometimes emails are actually stored in mixed-case.  So we need to build the email_lookup table based on the normalized lowercase in order to merge on email addresses properly.

This issue (bug) may have resulted in some contacts failing to merge automatically on email address.


src/webapp/ldap_data.py
```diff
@@ -91,7 +91,7 @@ def get_email_lookup(yaml_data):
             continue
         for Email in ('PrimaryEmail', 'SecondaryEmail'):
             if Email in ci:
```
matyasselmeci, line 93: Don't you also want this too?
```suggestion
            if Email.lower() in ci:
```

edquist: No - `Email` here loops over the two keys `'PrimaryEmail'` and `'SecondaryEmail'`, so `.lower()` there would look for items under `'primaryemail'` and `'secondaryemail'`, which is not what we want.

matyasselmeci (approved):

PR #2659, opened 2022-09-12
Added new project: UPRM_Ramos

rynge:


matyasselmeci:
We have (CMS) projects with a "University of Puerto Rico" organization -- should we use that? Alternatively we have some UPRM resources in Topology where it's spelled "University of Puerto Rico - Mayaguez"

rynge:
I think we should include the campus - that is what we doing for other universities. For example, we have UWMadision, not just UW. I vaguely remember somebody wanting comma instead of dash (see UCSD entries for example), but I have no strong preference. If it keep things aligned, I can change it to a dash.

matyasselmeci:
OK let's just do the dash, then.

PR #2669, opened 2022-09-26
Update MWT2.yaml to remove MWT2_CE_UIUC

luehring:
The CE at UIUC has been removed from service so I removes it from the topology.


matyasselmeci:
Discussed this in Slack -- the token testing system shouldn't be looking at inactive resources.

Resources can only be deleted once they no longer have GRACC records less than two years old.

PR #2682, opened 2022-09-27
only have validate-caches cron run once daily (SOFTWARE-3470)

edquist:
per @DrDaveD's suggestion in SOFTWARE-3470


brianhlin (approved):
LGTM

PR #2687, opened 2022-10-04
reporting downtime 2022-09-04

iraklic:


iraklic:
> @iraklic it looks like you modified a pre-existing downtime. I imagine that you only wanted to append?

I was under impression that this file only holds future/current downtimes, because I have never seen it keeping the previous reports, even the one I modified only had one downtime report. Should I just append to it in the future and it will keep getting ever longer with the new downtime? Or did you mean something else?

brianhlin:
Hi @iraklic, we do indeed keep track of old downtimes so you can go ahead and just append downtimes in the future. In the downstream data that we make available, we do filter on current + future downtimes but also provide views with old downtimes: https://topology.opensciencegrid.org/

iraklic:
Thanks. I will keep this in mind for the future reportings.

PR #2694, opened 2022-10-12
Add rest_data for github ids from comanage (SOFTWARE-5313)

edquist:
Includes most of the config plumbing.

get_osgid_github_map creates an osgid -> github identifier map

Need to make corresponding tiger changes to provide `COMANAGE_REST_PASSFILE` for osg-topology containers


src/webapp/rest_data.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def get_co_group(gid):
+    grouplist = call_api("co_groups/%s.json" % gid) | get_datalist("CoGroups")
+    if not grouplist:
+        raise RuntimeError("No such CO Group Id: %s" % gid)
+    return grouplist[0]
+
+
+# @rorable
```
matyasselmeci, line 103: I may like this thing but the comment
```
# x | foo -> foo(x)
```
and the way it's actually used
```
get_osg_co_groups() | get_datalist('CoGroups')
```
do not look similar.
What does this _actually_ do?

edquist: Right so as noted in the first commit message, this code was pulled in verbatim from group_fixup.py from the osg-comanage-project-usermap repo.

Note the "ror" in rorable refers to the reverse-or operator, `__ror__`, which is called in an `a | b` expression against `b` if `a.__or__` is not defined.

> What does this _actually_ do?

The comment:
```
# x | foo -> foo(x)
```

Attempts tersely to explain that if `foo` is decorated with `rorable`, it will allow you to write `x | foo` to call & return `foo(x)`.

Thus in
```
get_osg_co_groups() | get_datalist('CoGroups')
```

`x` is `get_osg_co_groups()` and `foo` is `get_datalist('CoGroups')`

so it actually does:
```
get_datalist('CoGroups')(get_osg_co_groups())
```

Which to me was annoyingly harder to read, which was my original motivation for concocting rorable...

matyasselmeci: so get_datalist() returns a function and not a, you know, list?

edquist: ...

Now in case you're askind about this part:

```
    return rorable(get)
```

...

Normally decorating a function definition is _shorthand_ for wrapping it, but in this particular instance, it was shorter (one fewer line) just to wrap the function directly.

What I'm saying is that where I have:

```
def get_datalist(listname):
    def get(data):
        return data[listname] if data else []
    return rorable(get)
```

I _could_ have equivalently written

```
def get_datalist(listname):
    @rorable
    def get(data):
        return data[listname] if data else []
    return get
```

but that is one line longer.

edquist: > so get_datalist() returns a function and not a, you know, list?

Yeah, as you can see from the definition:

```
def get_datalist(listname):
    def get(data):
        return data[listname] if data else []
    return rorable(get)
```

But again it's intended to be used like this:

```
get_osg_co_groups() | get_datalist('CoGroups')
```

And that expression does return a list (the datalist named 'CoGroups').

edquist: So, maybe I have explained well enough already what I am trying to do.  But i will offer yet another angle to explain my thinking.

See what'd I'd like to do for all the places were i use get_datalist is conceptually this:

```
get_osg_co_groups().get_datalist('CoGroups')
```

That is, to get the 'CoGroups' datalist (or whatever the other datalist names may be in the other cases) from the comanage query results.  But the results are just json, converted to a dict.  So I can't use the dot operator.  But using a pipe reads kind of like the shell and in that way feels a bit familiar to express the right hand side of the pipe is being used as a filter of sorts (in the shell sense).

edquist: >I may like this thing

Good i hope it inspires you to have lots of fun in your python exploits to come :)

edquist:
Alright get excited i think all the plumbing is in

src/webapp/rest_data.py
matyasselmeci: I shouldn't need to see the definition to find out what its true return type is. If you were to name it "get_datalist_getter" or something, that might make it less unclear.

brianhlin (commented):
We can lose all the changes in `default_config.py` and `src/webapp/models.py`, right? I'd also like a comment added to the top of `src/webapp/rest_data.py` that at least points to SOFTWARE-5313 and ideally explains that we're keeping the code around in case we need the COManage REST API in the future

src/webapp/rest_data.py
edquist: Alright i dropped the `rorable` class, and made get_datalist() take two parameters instead.  (The listname and the data.)

I guess rorable was too cute for the crowd.  Maybe it was just ahead of it's time.

edquist:
>We can lose all the changes in default_config.py and src/webapp/models.py, right?

Sure; I guess models.py contained what could be example code for how to use it in the future (all the `COMANAGE_REST_*` config vars), but we aren't using it anymore, after finding the much-better ldap solution.

Anyway, changes from default_config.py and models.py dropped.

Alternatively, we could put all the changes from default_config.py, models.py, and rest_data.py into a branch.

edquist:
>I'd also like a comment added to the top of src/webapp/rest_data.py that at least points to SOFTWARE-5313 and ideally explains that we're keeping the code around in case we need the COManage REST API in the future

Done.

brianhlin (approved):
LGTM!!! Feel free to keep the configs that you had before in a separate branch in this fork and mention it in the Jira ticket

PR #2698, opened 2022-10-17
Update AmsterdamPRPCachingInfrastructure.yaml - FQDNAliases fiona.uvalight.net

biozit:


matyasselmeci (dismissed):

matyasselmeci:
This change is harmless but FQDNAliases isn't used for much, at least not for caches. What problem are you trying to solve with this?

biozit:
This one:
https://github.com/opensciencegrid/topology/actions/runs/3231194189/jobs/5290456821#step:4:1

matyasselmeci:
Ah, OK. For that, you use endpoint_override and auth_endpoint_override like so: https://github.com/opensciencegrid/topology/tree/master/src#cache-endpoints

biozit:
Please, could you check it?

matyasselmeci:
Yep, looks good.

PR #2724, opened 2022-10-31
Add facility endpoint

CannonLock:
Added a linear search with cached result for a entity being in cc*\
Started a test class for API endpoint content and some basic tests to start


src/webapp/topology.py
```diff
@@ -217,6 +246,15 @@ def get_tree(self, authorized=False, filters: Filters = None) -> Optional[Ordere
 
         return new_res
 
+    @property
+    def is_ccstar(self):
+        """Check if this site is tagged as a CC* Site"""
+        try:
+            return "CC*" in self.data["Tags"]
+        except:
+            return False
```
matyasselmeci, line 255:
```suggestion
        return "CC*" in self.data.get("Tags", [])
```
Let's avoid an overly broad except.

CannonLock: @matyasselmeci Oh awesome, this a lot better!

matyasselmeci (approved):

PR #2751, opened 2022-11-10
LIGO: add new cilogon/igwn user token issuers

### LINKED ISSUE #2747: Add new IGWN token issuer for LIGO VO
The LIGO VO topology includes the https://scitokens.org/ligo token issuer. AFAIK we are transitioning to https://scitokens.org/igwn, which will include some new scopes etc for internal use.

Can we update the LIGO VO topology file to reference that issuer? I am listed as a VO Manager, so should I be the one to propose a Pull Request that implements that?

cc @josh-willis @rdt12 @jjg-123

duncanmmacleod:
This PR closes #2747 by adding references to the new cilogon.org/igwn token issuers for the LIGO VO. These are used for user tokens, not pilot tokens.

### DIFF
--- a/virtual-organizations/LIGO.yaml
+++ b/virtual-organizations/LIGO.yaml
@@ -79,6 +79,12 @@ DataFederations:
           - FQAN: /virgo
           - FQAN: /virgo/virgo
           - SciTokens:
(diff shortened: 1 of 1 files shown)


brianhlin (approved):
LGTM

PR #2759, opened 2022-11-10
Update cache_config_whitelist.txt

biozit:


src/tests/cache_config_whitelist.txt
```diff
@@ -2,3 +2,4 @@
 xrootd-local.unl.edu:1094
 stash.farm.particle.cz:8000
 stashcache.jinr.ru:8000
+stashcache.gwave.ics.psu.edu:8443
```
edquist, line 5: This whitelist is for caches configured in CVMFS, that are not in topology.

But my read of SOFTWARE-3470 is that `stashcache.gwave.ics.psu.edu:8443` is in the list "Caches in Topology, missing from CVMFS config ligo.osgstorage.org.conf".

So, I do not think it will solve the problem to add this item to the whitelist.

We need to do one of these:
1. add it to CVMFS config
2. remove it from Topology
3. change our whitelist semantics to apply to extra items in either CVMFS config OR Topology

@brianhlin would you care to weigh in on this?

brianhlin: I vote (1) if the cache is active and (2) if it isn't

edquist: @biozit, any idea whether the `stashcache.gwave.ics.psu.edu:8443` cache is still active?

edquist: Ping @biozit -

> any idea whether the `stashcache.gwave.ics.psu.edu:8443` cache is still active?

biozit: Looks offline:

[1155] root@test-007 ~/cachetest# /usr/bin/gfal-copy --cert /root/cachetest/c -f  https://stashcache.gwave.ics.psu.edu:8443/user/ligo/test_access/access_ligo  .
gfal-copy error: 110 (Connection timed out) - Could not stat the source: Connection timed out

[1204] root@test-007 ~/cachetest# ping stashcache.gwave.ics.psu.edu
PING stashcache.gwave.ics.psu.edu (10.136.28.245) 56(84) bytes of data.
^C
--- stashcache.gwave.ics.psu.edu ping statistics ---
5 packets transmitted, 0 received, 100% packet loss, time 3999ms

Let me check it with the admin

biozit: It is a private cache, @edquist

edquist: So in that case I think instead of adding to this whitelist, we should remove the `XRootD cache server` service from the `PSU-LIGO-CACHE` resource ([here](https://github.com/opensciencegrid/topology/blob/master/topology/Penn%20State%20University/PSU%20LIGO/PSU-LIGO.yaml#L63-L65)) ... @brianhlin, does that match your thinking?

biozit: Ron Tapia
  [12:33 PM]
Hi, yes it is. PSU-LIGO is using it for CVMFS.
Fabio Andrijauskas
  [12:36 PM]
is it a “private” cache?
Ron Tapia
  [12:40 PM]
Yes, in the sense that its ports are not exposed to the outside world.







Ron Tapia
  [12:40 PM](https://opensciencegrid.slack.com/archives/D04C0TTLPHP/p1668804022968489)
Yes, in the sense that its ports are not exposed to the outside world.

brianhlin: @djw8605 @matyasselmeci this is an interesting case here: even if the PSU-LIGO cache is private, it feels like we should keep it registered in Topology for troubleshooting, monitoring, and accounting.  However, if it's private then I imagine that they should be generally excluded from the lists of caches that are used by the `osdf-client` and CVMFS. Do we have a way to mark caches as "private"?

@biozit how do jobs at PSU know to use their cache, specifically? Or do they not do anything on their end and any jobs outside of PSU that determine that PSU is the closest cache attempt to use it then fail over to public caches?

matyasselmeci: > Do we have a way to mark caches as "private"?

Not for osdf-client; I think that's related to https://opensciencegrid.atlassian.net/browse/SOFTWARE-5203.

brianhlin: Ah, right. That's important but not sufficient if we want to keep the cache registered in Topology: we'd need to have a way to mark a cache as excluded from the default list used by the `osdf-client`

edquist:
@matyasselmeci and @brianhlin -

In SOFTWARE-3470 you've both suggested adding sites to the whitelist.

But see my [comment above](https://github.com/opensciencegrid/topology/pull/2759#discussion_r1019577083) - in particular:

>This whitelist is for caches configured in CVMFS, that are not in topology.

Are you suggesting that we do option 3?

>3. change our whitelist semantics to apply to extra items in either CVMFS config OR Topology

biozit:
@brianhlin it is possible to select the cache list on stashcp, using -n

biozit:
@brianhlin Should we merge this?

biozit:
For the cache information, we have a list of OSG caches and Ligo caches: https://github.com/cvmfs-contrib/config-repo/blob/osg/etc/cvmfs/domain.d/osgstorage.org.conf

The osdf-client can be configured to use different cache lists.

matyasselmeci:
Hi folks,
What's up with this PR? Should we merge it? Should we close it? Are additional changes necessary?

edquist:
>What's up with this PR?

Brian commented [above](https://github.com/opensciencegrid/topology/pull/2759#issuecomment-1335637920) saying to add a(nother) whitelist mechanic.  I think that requires code changes (and i don't think there is a ticket).  The whitelist addition in this PR does not work as-is for reasons I [explained earlier](https://github.com/opensciencegrid/topology/pull/2759#discussion_r1019577083).

biozit:
Should we create a Jira ticket and close this PR?

matyasselmeci:
I see. @biozit, would https://opensciencegrid.atlassian.net/browse/SOFTWARE-5482 address your needs?

biozit:
sounds good!

matyasselmeci:
OK, good.  Will this PR work for excluding the cache from CVMFS, or does that require something else? I'm not sure what mechanism CVMFS uses to get its cache list.

biozit:
We need to change here https://github.com/cvmfs-contrib/config-repo/blob/osg/etc/cvmfs/domain.d/osgstorage.org.conf for CVMFS

matyasselmeci:
OK, in that case I think we can close this PR.

PR #2769, opened 2022-11-15
Update type hints

matyasselmeci:
Most of our GlobalData.get_* methods may return None if obtaining the data the first time raises an exception.


matyasselmeci:
Let me know if you think I should update docstrings/comments with this info too...

PR #2770, opened 2022-11-16
Update UNLCachingInfrastructure.yaml

biozit:
@djw8605 this is the only way I can create the certs from UCSD :-(


djw8605:
I'm not super excited about this.  I would much rather put it under the opensciencegrid.org or osgstorage.org domains, just to keep it institutionally neutral.

But, if this is our only option, I'm ok with it.  At least we can change it in the future rather easily.

biozit:
You said something about requesting certs from somewhere

biozit:
certs installed

biozit:
Will be the same with Massachusetts Green High Performance Computing Center

bbockelm:
@biozit - should we be closing this PR?

biozit:
yes, I do it.

PR #2778, opened 2022-11-22
Move webapp.common.generate_dn_hash() to a new file, x509.py

matyasselmeci:
This lets you import webapp.common without having "asn1" installed, which is good because get-scitokens-mapfile uses webapp.common and asn1 isn't available as an RPM on EL7.

### DIFF
--- a/src/webapp/common.py
+++ b/src/webapp/common.py
@@ -4,13 +4,10 @@ import hashlib
 import json
 import os
 import re
-import shlex
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


matyasselmeci:
I accidentally introduced this dependency in 1e03a5eb4c180806b084510d5a7016b75cd735e9 when I did some refactoring for OSDF stuff and it's now preventing me from building osg-scitokens-mapfile.

brianhlin (approved):
LGTM

PR #2779, opened 2022-11-23
Make x509 (hence asn1) optional in webapp.vos_data

matyasselmeci:
get-scitokens-mapfile imports webapp.vos_data to handle scitoken auth so I need to do a little extra work to make asn1 optional.


brianhlin (approved):
LGTM

PR #2782, opened 2022-11-29
SOFTWARE-5347

AidenCohen31:


src/topology_utils.py
```diff
@@ -45,19 +49,33 @@ def get_auth_session(args):
     if args.key:
         key = args.key
 
-    session = requests.Session()
+    session = {}
+<<<<<<< HEAD
 
+=======
+    session['cert_reqs'] = 'CERT_REQUIRED'
+>>>>>>> master
```
brianhlin, line 57: There's a couple of diff-looking lines that seemed to have snuck in here somehow. Can you clean them up?

AidenCohen31:
Oh, that was an accident, I guess I didn't push the merge

AidenCohen31:
Revamped structure, should be good now

bin/osg-notify
```diff
@@ -21,6 +21,7 @@ if __name__ == "__main__" and __package__ is None:
     sys.path.append(_parent + "/src")
 
 import topology_utils
+from topology_utils import OSGPoolManager
```
brianhlin, line 24:
```suggestion
from topology_utils import TopologyPoolManager
```

bin/osg-notify
```diff
@@ -173,13 +174,13 @@ def has_non_printable_ascii_characters(contents):
 
 def main():
     args = parseargs()
-
+    pm = OSGPoolManager()
```
brianhlin, line 177:
```suggestion
    pm = TopologyPoolManager()
```

bin/osg-notify
```diff
@@ -190,7 +191,7 @@ def main():
                     exit(1)
                 else:
                     print(exc)
-        results = topology_utils.filter_contacts(args, results)
+        results = OSGPoolManager.filter_contacts(args, results)
```
brianhlin, line 194:
```suggestion
        results = TopologyPoolManager.filter_contacts(args, results)
```

bin/osg-notify
```diff
@@ -213,7 +214,7 @@ def main():
                     exit("Too many incorrect password attempts, exiting")
                 else:
                     print(exc)
-        results = topology_utils.filter_contacts(args, results)
+        results = OSGPoolManager.filter_contacts(args, results)
```
brianhlin, line 217:
```suggestion
        results = TopologyPoolManager.filter_contacts(args, results)
```

src/topology_utils.py
```diff
@@ -25,365 +25,375 @@ class InvalidPathError(AuthError):
...
-    and the arg `contact_list` is the contents of a single <ContactList>
-
-    If vosummary, this looks like:
-        <ContactTypes>
-            <ContactType>
-                <Type>Miscellaneous Contact</Type>
-                <Contacts>
-                    <Contact>
-                        <Name>...</Name>
+class OSGPoolManager(urllib3.PoolManager):
```
brianhlin, line 28:
```suggestion
class TopologyPoolManager(urllib3.PoolManager):
```

brianhlin (commented):
We have a service called the "OSPool" so I'm renaming the class to avoid tricky typos in the future

brianhlin (changes requested):
Hey @AidenCohen31 this looks solid but could you take all the static methods out of the `TopologyPoolManager` class?

As you already know, these methods don't act on a given instance of the class so they don't need to be there and we should be especially conservative when changing an interface of the library. It's anybody's guess who's using these methods from our library and if we change the way they need to invoke the methods, their code may break!

PR #2788, opened 2022-12-01
Create MGHPCCOrigin.yaml and fix cache

biozit:


virtual-organizations/IceCube.yaml
```diff
@@ -58,13 +58,24 @@ DataFederations:
         AllowedOrigins:
           - IceCube_StashCache_origin
           
-      - Path: /icecube/NRP-UNL/PROTECTED
+      - Path: /nrpnb/protected/icecube/
```
bbockelm, line 61: The custom is to have the top-level component of the origin reflect the VO.  It appears this is indicating the origin.

Can you do:

```
/icecube/nrpnb/protected
```

instead?

Note XRootD has various mechanisms to add or subtract prefixes before it tries to open the file on disk.

virtual-organizations/IceCube.yaml
```diff
@@ -58,13 +58,24 @@ DataFederations:
...
         AllowedOrigins:
           - NEBRASKA_NRP_OSDF_ORIGIN
         AllowedCaches:
           - ANY
+          
+      - Path: /mghpcc/protected/icecube/
+        Authorizations:
+          - SciTokens:
+              Issuer: https://chtc.cs.wisc.edu/icecube
+              Base Path: /mghpcc/protected/icecube/
```
bbockelm, line 76: Similar to above -- make the base component of the namespace the VO not the origin name.

bbockelm (changes requested):
I'd like to get at least the top-level of the namespace to align with the best practices on the OSDF.

virtual-organizations/IceCube.yaml
biozit: sure, solved!

virtual-organizations/IceCube.yaml
biozit: sure, solved!

biozit:
Please, is there anything to fix here? @bbockelm @djw8605

bbockelm (approved):
LGTM.  Ready to merge.

PR #2809, opened 2022-12-09
Add new OSDF Origin resource for ET VO to UCL-INGRID.

AndresTanasijczuk:
This is to register a new origin server for the newly created ET VO. The origin will be a public one. My plan is to run it in a container on the same host machine that is currently running the private origin `UCL-Virgo-StashCache-Origin`. Therefore this new origin has the same DN and FQDN as the `UCL-Virgo-StashCache-Origin`  origin. I discussed this with @djw8605 who said this configuration should be ok and there should be no problem in having two origin resources with same DN and FQDN in topology.

### DIFF
--- a/topology/Universite catholique de Louvain/UCL-INGRID/UCL-INGRID.yaml
+++ b/topology/Universite catholique de Louvain/UCL-INGRID/UCL-INGRID.yaml
@@ -1,11 +1,11 @@
 Production: true
 SupportCenter: Self Supported
-GroupDescription: StashCache Origin in Louvain
+GroupDescription: OSDF Origins in Louvain
 GroupID: 1097
 Resources:
   UCL-Virgo-StashCache-Origin:
     Active: true
-    Description: Virgo StashCache Origin
+    Description: Virgo OSDF Origin
     ID: 1108
     ContactLists:
       Administrative Contact:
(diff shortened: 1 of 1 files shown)


brianhlin:
@edquist IIRC you wrote the CI for checking duplicate FQDNs. Since this has come up before, how difficult would it be to skip that check for `XRootD origin server` services?

brianhlin:
> It is not correct that having two origins with the same FQDN will not cause problems. Authfile and scitokens.conf lookup is done using the FQDN as the key; if you have two origins with the same FQDN, it is indeterminate which one will "win".

Agh, right. And we haven't implemented resource name lookups for authfile/scitokens.conf, right?

matyasselmeci:
No we have not.  https://opensciencegrid.atlassian.net/browse/SOFTWARE-4347

brianhlin:
@AndresTanasijczuk since you're running a container, I think we could have a resource name based lookup done before Christmas. Would that work for you?

AndresTanasijczuk:
Yes, that would be fine for me.
Otherwise, would the problem be circumvented if I define a DNS alias for the new service?

brianhlin:
> Yes, that would be fine for me.
Otherwise, would the problem be circumvented if I define a DNS alias for the new service?

Yes, that should work, too!

AndresTanasijczuk:
I changed the FQDN to a new name, `et-origin.cism.ucl.ac.be`. This will be a DNS alias for `ingrid-se09.cism.ucl.ac.be` (we just requested the alias). I also removed the DN (this is not really needed for a public origin, right?).

edquist:
>edquist IIRC you wrote the CI for checking duplicate FQDNs. Since this has come up before, how difficult would it be to skip that check for XRootD origin server services?

@brianhlin, just to confirm, is this still something you want?

It seems from this issue history that it's not anymore; and in any case a dns alias should solve the problem.

But (just in case) if you still want to skip that check for XRoodT origin server services -- presumably you mean the check that "fqdns ... MUST be unique for XRootD services", in which case we could enforce for `XRootD cache server` but drop the check for `XRootD origin server` trivially by removing that name from the `xrd_svcs` list here:

https://github.com/opensciencegrid/topology/blob/master/src/tests/verify_resources.py#L806

matyasselmeci:
@edquist we don't want it anymore.

matyasselmeci (approved):

PR #2821, opened 2022-12-15
Move S3 demos to the OSG VO (SOFTWARE-5398)

brianhlin:
We want to use AP7, which is configured with the OSPool token issuer


brianhlin:
Lengthy comment in https://github.com/opensciencegrid/topology/pull/2821/commits/9911892e54ac6f7a1ef83e10ae9a0a7a3f9b0d84 describing the weird looking namespace config

This commit consolidates sections for the https://osg-htc.org/ospool
in the scitokens.conf generated by Topology so that caches will
respect the OSPool issuer + multiple base paths.

xrootd-scitokens has multiple problems with duplicate issuers in its
config (both the section name and the attribute name).

1.  Duplicate `[Issuer ]` sections currently yield malformed values
    for attributes that are specified across these duplicate
    sections (values are merged and `\n` separated). This is addressed by
    choosing the last value for the duplicated section + attribute:

    - https://github.com/xrootd/xrootd/pull/1860
    - https://github.com/xrootd/xrootd/pull/1862

    These changes have not yet been released.

2.  Even with the fix for (1), this will result in only the last value
    of the duplicate sections being respected, which could cause
    transfers for end users to fail

So to work around all that, we hack this all into a single namespace
entry, which is fine since we don't use X.509 credentials and Path is
only used for namespaces that respect X.509.

PR #2822, opened 2022-12-15
Use a config hash as the block name for each scitokens.conf "Issuer" block (SOFTWARE-5412)

matyasselmeci:
This keeps us from having multiple "Issuer" blocks with the same name.


djw8605:
Can we have the human readable name as well part of the section name?

matyasselmeci:
By human readable name do you mean the issuer URL? Also can I use a space to separate the URL from the hash?

jthiltges:
I think spaces in the section name should be fine. Just needs to start with `Issuer`
https://github.com/xrootd/xrootd/blob/65dd72f1249119546bd1506e56758225a4cbcc7c/src/XrdSciTokens/XrdSciTokensAccess.cc#L1113

Instead of hashes, could do something similar with a counter on the end, maybe:
```
class SciTokenAuth():
    counter = 0

    def __init__(self):
        SciTokenAuth.counter += 1
        self.idx = SciTokenAuth.counter
```

matyasselmeci:
The counter won't work; it would keep incrementing each time we reload the data (whenever it gets updated) until we restarted the webapp at which point it would reset.

brianhlin (approved):
LGTM!

PR #2824, opened 2022-12-15
Update OSG.yaml

biozit:
To test the XrootD patch


biozit:
@matyasselmeci, Could you check if this is ok? We need a VO configured in order to the origin to start correctly

matyasselmeci (approved):

PR #2827, opened 2022-12-16
Remove duplicate identical issuer blocks (SOFTWARE-5412)

matyasselmeci:


matyasselmeci:
Tagged as v1.24.1 and deployed on topology-itb.

PR #2828, opened 2022-12-16
Use separate Paths for the S3 namespaces (SOFTWARE-5398)

matyasselmeci:
This results in the same scitokens.conf files as if we had compressed the namespaces into one path, but lets us have different Writeback and DirList values.


matyasselmeci:
If I revert the addition of the hash to the block name, then #2827 will deduplicate all the issuer sections before generating the scitokens.conf file since they'll be completely identical.

matyasselmeci:
We discussed this and agreed that it's OK to merge once #2827 got deployed.

PR #2829, opened 2022-12-16
Revert "Use a config hash as the block name for each scitokens.conf "Issuer" block (SOFTWARE-5412)"

matyasselmeci:
Reverts opensciencegrid/topology#2822

I don't need a hash if I deduplicate identical sections in #2827.


brianhlin (approved):
LGTM

PR #2833, opened 2022-12-19
Limit caches for the S3 demo to CHTC-controlled ones

matyasselmeci:


brianhlin:
Turns out that there were issues at the origin and we don't need this

PR #2836, opened 2022-12-20
Topology Project Form ( EC-165 )

CannonLock:


src/webapp/forms.py
```diff
@@ -301,3 +302,145 @@ def get_yaml(self) -> str:
...
+        ('Statistics', 'Statistics'),
+        ('Technology', 'Technology'),
+        ('Training', 'Training'),
+        ('Zoology', 'Zoology')
+    ])
+
+    description = TextAreaField(None, render_kw={
+        "style": "font-family:monospace; font-size:small;",
+        "rows": "5"
+    })
```
brianhlin, line 401: Oof, can you generate this from the file on disk instead?

CannonLock: Good call

brianhlin:
One thing that just came up: the RCFs have a policy of naming their projects `<Institution short name>_<PI last name>`. Where the short name comes from mapping the project's institution via https://github.com/opensciencegrid/topology/blob/master/mappings/project_institution.yaml.

Would the form be able to set the default project name but allow an optional override?

CannonLock:
Demo of PR -> https://drive.google.com/file/d/1wSy7-nNBa_3AhGfo49Ptt450KOW_kmT4/view?usp=share_link

Closed PR -> https://github.com/opensciencegrid/topology/pull/2854

Demo of Automatic Project Name -> https://drive.google.com/file/d/1Xjxr4_QWEq-VexhnIMSXnmtTx_chBpme/view?usp=share_link

CannonLock:
Overview of changes:
- Added a new form to add OSG Projects
- Added a way to automatically create PRs
  - User logs in to Github with Oauth so I can grab their username and email
  - I push a commit to a branch on a bot account in the users name
  - I create a PR from the bot account branch to the main topology branch

I chose to do it this way because it minimizes the things I need to do on a users behalf.

brianhlin (commented):
I took a look at the form + the PRs that were generated but one thing it looks like we're missing is the `Sponsor` field, e.g. https://github.com/opensciencegrid/topology/blob/master/projects/ACE_NIAID.yaml#L17-L19.

Since there's a lot of attention being paid to Topology and what we call things, it wouldn't hurt to consider sorting out this data.

@djw8605 does GRACC use the sponsor data at all?

@matyasselmeci @ChristinaLK @rachellombardi @ppaschos tagging a project as OSPool probably makes sense. Would a project cross pools (i.e., should this be multi-valued)? What about collaboration projects?

matyasselmeci:
Besides what we call "campus grids" like osg-connect, the Sponsor field can be one of the collaborations in the `virtual-organizations` directory.  Even if it's not in GRACC, we may want to correlate project with collaboration for reporting.

matyasselmeci:
"Organization" is free form whereas "Sponsor" is checked against our known campus grids and collaborations.

brianhlin:
Ya know, if we're grabbing pool info from the accounting records (https://opensciencegrid.atlassian.net/browse/SOFTWARE-5367), I think we can ignore my previous questions except for:

> @djw8605 does GRACC use the sponsor data at all?

If the answer is no, then we can move onto the code review portion of this PR

brianhlin:
Derek's answer is no it doesn't so this form is good to go, at least in what it generates. @edquist can you add it onto your list to do the code review?

edquist:
Sure thing, will likely get through it tomorrow.  Looks like this is the first time someone else has dared to touch the github code, in the 4 years it's been around!  :)

src/webapp/models.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, config=None, strict=False):
         self.ligo_ldap_passfile = config.get("LIGO_LDAP_PASSFILE")
         self.ligo_ldap_url = config.get("LIGO_LDAP_URL")
         self.ligo_ldap_user = config.get("LIGO_LDAP_USER")
+        self.github_oauth_client_secret = config.get("GITHUB_OAUTH_CLIENT_SECRET")
+        self.osg_bot_github_token = config.get("OSG_BOT_GITHUB_TOKEN")
```
edquist, line 82: Is this `OSG_BOT_GITHUB_TOKEN` different from the `WEBHOOK_GH_API_USER` / `WEBHOOK_GH_API_TOKEN` a little bit above?  (The `osg-bot` user uses these for reviewing/commenting/merging PRs.)

If they are different i wonder if there's a way to come up with a clearer naming scheme here to distinguish them.

edquist: If i'm understanding the code right i think you are using a hard-coded `osg-doc-bot` user along with your `OSG_BOT_GITHUB_TOKEN` ?

If so then maybe to fit the model you could make that configurable also and use `DOC_GH_API_USER` / `DOC_GH_API_TOKEN` (or some other prefix that captures what this bot user is for).

edquist: maybe `FORM_*` ...

src/webapp/github.py
```diff
@@ -16,12 +18,70 @@ def mk_github_authstr(user, passwd):
...
+def http_to_dict(success, http_response, throw_error=True):
+    """http json response to dict"""
+
+    if success:
+        return json.load(http_response.fp)
+
+    if throw_error:
+
+        if http_response == "Not Found":
+            raise GithubNotFoundException(f"Request not successful: {str(http_response)}")
```
edquist, line 38: I believe `str(x)` is implicit for `f"{x}"` (as well as `"{}".format(x)` and `"%s" % x`)

```suggestion
            raise GithubNotFoundException(f"Request not successful: {http_response}")
```

src/webapp/github.py
```diff
@@ -16,12 +18,70 @@ def mk_github_authstr(user, passwd):
...
+
+        elif http_response == "Reference already exists":
+            raise GithubReferenceExistsException
+
+        else:
+            raise GithubRequestException(f"Request not successful: {str(http_response)}")
```
edquist, line 44:
```suggestion
            raise GithubRequestException(f"Request not successful: {http_response}")
```

src/webapp/github.py
```diff
@@ -16,12 +18,70 @@ def mk_github_authstr(user, passwd):
...
+        self.email = email
+
+    @classmethod
+    def from_token(cls, token):
+        gh_auth = GitHubAuth(None, token)
+        user_response = http_to_dict(*gh_auth.get_user())
+        email_response = http_to_dict(*gh_auth.get_user_email())
+
+        # Grab the first visible email to use, it doesn't matter
+        visible_email = list(filter(lambda x: x['visibility'] != 'private', email_response))[0]['email']
```
edquist, line 74: is there guaranteed to be at least one visible email?

src/webapp/github.py
```diff
@@ -76,7 +136,10 @@ def publish_pr_review(self, owner, repo, num, body, action, sha):
         # action: APPROVE, REQUEST_CHANGES, or COMMENT
         api_path = "/repos/:owner/:repo/pulls/:number/reviews"
         url = api_path2url(api_path, owner=owner, repo=repo, number=num)
-        data = {'event': action, 'commit_id': sha}
+        data = {
+            'event': action,
+            'commit_id': sha
+        }
```
edquist, line 142: Did your editor do this for you while you weren't looking?

src/webapp/github.py
```diff
@@ -88,35 +151,202 @@ def hit_merge_button(self, owner, repo, num, sha, title=None, msg=None):
         api_path = "/repos/:owner/:repo/pulls/:number/merge"
         url = api_path2url(api_path, owner=owner, repo=repo, number=num)
         data = {}
-        if sha:    data['sha']            = sha
-        if title:  data['commit_title']   = title
+        if sha:    data['sha'] = sha
+        if title:  data['commit_title'] = title
```
edquist, line 155: same question ... i rather like the stacked equal signs, don't you?

src/webapp/github.py
```diff
@@ -88,35 +151,202 @@ def hit_merge_button(self, owner, repo, num, sha, title=None, msg=None):
...
     # wrapper around GitHubAuth with (owner,repo) specified up front
 
-    def __init__(self, ghauth, owner, repo):
+    def __init__(self, ghauth: GitHubAuth, owner, repo):
         self.ghauth = ghauth
         self.owner = owner
         self.repo = repo
 
+        # Dump in the repos properties for easy access
+        repo_data = http_to_dict(*self.get_repo())
```
edquist, line 252: I'm not wild about making implicit api calls just because i've created one of these helper objects.  If you want to get the repo info for something can it be done with an explicit setup call or something?

src/webapp/github.py
```diff
@@ -88,35 +151,202 @@ def hit_merge_button(self, owner, repo, num, sha, title=None, msg=None):
...
+        for key, value in repo_data.items():
+            setattr(self, key, value)
```
edquist, line 254: I'm really not wild about setting object attributes to whatever is returned by the request.  Not only does it make it less clear what you're expecting to see from the response, it also seems like a security problem because you are potentially overriding method names here too.

I suggest you make some kind of setup method that you explicitly call, which does `self.repo_info = self.get_repo()` or similar.

src/webapp/github.py
```diff
@@ -49,7 +109,7 @@ def github_api_call(self, method, url, data):
         self._add_auth_header(req)
         #add_gh_preview_header(req)
         try:
-            resp = urllib.request.urlopen(req)
+            resp = urllib.request.urlopen(req, cafile=certifi.where())  # potentially need cafile=certifi.where()
```
edquist, line 112:
```suggestion
            # potentially need cafile=certifi.where()
            resp = urllib.request.urlopen(req, cafile=certifi.where())
```

:thinking: just curious did you run into a case where you needed this?

Also @CannonLock, it's about time I tell you because nobody else on the team will.

The 79-wide rule is the most important rule in PEP-8.

src/webapp/github.py
```diff
@@ -88,35 +151,202 @@ def hit_merge_button(self, owner, repo, num, sha, title=None, msg=None):
...
+        api_path = "/user/emails"
+        url = api_path2url(api_path)
+
+        return self.github_api_call("GET", url, None)
+
     def get_api_url(self, url):
         return self.github_api_call('GET', url, None)
 
-    def target_repo(self, owner, repo):
-        return GitHubRepoAPI(self, owner, repo)
```
edquist, line 101: You no likie?

This function is used elsewhere (in webhook_app.py) ... i'm thinking it can stay in - you may even want to try it out.

src/app.py
```diff
@@ -697,6 +720,127 @@ def render_form(**kwargs):
...
+        try:
+            # Gather necessary data
+            create_pr_data = {
+                "file_path": f"data/{request.values['project_name']}.yaml",
+                "file_content": form.get_yaml(),
+                "branch": f"add-project-{request.values['project_name']}",
+                "message": f"Add Project {request.values['project_name']}",
+                "committer": GithubUser.from_token(session["github_login"]['access_token']),
+                "fork_repo": GitHubRepoAPI(GitHubAuth("osg-doc-bot", app.config["OSG_BOT_GITHUB_TOKEN"]), 'osg-doc-bot', 'topology'),
+                "root_repo": GitHubRepoAPI(GitHubAuth("osg-doc-bot", app.config["OSG_BOT_GITHUB_TOKEN"]), 'opensciencegrid', 'topology'),
```
edquist, line 772: So for what it's worth i had intended the GitHubAuth/GitHubRepoAPI setup to look a bit more like this:

```suggestion
                "root_repo": GitHubAuth(self.form_gh_api_user, self.form_gh_api_token, app.logger).target_repo('opensciencegrid', 'topology'),
```

which returns the GitHubRepoAPI object ... I intended never to have to _type_ GitHubRepoAPI.

(Oh and the extra `app.logger` arg i added to GitHubAuth is optional.)

src/app.py
```diff
@@ -697,6 +720,127 @@ def render_form(**kwargs):
...
+        redirect_data = {}
+        try:
+            response = requests.post("https://github.com/login/oauth/access_token", data=data, headers={"Accept": "application/json"})
+            session['github_login'] = response.json()
+
+        except Exception as error:
+            redirect_data['error'] = f"Could not complete exchange with Github for Token: {error}"
+
+        return redirect(
+            f"{urllib.parse.urlparse(request.args['redirect_uri']).geturl()}?{urllib.parse.urlencode(redirect_data)}"
```
edquist, line 826: I know the cool kids are all about the f-strings these days, but when you have big expressions like these embedded in the interpolation parts, it's really difficult to parse visually to see what is string vs code.

Compare:

```suggestion
            "{}?{}".format(
                urllib.parse.urlparse(request.args['redirect_uri']).geturl(),
                urllib.parse.urlencode(redirect_data)
            )
```

Or even `"%s?%s" % (..., ...)`, if you are down with the primordial syntax.

Anyway food for thought for you, here and other such places.

f-strings are fine when you want to do `f"foo={foo} bar={bar}"` - this is supposed to enhance readability. But when f-strings get to be mostly code it loses that.

edquist:
@CannonLock i'll have to continue with this next week.  It's fun to see someone else playing with the github code.  I hope you found some of it useful / not too painful to use.  And, don't mind my fussy comments, they all come from a place of love :D

src/webapp/github.py
CannonLock: Yes, if a user decides they want their email private then Github will make a hashed public email for the user.

src/webapp/github.py
CannonLock: I probably did it well I wasn't looking, I sometimes mindlessly format when I am thinking.

PR #2838, opened 2022-12-20
Rename virtual-organizations -> collaborations and make compat symlinks

matyasselmeci:
Since it's hard to tell in a GitHub diff, I did:

```
git mv virtual-organizations collaborations
git mv template-virtual-organization.yaml template-collaboration.yaml
ln -s collaborations virtual-organizations
ln -s template-collaboration.yaml template-virtual-organization.yaml
git add virtual-organizations template-virtual-organization.yaml
```


brianhlin (approved):
What could possibly go wrong?

edquist:
note that `.dockerignore` references `virtual-organizations` - not sure if you wanna update that now that `virtual-organizations` is just a symlink.

... a number of places in the code reference `virtual-organizations`, but as far as i can tell i guess it should still work with the symlink

matyasselmeci:
Good catch, I didn't think about .dockerignore.

PR #2855, opened 2023-01-04
Remove Inactive Sites from Map Display

CannonLock:


djw8605 (approved):
So fancy, filter and reduce functions.  Looks good to me.

PR #2866, opened 2023-01-09
Add IGWN CILogon issuer to scitokens mapfile

jasoncpatton:
Attn @astroclark... we can change this to map to the `igwn` user, but that would require sites that are using the default OSG scitokens mapfile to create a `igwn` service account on their cluster(s). Either way, sites not using the default mapfile will have to update their custom mapfiles with the new issuer (should be added rather than replacing the old mapping).


matyasselmeci:
Jason, do we still need this?

brianhlin:
This appears to be abandoned, closing

PR #2887, opened 2023-01-18
Create Institution Endpoint (SOFTWARE-5443)

CannonLock:
Add Institution Endpoint
Add Tests for Institution Endpoint


CannonLock:
Had some fun and added the ability for client to use Accept header so I can use this as a json endpoint as well.

matyasselmeci:
I like the idea but can you move `create_accepted_response` out of `common.py`? `common.py` gets used (via `models.py`) by some of the scripts in `bin/` such as `get-scitokens-mapfile`, so I don't want any Flask dependencies in there. I don't think it fits into any of the others in `webapp` so make a new file, maybe called `flask_common.py`.

(In retrospect, I shouldn't have named that directory `webapp` since it contains modules used by scripts as well...)

CannonLock:
Works for me!

CannonLock:
@matyasselmeci Just made the switch, tests still pass

src/webapp/flask_common.py
```diff
@@ -0,0 +1,22 @@
+from flask import Response
+from .common import to_csv, to_json_bytes
+
+
+def create_accepted_response(data: list, headers, default=None) -> Response:
```
matyasselmeci, line 5:
```suggestion
from typing import List

def create_accepted_response(data: List[str], headers, default=None) -> Response:
```
If it's a list of strings you should put that in the annotation. Note that the webapp uses Python 3.6 which means you can't use the lowercase `list` for a type hint (that was added in [3.9](https://peps.python.org/pep-0585/) )

CannonLock:
Hey Mat, I just switched it to a List because I am actually expecting a List[List[str]]. Thanks for the hint on the 3.6 annotations I didn't know that.

matyasselmeci (approved):
LGTM

PR #2896, opened 2023-01-19
have osg-bot ping software-and-release team for attention

edquist:
Feel free to hold off on this one or reject it completely if you aren't wild about the idea.


matyasselmeci:
The code is fine; I already get pinged for everything Topology so this has no effect on me. @brianhlin what do you think?

brianhlin:
I like the idea in principle but the rest of the folks may not https://github.com/orgs/opensciencegrid/teams/software-and-release/members. I like this PR as a proof-of-concept at the very least, we may want to ping the project office at some point and it's useful to know where we need to make the changes!

matyasselmeci:
Let's not leave this open forever.

PR #2899, opened 2023-01-21
Add simple Prometheus update metrics

bbockelm:
We are observing some timeouts in the topology webapp logs around update intervals.  Provide some clarity by adding a Prometheus endpoint and exporting the update timing.

### DIFF
--- a/requirements-apache.txt
+++ b/requirements-apache.txt
@@ -17,6 +17,7 @@ Jinja2>=2.11.3,<2.12
 ldap3>=2.6.1
 MarkupSafe<2.1
 # ^^ MarkupSafe stuck on 2.0.x due to removal of "soft_unicode" symbol
+prometheus-client>=0.15.0,<0.16
 python-dateutil>=2.7.3,<2.8
 python-gnupg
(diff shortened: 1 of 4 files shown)


matyasselmeci (approved):
LGTM though I don't have any experience with either prometheus-client and or DispatcherMiddleware. How do you test it?

Also, can you add a link to the `/metrics` endpoint to the front page (`src/templates/homepage.html.j2`, under "XML and other machine-readable data")

matyasselmeci:
I can add it too.

matyasselmeci:
I just realized this adds a dependency on prometheus-client to models.py. We can't do that -- models.py is also used by various scripts such as osg-scitokens-mapfile. Sorry, I have to revert.

bbockelm:
@matyasselmeci - how should we resolve this?  Create dummy objects if the import fails?

matyasselmeci:
That's the approach I'm thinking of, too.

brianhlin:
> Create dummy objects if the import fails?

@matyasselmeci did you get around to this?

PR #2910, opened 2023-01-25
Fix hostname of ComputeCanada-Cedar-Cache

matyasselmeci:
From the logs in FD #71777, and the DN, it looks like this is the correct hostname.


biozit:
We had floated a few hostnames around in our Slack channel, so I mistyped this when adding the github topology entry.  These should all be osg-stash.sfu.computecanada.ca (that is the actual FQDN).

Leslie

biozit (commented):
Looks good!

biozit (commented):
looks good!

PR #2911, opened 2023-01-25
Add name filters (SOFTWARE-5442)

CannonLock:
Added ability to filter based on the name of a facility, support center, resourcegroup, site, vo, voown, and service

Did a little refactoring of the Filter class

Added tests for the above


CannonLock:
I can delete the parts that add the ability to filter by name for this other the rg if they aren't need. Just made sense to do them all at once.

PR #2913, opened 2023-01-27
Security team GPG key is for @opensciencegrid.org (SOFTWARE-5455)

brianhlin:


matyasselmeci:
Can't they add the new email address to their GPG key?  GPG keys can have multiple identities associated with them -- [mine](https://keyserver.ubuntu.com/pks/lookup?search=matyas+selmeci&fingerprint=on&op=index) certainly does.

brianhlin:
Ya but until they do that I want to make sure that they don't have any issues with their announcements

matyasselmeci (approved):

PR #2918, opened 2023-01-31
Adding FTS only downtime 

DAMason:


matyasselmeci (approved):

brianhlin:
@DAMason could you add yourself as a contact to the relevant resources? Otherwise your downtime requests will sit idle until someone on the OSG side can review and merge them

PR #2921, opened 2023-01-31
Update PATh origin FQDN (INF-639)

brianhlin:
chtc.wisc.edu isn't Tiger managed so we can't automatically get certs for it


edquist (approved):
If you say so Boss! :)

PR #2930, opened 2023-02-02
Issuer should not have a trailing slash (INF-406)

brianhlin:
As specified in https://github.com/path-cc/path-cc.github.io/blob/master/.well-known/openid-configuration#L2


edquist (approved):
LGTM

PR #2936, opened 2023-02-07
Setting "offline" caches due to maintenance

biozit:
Removing the following caches due to ongoing maintenance:

 - ComputeCanada-Cedar-Cache: Still in the configuration step
 - T1_STASHCACHE_CACHE: Updading to OSG 3.6
 - Stashcache-Cardiff: Trying to change the certificate


matyasselmeci (approved):

matyasselmeci:
Merging -- the CI failure is because we look at the Cardiff cache in our CI tests.  I'll fix that separately.

PR #2937, opened 2023-02-07
Update MWT2.yaml

ivukotic:
removed squids that don't exist anymore.


brianhlin (approved):
LGTM! Have these been replaced by Varnish? We should consider registering those hosts @ivukotic @LincolnBryant @jlstephen @djordan66

ivukotic:
These are just the ones that don’t exist anymore.
Slate Squids are still in, and Slate Varnishes are also already in.
Ilija

PR #2968, opened 2023-02-21
Use Python 3.9 and a newer version of Werkzeug

matyasselmeci:
Move the Topology webapp to Python 3.9 so I can use a new version of Werkzeug.

It was already on EL8 but I moved it to Alma 8 explicitly. I used "alternatives" to point /usr/bin/python3 and friends to Python 3.9 so I don't have to change shebang lines.


matyasselmeci:
What a rabbit hole!

brianhlin (approved):
LGTM!

PR #2975, opened 2023-02-22
move Maine site yaml into the directory

coldbeer101:
moving the site yaml into the appropriate directory


matyasselmeci:
You'll need to move the SITE.yaml file as well.

coldbeer101:
ok moved that one as well. Thanks for catching that!

matyasselmeci (approved):

PR #2980, opened 2023-02-23
disable old alabama resource and change FQDN to avoid collision with …

jdost321:
…new Alabama-CHPC-CE1 resource

Since my last change we are not getting GRACC accounting. After discussing with Derek we agreed it might be because we used the same FQDN in the old and new resources and that should be unique. Disable old resource and change its FQDN to avoid collisions.


brianhlin (approved):
LGTM

PR #2982, opened 2023-02-24
Update to filter by GRACC Sites over Facilities

CannonLock:


matyasselmeci (approved):

matyasselmeci (approved):

brianhlin:
@CannonLock at some point can we split that const up into a site-per-line? It'll make diffs easier to parse if we add/remove sites

jasoncpatton:
> @CannonLock at some point can we split that const up into a site-per-line? It'll make diffs easier to parse if we add/remove sites

(And sorted!)

PR #2989, opened 2023-02-28
Map Update

CannonLock:
- Resource Level Displays for OSPool, CC*, OSDF, Compute and Storage
- Fix bug that appends #close to url when consecutive icons are clicked
- Updated the Icons
- Added number values for Sites and Institutions


CannonLock:
OSPool sites goes down by two for this due to the resource level generation discovering that the Site was active for that contributing resource, but not that resource.

djw8605 (approved):
LGTM

PR #2990, opened 2023-02-28
Add OSDF origin server to SDSC-NSG

mkandes:
NSG would like to evaluate the capabilities provided by running its own origin server within the OSDF.

PR #2993, opened 2023-02-28
Registering Neuroscience Gateway (NSG) as a VO

mkandes:
NSG at SDSC is registering as a VO to setup an OSDF origin server.


matyasselmeci:
Merging this as-is; after #2990 is merged you can uncomment the AllowedOrigins list.

PR #2994, opened 2023-03-01
OSDF namespaces JSON: exclude caches that are down or inactive

matyasselmeci:
SOFTWARE-5440 and SOFTWARE-5471

### DIFF
--- a/src/stashcache.py
+++ b/src/stashcache.py
@@ -543,12 +543,15 @@ def get_namespaces_info(global_data: GlobalData) -> PreJSON:
         }
 
         for cache_name, cache_resource_obj in cache_resource_objs.items():
-            if resource_allows_namespace(cache_resource_obj, ns) and namespace_allows_cache_resource(ns, cache_resource_obj):
+            if resource_allows_namespace(cache_resource_obj, ns) and namespace_allows_cache_resource(ns,
+                                                                                                     cache_resource_obj):
                 nsdict["caches"].append(cache_resource_dicts[cache_name])
         return nsdict
+
     # End helper functions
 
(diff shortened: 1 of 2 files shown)


src/stashcache.py
```diff
@@ -543,20 +543,38 @@ def _namespace_dict(ns: Namespace):
...
+    def _resource_has_downed_cache(r: Resource, t: Topology):
+        if r.name not in t.present_downtimes_by_resource:
+            return False
+        downtimes = t.present_downtimes_by_resource[r.name]
+        for dt in downtimes:
+            try:
+                if XROOTD_CACHE_SERVER in dt.service_names:
+                    return True
+            except (KeyError, AttributeError):
+                continue
```
brianhlin, line 560: Too bad we don't have a filter for this :(

brianhlin (approved):
One question otherwise LGTM

src/stashcache.py
matyasselmeci: A lot of these queries would be soooo much easier if I could do SQL queries on Topology...

src/stashcache.py
```diff
@@ -543,20 +543,26 @@ def _namespace_dict(ns: Namespace):
         }
 
         for cache_name, cache_resource_obj in cache_resource_objs.items():
-            if resource_allows_namespace(cache_resource_obj, ns) and namespace_allows_cache_resource(ns, cache_resource_obj):
+            if resource_allows_namespace(cache_resource_obj, ns) and namespace_allows_cache_resource(ns,
+                                                                                                     cache_resource_obj):
```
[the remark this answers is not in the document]
matyasselmeci: Nope. Fixed.

PR #3009, opened 2023-03-07
SPRACE downtime due to power outage on campus

matyasselmeci:
https://support.opensciencegrid.org/a/tickets/72319


matyasselmeci:
augh, too slow! #3008 beat me to it

PR #3011, opened 2023-03-08
replacing USCMS pilot credentials

ddavila0:


ddavila0:
The old client ID was removed from CMS's IAM by mistake, I had to create a new client and hence getting a new ID.
I tested the new ID by manually adding the mapping  in UCSD's CEs and it worked fine

matyasselmeci (approved):

matyasselmeci:
@brianhlin will we require a new release of osg-scitokens-config?

PR #3041, opened 2023-03-15
Move data federation related code from vos_data.py to data_federation.py

matyasselmeci:
This is just an organizational thing. `vos_data.py` was mostly meant for generating the [vosummary XML](https://topology.opensciencegrid.org/vosummary/xml). The data federation stuff (which I am planning to add more to) is fairly unrelated to that, and it was starting to clutter up the file.

Aside from imports, no actual code has been changed.


CannonLock (approved):
LGTM. I'd just double check that one import.

PR #3045, opened 2023-03-20
Add CredentialGeneration block for the /ospool/protected namespace (SOFTWARE-5381)

matyasselmeci:


matyasselmeci:
I'll merge this because it's harmless; we can revise it later.

bbockelm:
For reference, the schema here looks good!

PR #3046, opened 2023-03-20
Allow and prefer spaceless versions of the SciTokens auth type keys

matyasselmeci:
"BasePath" instead of "Base Path", "RestrictedPath" instead of "Restricted Path", "MapSubject" instead of "Map Subject" This makes those attributes more consistent with the attribute names everywhere else in the file. The old names are still supported.


brianhlin (approved):
We may as well update the data after this is deployed, too.

PR #3062, opened 2023-03-30
Update SPRACE_downtime.yaml

jmarra13:
Network outage


brianhlin:
Thanks for the downtime PRs @jmarra13 ! One thing to note is that we have a separate form for resource group downtime generation (https://topology.opensciencegrid.org/generate_resource_group_downtime), which would save you from having to open multiple separate PRs in the future!

jmarra13:
Thanks Brian that's really good to know, saving It in my notes.

PR #3067, opened 2023-04-04
Add HEPScore23Percentage to WLCGInformation

djw8605:
HEPScore23Percentage is the percentage of the resource that has been benchmarked with HEPScore23 benchmark, the new benchmark for HEP.  Initially, it would be 0.  Sites will set this to the percent of their resources benchmarked with the new benchmark.   Since the old benchmark (HEPSPEC) and the new (HEPScore23) have a 1 to 1 normalization factor, no changes to the APELNormalFactor are required.

How the percentage of a resource is defined, it left as an exercise for the reader.  Percent of Cores?  Percent of job slots?  Percent of Memory?

### DIFF
--- a/src/schema/rgsummary.xsd
+++ b/src/schema/rgsummary.xsd
@@ -151,6 +151,7 @@
                                 <xsd:element name="StorageCapacityMax" type="xsd:string"/>
                                 <xsd:element name="HEPSPEC" type="xsd:string" />
                                 <xsd:element minOccurs="0" name="APELNormalFactor" type="xsd:string" />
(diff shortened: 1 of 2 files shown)


brianhlin:
Can you also add it to the template? https://github.com/opensciencegrid/topology/blob/master/template-resourcegroup.yaml#L124

djw8605:
Added to template.

brianhlin (approved):
LGTM, I think. I'm not aware of any requirement to change Topology itself so it'd be good to test that the XML actually gets updated when a site adds it to their YAML

PR #3069, opened 2023-04-04
Add downtime for USCMS-FNAL-WC1-CE3

farrukh-aftab-khan:
Hardware is being replaced on April 10. Two day buffer before April 10 to drain the CE properly and one day afterwards for unexpected suprises


brianhlin (approved):
Verified that this is Farrukh. LGTM

PR #3074, opened 2023-04-05
Remove "ligo" token issuers from /user/ligo OSDF

josh-willis:
As of yesterday, both https:/test.cilogon.org/ligo and https://cilogon.org/ligo have been retired as issuers for tokens for our users, in favor of the "igwn" issuer only. This PR therefore removes those.  In addition, we are removing the scitokens.org issuer because we do not believe it is needed for accessing files via OSDF or CVMFS.  However we do still need to keep it in the Credentials section of this file, as our pilots use that issuer.


brianhlin (approved):
LGTM

PR #3077, opened 2023-04-07
Remove PATh Facility access to OSPool namespaces (INF-406)

brianhlin:
@rynge this will revoke access to `/ospool/PROTECTED` from AP1 users. I know that we have some users that need to move data between the two namespaces so we should probably hold off on merging this until that's done.


rynge:
We did notify the users March 29th. I think we are good to merge.

brianhlin:
Alright, let's do it. Merging

PR #3079, opened 2023-04-07
Add PyYaml module and function that reads services.yaml.

jeff-takaki:
This commit includes a new function that reads and parses services.yaml within the topology directory and returns the dictionary. It is intended to replace the hardcoded list of SERVICE_IDS filters and will be used with osg-notify.


matyasselmeci:
Doesn't look like we're planning to use this. Sorry; closing.

PR #3087, opened 2023-04-12
adding an outage for USCMS-FNAL-WC1-CE3

farrukh-aftab-khan:
CE came out of the downtime and started failing pilots immediately. Need some time to debug the cause.


bbockelm (approved):
@farrukh-aftab-khan / @brianhlin - this reminds me, Farrukh is not yet properly associated with FNAL.

However, given Brian's comment on #3069, I'm going to approve here.

PR #3088, opened 2023-04-13
Add ospool ap20, 21, 22 osdf paths

djw8605:


brianhlin (changes requested):
The `base_path` for all the `https://osg-htc.org/ospool` issuers need to be updated to include each new basepath + everything on this line https://github.com/opensciencegrid/topology/pull/3088/files#diff-6c9147702941595ea238999fbc3c4e5f8870dad0f29097490dc085b9b8a8cd20R158 since `xrootd-scitokens` only respects the last `[Issuer]` block

brianhlin (approved):
Oops, I misread this earlier. The `Authorizations.SciTokens.Issuer` is unique for each path so we don't need to muck with base paths as I mentioned above.

PR #3091, opened 2023-04-14
Extend credential generation scheme for base_path and issuer

bbockelm:
The credential generation information was missing:
- `base_path`: for calculation of requested scopes using the OAuth2 scheme.
- `vault_issuer`: Vault servers have a separate "issuer" concept to specify which token to request; this is in use by at least IGWN.


src/README.md
```diff
@@ -339,17 +339,22 @@ DirList is optional.
 CredentialGeneration:
   Strategy: "Vault" or "OAuth2"
   Issuer: "<ISSUER URL>"
+  BasePath: "<PATH>"
   MaxScopeDepth: <INTEGER>
   VaultServer: "<HOST>:<PORT>"
```
matyasselmeci, line 344:
```suggestion
  VaultServer: "<HOST>:<PORT>"
  VaultIssuer: "<ISSUER STRING>"
```

matyasselmeci (approved):
A small doc tweak but otherwise it's fine. Pre-approving.

PR #3104, opened 2023-04-20
Use new jquery

matyasselmeci:


CannonLock (approved):
Looks like only the downtime forms use it, and those function as expected so I am happy.

PR #3107, opened 2023-04-21
Fix type issue when adding to the warning list

matyasselmeci:
TIL that Python will happily let you `+=` a string to a list, adding each individual character as a separate list item. 🤦🏻


CannonLock (approved):
LGTM, ran into a similar bug somewhere else this week so maybe it is the time of the year.

PR #3117, opened 2023-05-04
Make HEPScore23Percentage a float

djw8605:


matyasselmeci (approved):
Sure. All our existing data passes.

PR #3121, opened 2023-05-05
topology reg for USD Hosted CE

jdost321:


brianhlin:
Hrm, I'm actually not sure what we should do here since this is a collision of the "University of South Dakota" site across the GPN and USD facilities. The former has an active (?) backfill container setup and this is for a the Hosted CE. Kinda feels like we should allow name collisions on sites to allow for shared accounting across facilities? Feels dangerous, though.

brianhlin:
@djw8605 @matyasselmeci Hrm, I'm actually not sure what we should do here since this is a collision of the "University of South Dakota" site across the GPN and USD facilities. The former has an active (?) backfill container setup and this is for a the Hosted CE. Kinda feels like we should allow name collisions on sites to allow for shared accounting across facilities? Feels dangerous, though.

matyasselmeci:
I don't want to go there, let's just come up with a different name for the site. "University of South Dakota Campus"? There's probably a better one.

brianhlin:
> I don't want to go there, let's just come up with a different name for the site. "University of South Dakota Campus"? There's probably a better one.

We have our hands tied here a bit as the site name affects GRACC dashboard drop downs. If USD wants to see their contributions, they'll have to select USD + whatever we pick. The UX is painful here but I don't know if there's any happy middle ground.

matyasselmeci:
We've had the "no duplicate site" restriction since the beginning of Topology. I have no idea what the downstream consequences of violating that restriction are, but IMO they are not worth it to keep one site from having to check _two_ boxes instead of _one_ when looking at their dashboard.

brianhlin:
Alright, fair. @jdost321 please pick a new site name. Maybe something that reflects their unit like `USD - Research Computing` or something

jdost321:
@brianhlin @matyasselmeci I gave it a new site name, can it be merged now?

matyasselmeci (approved):
LGTM

PR #3123, opened 2023-05-05
Add HEPScore23Percentage to WLCG ordering

djw8605:


matyasselmeci:
Looks like the schema doesn't like it when the value is blank (because the empty string is not a valid float).

src/webapp/topology.py
```diff
@@ -349,7 +349,7 @@ def _expand_wlcginformation(wlcg: Dict) -> OrderedDict:
 
         new_wlcg = OrderedDict.fromkeys(["InteropBDII", "LDAPURL", "InteropMonitoring", "InteropAccounting",
                                          "AccountingName", "KSI2KMin", "KSI2KMax", "StorageCapacityMin",
-                                         "StorageCapacityMax", "HEPSPEC", "APELNormalFactor", "TapeCapacity"])
+                                         "StorageCapacityMax", "HEPSPEC", "APELNormalFactor", "HEPScore23Percentage", "TapeCapacity"])
         new_wlcg.update(defaults)
         new_wlcg.update(wlcg)
         return new_wlcg
```
matyasselmeci, line 355:
```suggestion
        if new_wlcg["HEPScore23Percentage"] is None:
            del new_wlcg["HEPScore23Percentage"]
        return new_wlcg
```

matyasselmeci (approved):

PR #3132, opened 2023-05-11
gzip json responses if headers say its okay

CannonLock:
Checks for accept-encoding:gzip and if it is there then gzip json responses to save some bandwidth

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -888,9 +888,9 @@ def make_json_response(json_response: dict) -> flask.Response:
     """Trys to gzip json response"""
 
     if 'gzip' in flask.request.headers.get("accept-encoding", ""):
-        content = gzip.compress(json.dumps(json_response).encode('utf8'), 5)
+        content = gzip.compress(json.dumps(json_response).encode('utf8', errors='replace'), 5)
         response = flask.make_response(content)
-        response.headers['Content-length'] = len(content)
+        response.headers['Content-Length'] = len(content)
         response.headers['Content-Encoding'] = 'gzip'
         response.headers['Content-Type'] = "application/json"
         return response
--- a/src/app.py
+++ b/src/app.py
@@ -12,6 +12,8 @@ import re
 import sys
 import traceback
 import urllib.parse
+import gzip
+import json
 
 from webapp import default_config
 from webapp.common import readfile, to_xml_bytes, to_json_bytes, Filters, support_cors, simplify_attr_list, is_null, escape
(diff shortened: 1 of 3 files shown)


src/app.py
```diff
@@ -884,6 +884,20 @@ def _get_authorized():
     # If it gets here, then it is not authorized
     return default_authorized
 
+def make_json_response(json_response: dict) -> flask.Response:
+    """Trys to gzip json response"""
+
+    if 'gzip' in flask.request.headers.get("accept-encoding", ""):
+        content = gzip.compress(json.dumps(json_response).encode('utf8'), 5)
+        response = flask.make_response(content)
+        response.headers['Content-length'] = len(content)
```
matyasselmeci, line 893:
```suggestion
        content = gzip.compress(json.dumps(json_response).encode('utf8', errors='replace'), 5)
        response = flask.make_response(content)
        response.headers['Content-Length'] = len(content)
```

src/tests/test_api.py
```diff
@@ -1,7 +1,10 @@
 import re
-import flask
-import pytest
+import gzip
+import json
 import urllib.parse
```
matyasselmeci, line 4:
```suggestion
from typing import Generator
import urllib.parse
```

src/tests/test_api.py
```diff
@@ -59,25 +62,36 @@
...
+    '/miscresource/json',
+    '/vosummary/json',
+    '/resources/stashcache-files',
+    '/stashcache/namespaces'
+]
+
 
 @pytest.fixture
-def client():
+def client() -> FlaskClient:
```
matyasselmeci, line 78:
```suggestion
def client() -> Generator[FlaskClient]:
```
since you're using a `yield` statement

matyasselmeci (approved):
Some tweaks but otherwise LGTM. Pre-approving.

CannonLock:
@djw8605 Same. That said its not often I get to implement 10x performance upgrades so I am happy they give me the opportunity. :)

src/tests/test_api.py
```diff
@@ -59,25 +63,36 @@
...
+    '/miscresource/json',
+    '/vosummary/json',
+    '/resources/stashcache-files',
+    '/stashcache/namespaces'
+]
+
 
 @pytest.fixture
-def client():
+def client() -> Generator[FlaskClient]:
```
matyasselmeci, line 79:
```suggestion
def client() -> Generator[FlaskClient, None, None]:
```
Sorry, apparently the syntax for generators is
```
Generator[yield type, send type, return type]
```

jthiltges:
This might be an alternative method: https://github.com/opensciencegrid/topology/commit/ba9b9fe6b2ab7156e46b3beee9939b6c12e1f709

CannonLock:
@jthiltges Does that zip everything, header independent? I have some things that will bork if they get back a zipped response.

jthiltges:
> @jthiltges Does that zip everything, header independent? I have some things that will bork if they get back a zipped response.

Good question! Quickly glancing at the docs, I'd say yes: [AddOutputFilterByType](https://httpd.apache.org/docs/2.4/mod/mod_filter.html#addoutputfilterbytype) uses [mod_deflate](https://httpd.apache.org/docs/2.4/mod/mod_deflate.html#enable), which sounds like it does the right thing in respecting accept-encoding (unless forced otherwise).

CannonLock:
Closing in favor of doing this on the apache level -> #3133

PR #3133, opened 2023-05-12
Add Cache-Control headers and enable Apache in-memory caching of topology output, along with compression

jthiltges:
This takes a neglected work-in-progress to use Varnish, and simplifies it. This builds on the ideas started in #3132.

Varnish would be a better choice for caching, but adds more complexity to keep gridsite authentication working. Unfortunately, Apache doesn't support `Cache-Control: stale-while-revalidate=...`. Some clients will still get slow responses as content is refreshed, but it's slightly better.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN alternatives --set python3 /usr/bin/python3.9
 WORKDIR /app
 
 # Install application dependencies
-COPY requirements-apache.txt src/ ./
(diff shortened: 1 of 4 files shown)


CannonLock:
LGTM! I did a quick check that deflate would respect encoding headers and everything looks good. Love the added bonus of some caching.

CannonLock (approved):

PR #3140, opened 2023-05-17
CEPH (storage) node maintenance

andres-moya:
One of the CEPH storage node refuses to boot correctly on last power outage. 
We plan to investigate and probably rebuild the node. We expect relatively short downtime but declaring entire day for contingency.

### DIFF
--- a/topology/California Institute of Technology/Caltech CMS Tier2/CIT_CMS_T2_downtime.yaml
+++ b/topology/California Institute of Technology/Caltech CMS Tier2/CIT_CMS_T2_downtime.yaml
@@ -1690,3 +1690,85 @@
   ResourceName: CIT_CMS_T2_XRD_Ceph
   Services:
     - XRootD component
+
+
+# ----------------------------------------------------------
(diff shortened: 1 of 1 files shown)


andres-moya:
Hello... i just seen another push once we was doing update. 
https://github.com/opensciencegrid/topology/commit/bbe4193a9c952982a5c123512aa73f84dc2d9b95

It looks like some of us using wrong tool to generate downtime id

We use bin/next_downtime_id:
```
#!/bin/bash
set -o errexit
set -o pipefail
cd "$(dirname "$0")"/..

unset GREP_OPTIONS
egrep -h '^ *ID: *[0-9]{1,9}\>' topology/*/*/*_downtime.yaml |
  awk '$2 > max {max = $2}; END {print max + 1}'
```

The script is looking for from 1 to 9 numeric. While our ID is 9 digits. ATLAS ID is longer. So that our IDs belong to "parallel" universes?  Should we update script to use  ^ *ID: *[0-9]{1,16}\>

matyasselmeci (approved):

PR #3142, opened 2023-05-17
Modify next_downtime_id to deal with bigger (>9 digit) downtime IDs, …

matyasselmeci:
…since the timestamp-based ones from the webform are now larger than that.


matyasselmeci:
Noticed by @andres-moya in #3140.

brianhlin (approved):
lol oof. LGTM

PR #3151, opened 2023-05-23
Update Lehigh - Hawk.yaml

ultramathman:
Adding new XRootD cache server to Lehigh - Hawk.


matyasselmeci:
Do you have a host certificate for the new cache server? The DN needs to be added to the registration (example: https://github.com/opensciencegrid/topology/blob/master/topology/Internet2/Internet2GreatPlains/I2GreatPlainsInfrastructure.yaml#L21)

ultramathman:
Ah, I thought that was only required when caching protected data. I'll be able to generate and add one once the firewall holes for this host are in place, either tomorrow or early next week. Thanks.

ultramathman:
I believe I added the DN for the Let's Encrypt host certificate.

matyasselmeci (approved):

PR #3157, opened 2023-05-24
Upgrade all the things

matyasselmeci:
Upgrade all dependencies to the latest; use the `~=` operator (https://peps.python.org/pep-0440/#compatible-release) so we don't completely lock versions in. Update the downtime forms for WTForms 3.0 compatibility.


matyasselmeci:
this apparently causes booleans in XML to be rendered as `true` and `false`, not `True` and `False` which breaks our XML schema.

matyasselmeci:
I updated the schema and gave the CRIC folks a heads up. Can I get an approval?

brianhlin (approved):
LGTM

PR #3159, opened 2023-05-25
Add referenced requirements file to Dockerfile

CannonLock:
Add referenced requirements file into Dockerfile.

https://github.com/opensciencegrid/topology/actions/runs/5082972128/jobs/9133369816#step:7:506


CannonLock:
@matyasselmeci I can't think of when I would ever use a requirements reference again but this is a really nice feature. 

Love a good reference.

matyasselmeci (approved):

PR #3171, opened 2023-05-31
Add BasePath to credential generation info for the /chtc namespace

matyasselmeci:


brianhlin (approved):
This looks reasonable to me, though IDK if I fully understand the change.

PR #3184, opened 2023-06-02
Fix ports on new UChicago origin

matyasselmeci:
@LincolnBryant 
The xrootd instance that serves authenticated data listens on port 1095; the instance that serves unauth data on port 1094.


LincolnBryant (approved):
Thanks for catching that!

PR #3197, opened 2023-06-09
Create aionedge.yaml

shreyb:
This is a request to add an aionedge project to OSG topology so that a CVMFS repository can be created for aionedge.  This will be used by DUNE and SBND, Fermilab experiments.

### DIFF
--- a/projects/aionedge.yaml
+++ b/projects/aionedge.yaml
@@ -1,0 +1,7 @@
+Department: N/A
+Description: 'Explore approaches of implementing OnEdge Real-Time AI-based algorithms
+  in identifying particle interaction in liquid Argon using GPU, FPGA or low-power
+  edge computing devices in the Warm Electronics Interface Boards (WIB) data stream. '
+FieldOfScience: High Energy Physics
+Organization: Fermilab
+PIName: Shekhar Mishra


matyasselmeci:
As far as I'm aware, projects are not used to create areas for CVMFS; that information is in the "virtual-organizations" directory.

matyasselmeci:
Yes, that should work. @ppaschos, have you been working with them? Do this sound good to you?

ppaschos:
These are in Ken Herner and Steven Timm's court

ppaschos:
If this is a repository to will be used by mainly fermilab experiments, can't it not be under the fermilab VO?

ppaschos:
yes that's exactly what I meant. FNAL being the parent  VO means that you don't have to assign/create a separate one.. in my opinion unless there are other 'political' issues here.

shreyb:
Let me just ask Ken to make sure, but I think that'll be OK.  Thanks very much @ppaschos  and @matyasselmeci  for walking me through this.

@kherner:  are you OK with us modeling the `aionedge` entry after the `LArSoft` entry here?  Since AIOnEdge isn't actually a VO, per se, we'd have Fermilab be the parent VO, like it is in the LArSoft entry (https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/LArSoft.yaml).  I think as long as this still allows AIOnEdge to have the repo name `aionedge.opensciencegrid.org`, the requestors/stakeholders don't really care how we have the OSG end set up.

kherner:
Yes, let's just put it under the Fermilab VO. After all several of the FIFE experiments already are.

shreyb:
Great - then I'm good to close this PR out and open another one under `virtual-organizations`.

matyasselmeci:
I merged #3211; we no longer need this PR open.

PR #3202, opened 2023-06-14
Fix GHA deprecation warnings (Software-5354)

jtakaki-matc:
Fixes set-output warnings for build-client-container.yml and build-sw-container.yml


brianhlin (approved):
LGTM

PR #3206, opened 2023-06-15
Adding token issuer

biozit:
Adding token issuer


matyasselmeci (approved):

djw8605:
This issuer won't work.  It needs to be a proper website, like a github pages page.

PR #3211, opened 2023-06-16
Create AIOnEdge.yaml

shreyb:
As suggested in #3197,  we are opening this request under the virtual-organizations directory.

This is a request to add an aionedge virtual organization to OSG topology so that a CVMFS repository can be created for aionedge. This will be used by DUNE and SBND, Fermilab experiments.

As mentioned in the instructions, I left any ID fields blank for which I could not find the ID.

### DIFF
--- a/virtual-organizations/AIOnEdge.yaml
+++ b/virtual-organizations/AIOnEdge.yaml
@@ -1,0 +1,33 @@
+Active: True
+AppDescription: Explore approaches of implementing OnEdge Real-Time AI-based algorithms
+  in identifying particle interaction in liquid Argon using GPU, FPGA or low-power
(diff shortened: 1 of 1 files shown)


matyasselmeci:
I added an ID for the VO; can Shekhar register in COManage so I can add an ID for him? There are instructions here: https://osg-htc.org/docs/common/contact-registration/

shreyb:
I've sent him a message; I'll get back to you when he responds.  Fermilab has a holiday for Juneteenth, so it's possible we won't hear back from him until Tuesday at the earliest (unless he goes ahead and does it over the weekend).

matyasselmeci (approved):

PR #3200, opened 2023-06-13
Update bitbucket SSH host keys

brianhlin:
https://bitbucket.org/blog/ssh-host-key-changes

PR #3234, opened 2023-06-27
Replace bitbucket.org host key

matyasselmeci:
bitbucket.org changed their RSA SSH host key; here is their ECDSA one instead.


matyasselmeci:
Never mind, #3200 did this

PR #3237, opened 2023-06-28
Add UWMadison_Pool Project

cjabplanalp:


matyasselmeci (approved):

matyasselmeci:
I'll have to debug why osg-bot still doesn't recognize you. I'll let you know if any action needs to be taken on your part but the problem is probably on our end.

brianhlin:
@matyasselmeci I'm going to guess it's because Cameron wasn't in the OSG Topology Contacts COU. I gave him that role in COManage so 🤞🏽 that it works next time around.

PR #3246, opened 2023-07-05
Add SBU -> "State University of New York at Stony Brook" mapping

matyasselmeci:
First used in #3243; we have other SUNY at Stony Brook projects but they don't use the Institution_PI naming convention.


brianhlin (approved):
LGTM

PR #3273, opened 2023-07-13
Adding EOS SE

smithnp:
Went to create a scheduled downtime for next week and noticed that I could not select our new SE as a thing to register downtime for.


smithnp:
Brian, the downtime was yesterday.  I believe Stefan was able to find a place or at least talk to some people to get our downtime changed from being listed as "partial" to "full" toward the end of the week last week, so we should be fine for all of our site availability numbers.  I would say this issue is fairly non-urgent at this point.

brianhlin:
@matyasselmeci using my powers of delegation: can you reach out to the CRIC folks to coordinate addition of an EOS service to Topology?

matyasselmeci:
Sure. https://opensciencegrid.atlassian.net/browse/SOFTWARE-5674

matyasselmeci (approved):

PR #3287, opened 2023-07-19
Add WISC-PATH-DEV ResourceGroup with an EP, WISC-PATH-DEV-EP

matyasselmeci:
This resource group is marked as ITB; the EP points to the test PF central manager instead of production


brianhlin (approved):
LGTM

PR #3304, opened 2023-08-02
Create MOLLER.yaml

rahmans1:
Creating file for the MOLLER experiment at Jefferson Lab


MOLLER.yaml
```diff
@@ -0,0 +1,119 @@
...
+  Administrative Contact:
+  - ID: <EMAIL HASH>
+    Name: <FULL NAME>
+  # - ...
+
+  # Security Contact is a list of people to contact regarding security issues
+  Security Contact:
+  - ID: <EMAIL HASH>
+    Name: <FULL NAME>
+  # - ...
```
brianhlin, line 31: We'll need registered contacts for both of these fields. Could you have those contacts register through these instructions?

brianhlin: Hi @rahmans1, I 've approved your registration so we should be able to fill these fields in for you. However, it would be ideal if we would two or more contacts for both the administrative + security fields. Is there anyone else from MOLLER that would be able to register as a contact?

PR #3317, opened 2023-08-09
Adding the first resourcegroup yaml file to the NET2 project.

edubach:
Adding the first resourcegroup yaml file to the NET2 project.


brianhlin (approved):
LGTM

PR #3336, opened 2023-08-16
Create UNL_Hebets.yaml

showmic09:


mwestphall:
@jstathas we have projects listed under both "University of Nebraska-Lincoln" and "University of Nebraska - Lincoln". Which name should we prefer going forwards?

brianhlin:
@jstathas I'd also like to note that on the resource contributing side, there's the Facility University of Nebraska (https://github.com/opensciencegrid/topology/blob/master/topology/University of Nebraska/), which contains sites "Nebraska-Lincoln" and "Nebraska-Omaha"

jstathas:
Brian,

I think we agree that theoretically both the contributor facilities and the user institutions should have the same value.

There is no "real" institution by the name "University of Nebraska" or for that matter "University of Wisconsin". (https://ope.ed.gov/dapip/#/home)

However, given that is not the convention that has been historically used on the resource contributing side, is the directive to use the facility side naming convention/structure on the user side?

IMO, the answer should be no - the directive is to have accurate, correct institution names and avoiding a nested approach is a simpler data structure for reporting purposes.

However, if we are duplicating the data structures used on the facility side on the institution user side, then I could be persuaded to say that we should duplicate the naming convention used on the Facility side.

However, as far as I can tell they are completely independent lists and are organized differently.

I need a better understanding of what we are trying to accomplish here and an understanding of the scope.

Janet






Janet Stathas (she/her)

Project Manager

Morgridge Institute for Research

PATh<https://path-cc.io/>

Email: ***@***.***

Mobile: 608.345.9605

________________________________

bbockelm:
There's a bit of risk in renaming the facility to the proper institution name -- it's unclear what the downstream dependencies are (especially for WLCG).

That said, @djw8605 is probably the expert here.  My recollection is that all the WLCG naming is currently factored out... that said, since we need to change both UW-Madison and UNL's facility names, we should at least serialize the changes and wait ~month between the two.

djw8605:
WLCG uses the topology `Resource Group`.  So changing the facility name shouldn't matter.

mwestphall (approved):

showmic09:
So moving forward-if any new project shows up from UNL-then the name should be "University of Nebraska-Lincoln"?

PR #3350, opened 2023-08-24
access /ospool/uc-shared from ap[20-22]

LincolnBryant:
Try to update origin configs to allow users from ap[20-22] access the origin on ap23 for /ospool/uc-shared


mwestphall (approved):

matyasselmeci:
We've made some changes since then and this PR doesn't apply so I'm closing it.

PR #3352, opened 2023-08-24
Added Arizona science gateway

rynge:


topology/University of Arizona/Research Technologies/ResearchTechnologies.yaml
```diff
@@ -0,0 +1,25 @@
+GroupDescription: ResearchTechnologies
+GroupID: 1366
+Production: true
+Resources:
+  Gateway:
```
matyasselmeci, line 5:
```suggestion
  Arizona-ResearchTechnologies-Gateway:
```
Resource names need to be globally unique so you should use a less generic name.

matyasselmeci (approved):

PR #3354, opened 2023-08-24
/ospool/uc-shared/public should be public

matyasselmeci:


brianhlin (approved):
How do we allow writeback on a part of a namespace that doesn't do authz? Otherwise LGTM

matyasselmeci:
That's a good question. Since the writeback is to the authenticated xrootd instance on the origin, you'll need a token anyway and then the origin will accept whatever it accepts. But we need this namespace to be "public" in order to avoid having stashcp prompt the user for auth.

PR #3378, opened 2023-09-07
Update CIT_LIGO.yaml for staging origin

josh-willis:
This creates the entry for the staging server at the CIT LIGO cluster. The ID has been left blank to be filled in.  A separate PR will be forthcoming to add this origin as the source for a new Namespace in the LIGO VO topology file.


biozit:
Can I add this to the monitoring system?

mwestphall (approved):

PR #3385, opened 2023-09-11
Rename the UWMadison projects using new scheme

CannonLock:
UWMadison_\<project-name>.yaml


CannonLock:
As per Elastic search there are 8 chtc projects running on the OSPool 

['BMI_Gitter', 'BiochemSenes', 'GLOW', 'GeneticsPool', 'Pharmacy_Kwan', 'Psychology_Curtin', 'CHTC', 'SmallMolecule_Hoffman']

GLOW doesn't have a mapping so I never created a project and Senes was added last week and has yet to be added.

brianhlin:
Should `projects/UWMadison_GeneticsPool.yaml` be `projects/UWMadison_Genetics_Pool.yaml`? Looks like it's just `GeneticPool` in the CHTC but while we're renaming, maybe we should align the name? @ChristinaLK ?

ChristinaLK:
the webapp has both a "Project Name" and "Accounting Group" and for this project it's `GeneticsPool` for one and `Genetics_Pool` for the other.  
We mainly use the Accounting group name for reporting, but I bet the CHTCProjects attribute is pulling from Project Name. 🙃 They are mainly consistent but there are a few holdovers where they *don't* match. Might be worth figuring out which those are and manually editing.

brianhlin:
> We mainly use the Accounting group name for reporting, but I bet the CHTCProjects attribute is pulling from Project Name.

I thought that @jasoncpatton mainly used `ProjectName` for reporting

brianhlin:
@ChristinaLK we do indeed use ProjectName for reporting. What reporting are you referring to that uses `AccountingGroup`?

`CHTCProjects` is indeed populated from the user app projects.

brianhlin:
@ChristinaLK 

> I vote that we align on the project name as doing the opposite will cause long term issues with ElasticSearch, while the former will cause temporary pain with the usage reports for ~1y.

How does this sound to you?

PR #3405, opened 2023-09-25
Changing the path to configure the access

biozit:
Changing the path to configure the access


biozit:
@matyasselmeci @brianhlin Could you check this?

matyasselmeci (approved):

PR #3413, opened 2023-10-04
Update NET2.yaml

edubach:
Adding SQUID services


edubach:
I used the ID value returned by topology/bin/next_resource_id.

PR #3432, opened 2023-10-18
Add token and voms config

biozit:
Add token and voms config


matyasselmeci (approved):

matyasselmeci (approved):
This is fine; I'll make a PR to fix our test tomorrow morning

PR #3451, opened 2023-10-27
Register /ospool/ap22/.well-known public namespace

matyasselmeci:
The issuer and public key for ap22 live there and are served by a public XRootD stash-origin instance


brianhlin (changes requested):
I'd like to hold off on this for now until we understand how AP20/21 serve the same files without this config

matyasselmeci:
Now that we've figured out how things work (or don't) over Slack and Zoom, how do you feel about this PR?

matyasselmeci:
What are we going to do with this?  The APs are currently only working because of an oversight in xcache so I think this PR is closest to the correct solution.

brianhlin:
> The APs are currently only working because of an oversight in xcache so I think this PR is closest to the correct solution.

Could you remind me what this oversight is exactly?

matyasselmeci:
> Could you remind me what this oversight is exactly?

There's no systemd override for `xrootd-privileged@stash-origin` so you can write an authfile by hand and the authfile-update script failing doesn't keep the service from starting.

brianhlin (approved):
LGTM

PR #3457, opened 2023-10-30
Limit Issuer block titles to 50 characters (SOFTWARE-5659)

matyasselmeci:
Ensure the title of the 'Issuer' block is no more than 50 characters

- chop off the `https://` to give us a little more room
- use a sha1 hash if the title is still too big

### DIFF
--- a/src/tests/data/testrg.yaml
+++ b/src/tests/data/testrg.yaml
@@ -1,0 +1,246 @@
+Production: true
+SupportCenter: Self Supported
+
+GroupDescription: Test Resource Group
+
+GroupID: 91359
+
+Resources:
+  TEST_STASHCACHE_CACHE:
+    Active: true
+    Description: This is a StashCache cache server at TEST.
+    ID: 99958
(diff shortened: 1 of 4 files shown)


brianhlin (approved):
LGTM. Though I thought this was fixed in XRootD?

djw8605:
This is fixed in xrootd.  The new maximum is 1024 characters:
https://github.com/xrootd/xrootd/commit/c87bec48ca0162085942d70f6265975da8576a80

Looks like the fix has been in releases starting in 5.6.2

matyasselmeci:
Yeah this is more for compat with earlier xrootd versions in case people need to downgrade because of an issue or something.

brianhlin:
I think we should just abandon this and direct people to upgrade to 5.6.3 if they have any issues

matyasselmeci:
Mmm. OK. I'm going to keep some of the test scaffolding because I'm using it in #3463; maybe split it out into a separate PR.

PR #3463, opened 2023-10-31
Wip/itb public read auth write

matyasselmeci:


matyasselmeci:
Merged into itb so I can try it out on topology-itb

PR #3474, opened 2023-11-07
Add CE for hypatia cluster at AEI

coldbeer101:
Needs IDs created. Thanks!


matyasselmeci (approved):
LGTM. Good news: you don't need those IDs anymore. I'll update the templates.

PR #3476, opened 2023-11-08
Additional tests for stash/osdf scitokens.conf generation

matyasselmeci:
These are the tests pulled out of #3457 .  If it's approved, I will squash.


brianhlin (approved):
LGTM

PR #3478, opened 2023-11-08
Retiring brown-osg CE

smithnp:
We are retiring the brown cluster, so I have set the brown-osg-ce active status to false.  Also I noticed hadoop-osg was still set active, it has been retired for some time now, I have also set it to false.


brianhlin (approved):
LGTM

PR #3488, opened 2023-11-13
Add /chtc/PROTECTED/sc-origin2000, a namespace served by sc-origin2000.chtc.wisc.edu

matyasselmeci:
I would like to serve some protected data off of sc-origin2000 (a non-container instance) to test systemd service files.


mwestphall (approved):
LGTM

PR #3490, opened 2023-11-14
Updated UERJ_downtime.yaml

oracle7:


matyasselmeci:
Looks like this downtime was added in #3492 so I will close this PR. We'll take care of adding you to the relevant resources for next time.

PR #3500, opened 2023-11-23
Update Project FieldOfScience

CannonLock:
- Aligns all project FieldOfScience to Valid values and defined list
- Completed with script here: https://github.com/CannonLock/general/blob/master/topology_projects_that_comply/applying_changes/main.ipynb


djw8605 (commented):
Just a quick comment.

PR #3504, opened 2023-11-27
Tag ap1 for the OSPool

rynge:
Allow some jobs from ap1 to flow to the OSPool


brianhlin (approved):
LGTM. This may just be temporary for a demo.

PR #3513, opened 2023-12-06
Update OSG.yaml

jlstephen:
Update /ospool/ap22/data SciTokens Base Path to match the configuration for AP20 and AP21


brianhlin (approved):
LGTM

PR #3518, opened 2023-12-10
Disable non-functional shoveler cache

bbockelm:
The fact this is enabled is causing problems with the directors as it is assumed caches registered in topology are usable.

@matyasselmeci


bbockelm:
Self-merging as this is causing a ruckus -- a lot of requests are going the way of this cache!  It's worse this weekend because an IPv6 address for the host has appeared but there's no IPv6 connectivity.

PR #3519, opened 2023-12-12
Add KAGRA OSDF Origin

brianhlin:
It's not 100% clear that we have the right FQDN here but we'll follow up with the admin and update later as appropriate.


brianhlin:
With remote queries, it /looks/ like the FQDN is correct

mwestphall (approved):
LGTM

PR #3529, opened 2023-12-18
Fix issue preventing COManage Topology contact sync (SOFTWARE-5766)

brianhlin:
Updated the search base:

1. Reduce the search base to only CO Persons since we only need the voPersonID (associated with the CO Person)
2. Searching without an `ou` just gives a list of `ou`'s now

Then we can further filter on the active Topology/OASIS Manager contacts

Untested in Topology but this returns 99 entries when doing manual CLI searches

### DIFF
--- a/src/webapp/ldap_data.py
+++ b/src/webapp/ldap_data.py
@@ -19,7 +19,13 @@ def get_contact_cilogon_id_map(global_data):
 # cilogon ldap query constants
 #_ldap_url = "ldaps://ldap.cilogon.org"
 #_username = "uid=readonly_user,ou=system,o=OSG,o=CO,dc=cilogon,dc=org"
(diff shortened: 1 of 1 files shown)


brianhlin:
@williamnswanson how does this look to you?

williamnswanson:
Which change causes `Searching without an ou just gives a list of ou's now`? Looks good for the other changes though.

brianhlin:
> Which change causes Searching without an ou just gives a list of ou's now? 

I just saw this when running `ldapsearch` manually from the command line when using previous base search, i.e. without the `ou=people`. That being said, we were asking for a `subtree` scope so that probably got _everything_ out of LDAP

matyasselmeci (approved):
sure, let's give it a shot.

PR #3542, opened 2023-12-21
Add test origins to LIGO.yaml

josh-willis:
This PR integrates two new test origins---`CIT_LIGO_ORIGIN_TEST` and `CIT_LIGO_ORIGIN_TEST_WRITE`---into the OSDF for the LIGO VO. The former is to test a combination of vault-issued tokens from CILogon and AP-issued tokens in a standard xrootd origin; the second is to test the same mixture in writeback mode to an origin with xrootd-privileged.

A couple of notes:
1. It is deliberate that we need two new base-paths, `/igwn/test` and `/igwn/test-write`. We do this so that we can use the existing scopes of our CILogon tokens, and therefore are testing with tokens that match what users will get (whether they use CILogon-issued tokens or AP-issued tokens). This has the side-effect that we have to add those two new paths to the `Base Path` for all of our Namespaces, since those are grouped together by token issuer. Hence the other changed lines in this file.
2. The new AP-issuer is for now only added to these test origins, since part of what we want to test is that this doesn't break anything before extending it to production origins. Therefore that issuer only supports the two new `Base Path`s for now.

### DIFF
--- a/virtual-organizations/LIGO.yaml
+++ b/virtual-organizations/LIGO.yaml
@@ -86,10 +86,10 @@ DataFederations:
           - FQAN: /virgo/ligo
           - SciTokens:
               Issuer: https://cilogon.org/igwn
-              Base Path: /user/ligo,/igwn,/igwn/cit
+              Base Path: /user/ligo,/igwn,/igwn/cit,/igwn/test,/igwn/test-write
           - SciTokens:
               Issuer: https://test.cilogon.org/igwn
-              Base Path: /user/ligo,/igwn,/igwn/cit
+              Base Path: /user/ligo,/igwn,/igwn/cit,/igwn/test,/igwn/test-write
         AllowedOrigins:
           - CIT_LIGO_ORIGIN
           - LIGO-StashCache-Origin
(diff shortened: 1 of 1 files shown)


josh-willis:
Hmm, I don't understand the assertion error in the test failure; can someone from PATh advise me on what might be wrong with this PR that it causes this failure?

virtual-organizations/LIGO.yaml
```diff
@@ -359,6 +359,107 @@ DataFederations:
...
         Writeback: https://origin-staging.ligo.caltech.edu:1095
+      - Path: /igwn/test
+        Authorizations:
+          - FQAN: /osg/ligo
+          - FQAN: /virgo
+          - FQAN: /virgo/virgo
+          - FQAN: /virgo/ligo
+          # No Scitokens issuers because they are duplicated with the /user/ligo path.
+          # As long as the caches are kept in sync, this below fix should be fine.
+          # But the problem should be solved either by topology (ideally) or the SciTokens XRootD plugin.
```
matyasselmeci, line 370: Is this comment accurate? I see 3 Scitokens issuers below...

josh-willis: No, I think that got grabbed from copy-and-paste. It's also above for all of our other origins I think. Maybe it was a different attempt to get around needing all base paths that an issuer would support? It looks like it first appears with the `/igwn/virgo` namespace, which also has (two) SciTokens issuers.

matyasselmeci (approved):
LGTM; we can get rid of the comments later. You can ignore the test failures: we've got some special code for handling LIGO namespaces so one of our tests is sensitive to how many there are. I'll fix the test after merging this PR.

PR #3548, opened 2023-12-29
Be lenient in Authfile generation (SOFTWARE-5787)

matyasselmeci:
Instead of failing (4xx error code), just return an explanatory comment when asking for an Authfile in the following cases:

- public instance Authfile for an origin that supports only protected namespaces
- auth instance Authfile for an origin that supports only public namespaces
- public instance Authfile for a cache that supports only protected namespaces
- auth instance Authfile for a cache that supports only public namespaces

Continue to fail when requesting any instance Authfile for a cache or origin that supports no namespaces at all


mwestphall (approved):
LGTM

PR #3559, opened 2024-01-06
Add Institution/ID mapping (SOFTWARE-5786)

matyasselmeci:
See [ticket](https://opensciencegrid.atlassian.net/browse/SOFTWARE-5786) for details. In addition to the endpoint, there is a test for duplicate names and osg_ids.


brianhlin (approved):
LGTM. And it looks like it wouldn't be too bad to replace the mapping file once @mwestphall  institution DB work down the line

PR #3590, opened 2024-01-17
Enabled multiple contact types and added "local" types

osg-cat:
This change affects the `--oim-contact-type` command-line option. Now it can be specified more than once, and all matching contacts are used; also, I added the new “Local Operational” and “Local Security” contact types.

I tested this change with:

- No contact type argument (= all)
- The “all” contact type
- The “security” contact type
- The “security” and “local operational” contact types

### DIFF
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -133,7 +133,7 @@ def parseargs():
         oparser.error("--oim-owner-filter and --oim-recipients=vos options are conflicting")
     if args.name_filter and args.fqdn_filter:
         oparser.error("Can't specify both --oim-name-filter and --oim-fqdn-filter")
-    if (len(args.contact_type) == 0) or ("all" in args.contact_type):
+    if not args.contact_type or ("all" in args.contact_type):
         args.contact_type = ["all"]
 
     if args.from_name == 'security':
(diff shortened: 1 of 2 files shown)


bin/osg-notify
```diff
@@ -116,8 +116,9 @@ def parseargs():
     oparser.add_argument("--oim-owner-filter", dest="owner_vo",
                          help="Filter on resources that list VO(s) as a partial owner")
 
-    oparser.add_argument("--oim-contact-type", default="all", dest="contact_type",
-                         choices=["all", "administrative", "miscellaneous", "security", "submitter", "site"],
+    oparser.add_argument("--oim-contact-type", action="append", default=[], dest="contact_type",
```
brianhlin, line 119:
```suggestion
    oparser.add_argument("--oim-contact-type", action="append", default=set(), dest="contact_type",
```

To avoid duplicates

bin/osg-notify
```diff
@@ -131,6 +132,8 @@ def parseargs():
         oparser.error("--oim-owner-filter and --oim-recipients=vos options are conflicting")
     if args.name_filter and args.fqdn_filter:
         oparser.error("Can't specify both --oim-name-filter and --oim-fqdn-filter")
+    if (len(args.contact_type) == 0) or ("all" in args.contact_type):
```
brianhlin, line 135:
```suggestion
    if not args.contact_type or ("all" in args.contact_type):
```

More Pythonic

brianhlin (approved):
LGTM. Two suggestions that I'll merge

osg-cat:
Yeah, I don’t write enough code any more to be Pythonic. Thanks for the tweaks!

PR #3595, opened 2024-01-19
Fix contact type arg default (SOFTWARE-5770)

brianhlin:


brianhlin:
I agree but @osg-cat is going to be the main user and this is the interface that he added in #3590

matyasselmeci (approved):

osg-cat:
@matyasselmeci, dare I say it… “feel free to make a PR”? (^_^)

But yeah, I think the users of the new contacts types are likely to be Christina, Josh D. (and maybe Mike), and me. If it really bothers you, and if you have a solution that does not involve going back and making lots of Topology YAML updates, then propose something concrete.

osg-cat:
Sure, that seems quite reasonable. “instead of” is fine with me.

PR #3603, opened 2024-01-25
Create NOAA_Bell.yaml

rachellombardi:


rachellombardi:
Is it necessary to have the: 

Sponsor:
  CampusGrid:
    Name: OSG Connect

Section of the .yaml?

brianhlin:
@rachellombardi nope, that shouldn't be necessary (whatever the form generates should be right https://topology.opensciencegrid.org/generate_project_yaml).

It looks like Mat's already got a fix for the issue: it was a trailing, unmatched `'`

mwestphall (approved):

PR #3604, opened 2024-01-25
New access point: lambda06.rowan.edu

rynge:


matyasselmeci:
Can you also rename the ResourceGroup file (APs.yaml). Those also need to be globally unique.

Also, you can delete FACILITY.yaml and the ID from SITE.yaml. Those IDs are autogenerated by the webapp now.

matyasselmeci (approved):

PR #3642, opened 2024-02-12
Update Cache DN in SyracuseCachingInfrastructure.yaml

djw8605:
Updated the Cache DN, as shown by FD ticket 75355

@matyasselmeci could you verify that the DN looks correct.  I hand edited it from the output from openssl s_client (replace commas with "/"


matyasselmeci (approved):

matyasselmeci:
Looks fine to me

PR #3646, opened 2024-02-14
Fixed bug where osg-notify gets stuck in a loop while prompting for user password (SOFTWARE-5824)

sqqueak:
While testing osg-notify, it repeatedly prompted me to enter a password despite the password being correct. The script only stopped asking for a password once three incorrect attempts were made. This behavior can be traced back to these two while loops. There is no exit or break condition when there are no exceptions, so I added a break statement which fixed the issue. See [this](https://opensciencegrid.atlassian.net/browse/SOFTWARE-5824) Jira ticket.


brianhlin:
Uh-oh @sqqueak this It looks like this has gone a bit wonky based on the diff + commits in this PR. Feel free to swing by and we can do some surgery to sort it all out

brianhlin (approved):
LGTM!

PR #3654, opened 2024-02-16
Fixed pattern matching error with NetworkInterface objects (SOFTWARE-5423)

sqqueak:
An error occurs when running the osg-notify script with specific parameters (see more at [this](https://opensciencegrid.atlassian.net/browse/SOFTWARE-5423) ticket). This is due to a utility function trying to perform pattern matching on a `NetworkInterface` object, which does not have a default string representation. I modified the pattern matching logic so that it would look for patterns in the correct location and data type.

### DIFF
--- a/src/net_name_addr_utils.py
+++ b/src/net_name_addr_utils.py
@@ -307,8 +307,9 @@ def iface_matches(network_iface, pattern):
     `network_iface` matches `pattern`, False otherwise
 
     """
-    if fnmatch.fnmatch(network_iface, pattern):
-        return True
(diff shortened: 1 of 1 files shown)


brianhlin:
@sqqueak could you rebase this on the latest `master` branch and force push? That'll save us a merge commit

brianhlin:
D'oh, we've had a lot of activity on this repo today. Next time could you do the force push and ping me on Slack so I can hit merge?

brianhlin (approved):
LGTM

PR #3658, opened 2024-02-19
Removed static methods from TopologyPoolManager (SOFTWARE-5347)

sqqueak:
Completed requested changes from previous [PR](https://github.com/opensciencegrid/topology/pull/2782).


brianhlin (changes requested):
Something is a little fishy with the PR as the diff is much larger than I'd expect: https://github.com/opensciencegrid/topology/pull/3658/files. Did you make sure that your `master` was up-to-date with `opensciencegrid/topology`'s `master` when merging it into your branch?

Also, I'd like to see the relevant methods live at the module level vs the class level: the methods fit more as generic "Topology Utilities" instead of methods that are specific to the TopologyPoolManager

PR #3659, opened 2024-02-19
Add "include_downed" and "include_inactive" params to the /osdf/namespaces endpoint

matyasselmeci:
If set, the output will _not_ exclude caches/origins that are in downtime, or marked as inactive.
(SOFTWARE-5826)

Along the way, I noticed that if I try to pass a downtime file (with an entire list of downtimes) to Topology.add_downtime() it will fail with a KeyError because it was expecting a dict.  This was unexpected so I added checks.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -512,7 +512,10 @@ base_path = /ospool/PROTECTED
 
 ### Namespaces JSON generation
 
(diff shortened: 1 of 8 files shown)


brianhlin (approved):
LGTM

PR #3689, opened 2024-02-28
Change /chtc-itb to /chtc/itb/osdf-pelican (since it's served by CHTC-ITB-OSDF-PELICAN-ORIGIN)

matyasselmeci:
Set the Base Path in the CHTC issuer definition of that namespace to `/chtc` to match the other uses of that issuer.

### DIFF
--- a/virtual-organizations/GLOW.yaml
+++ b/virtual-organizations/GLOW.yaml
@@ -130,11 +130,11 @@ DataFederations:
           Issuer: https://chtc.cs.wisc.edu
           MaxScopeDepth: 3
 
-      - Path: /chtc-itb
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
Yes, I think this should work.

That said, I suspect we don't need any registration here at all.  An identical registration, however, shouldn't hurt.

brianhlin:
> That said, I suspect we don't need any registration here at all. An identical registration, however, shouldn't hurt.

@bbockelm could you elaborate on why? Is the idea that the workflow would be:

1. Admin adds their own `Xrootd.ScitokensCfg` that contains a block for the CHTC issuer
2. The origin pushes that config up to the director
3. The caches pull relevant config from the director

PR #3690, opened 2024-02-28
Use YAML references to reduce duplication in GLOW DataFederation definition

matyasselmeci:


matyasselmeci:
Augh cancel this. The "Map Subject" is different in some uses of the issuer.

PR #3708, opened 2024-03-05
Add Prometheus metrics take 2

matyasselmeci:
#2899 but with a dummy class to avoid adding a dependency on prometheus-client to models.py (which is used in scripts such as `get-scitokens-mapfile` that don't pip install stuff).


brianhlin (approved):
LGTM

PR #3709, opened 2024-03-05
Run one WSGI process per Topology container

matyasselmeci:
(we will use Kubernetes replicas for load balancing)

This should avoid conflicts from concurrent git pulls, and give us better metrics (since we'll have one `/metrics` endpoint per pod).


jthiltges (approved):
LGTM!

PR #3718, opened 2024-03-07
adding /knightlab namespace

biozit:
adding /knightlab namespace

### DIFF
--- a/bin/osg-notify
+++ b/bin/osg-notify
@@ -22,6 +22,7 @@ if __name__ == "__main__" and __package__ is None:
     sys.path.append(_parent + "/src")
 
 import topology_utils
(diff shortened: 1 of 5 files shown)


matyasselmeci:
I'm not sure having an authenticated namespace _under_ a public namespace works. I think the public cache instances are going to serve that data without authentication.

biozit:
@matyasselmeci The idea is to send and read data publicly using HTTP requests.

matyasselmeci:
Yes, but /knightlab/mirror/ncbi-genbank-refseq will also be public. Is that what you want?

biozit:
yes, it is

matyasselmeci (approved):

matyasselmeci:
ok :)

PR #3724, opened 2024-03-13
Add new commented-out LHC IAM development instances

brianhlin:
Comment them out since the osg-scitokens-mapfile uses these sections for generating the HTCondor-CE example mapfiles. To do that, we need a reasonable subject and default Unix user.


mwestphall (approved):
LGTM

PR #3736, opened 2024-03-20
adding sage

biozit:
adding sage


matyasselmeci (approved):

bbockelm:
NACK.  This should be a top-level entity.  It's not part of UCSD nor NDP.  We also should be asking the SAGE team what their preference is for issuer name.

PR #3764, opened 2024-03-29
Add mgm.hpc4l.org to HPC4L.yaml

mwestphall:


brianhlin:
Did you run the WLCG Info numbers by Saad? They have effects on WLCG accounting so they need to be right.

Also I'd double check with him on the GitHub account and fix it in COmanage as it doesn't seem quite right.

mwestphall:
@brianhlin Saad confirmed the WLCG numbers are the same for this new host, we should be good to merge these changes in.

brianhlin (approved):
LGTM.

(What about his GitHub username?)

PR #3774, opened 2024-04-03
Fix KAGRA OSDF origin location

xkzl:
Hello @matyasselmeci ,

I noticed that KAGRA OSDF origin is registered in the wrong location. 
It is located in the Kashiwa Data Center. Could you please check if this is correct fix?
Maybe additional tuning might be required.

Best regards,
Marco


brianhlin (approved):
LGTM

PR #3776, opened 2024-04-03
Fix ITB osdf pelican origin endpoint

williamnswanson:
It seems that topology has some default port values for origns and caches that don't match up with what we use for pelican components, and that these are making their way into the pelican director's Web UI:

image

> CHTC-ITB-OSDF-PELICAN-ORIGIN | https://itb-osdf-pelican-origin.osgdev.chtc.io:1095 | http://itb-osdf-pelican-origin.osgdev.chtc.io:1094 |  

[topology origin resource default ports](https://github.com/opensciencegrid/topology/blob/8589aed9ede261633c73f0ee0ee7ce12d64e4013/src/stashcache.py#L573)

I believe this should change the ITB osdf pelican origin's port to the correct 8443, where XRootD is running/listening.


brianhlin (approved):
LGTM

PR #3786, opened 2024-04-09
Add Executive Contact

brianhlin:
This was faster than creating the ticket


matyasselmeci (approved):

brianhlin:
Nope, I `grep`'ed around for existing contact classes and these were the only two locations

PR #3791, opened 2024-04-11
Set ports for the NCAR origin

matyasselmeci:
This is a Pelican origin so both auth and unauth are served on the same port


biozit (approved):
LGTM

matyasselmeci:
Looks like you took care of this in another PR so I'm closing this.

PR #3794, opened 2024-04-12
/sage namespace should not be publicly readable

matyasselmeci:
Discussed on a Zoom call; the /sage namespace is not meant to be publicly readable so shouldn't have PUBLIC auth set.


matyasselmeci:
cc @jhiemstrawisc

biozit (approved):
LGTM

PR #3832, opened 2024-04-25
Add CHTC-KAPEL-TEST EP

matyasselmeci:
This is a set of test EPs for KAPEL integration.
These report to production so we can cross-check records against GRACC.


brianhlin (approved):
LGTM

mwestphall (approved):

PR #3845, opened 2024-05-01
Add topology checkouts to root's git "safe.directory" list

matyasselmeci:
Current versions of git bail out if the git repo is not owned by the current user (even for read-only ops).  Allow root to look at the checkouts even though they're owned by apache.


jthiltges (approved):
If root modifies the repo, permission results will be unpleasant, but that's probably unlikely. The other way around would be to switch to the apache user to view the repos.

Overall, the change looks fine and makes it friendlier for quick access. LGTM.

PR #3856, opened 2024-05-03
Fix Windows newlines and some other spacing issues in project descriptions

matyasselmeci:
Before:
```
Description: "This is a sample file\r\n  with windows line endings\r\nlet's see what\
  \ happens\r\n"
```

After:
```
Description: 'This is a sample file

  with windows line endings

  let''s see what happens'
```

I can't figure out how to make PyYAML generate good-looking multi-line strings but this gets rid of the worst of it.


CannonLock (approved):
LGTM

PR #3865, opened 2024-05-07
Update BNL-ATLAS.yaml

orind:
Updated DNS name for two compute elements that are being rebuilt gridgk0[7,8].racf.bnl.gov -> gridgk0[7,8].sdcc.bnl.gov

### DIFF
--- a/topology/Brookhaven National Laboratory/BNL ATLAS Tier1/BNL-ATLAS.yaml
+++ b/topology/Brookhaven National Laboratory/BNL ATLAS Tier1/BNL-ATLAS.yaml
@@ -220,14 +220,14 @@ Resources:
           ID: OSG1000212
     Description: An OSG CE resource at BNL, mainly serving USATLAS production jobs,
       not open to general OSG users.
-    FQDN: gridgk07.racf.bnl.gov
+    FQDN: gridgk07.sdcc.bnl.gov
     ID: 438
     Services:
       CE:
         Description: Compute Element
         Details:
           hidden: false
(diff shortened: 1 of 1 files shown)


mwestphall (approved):

bbockelm:
Note - this might cause issues with prior accounting data as the accounting data is based on hostname.

@djw8605 - any concerns?

djw8605:
Indeed, this will impact prior accounting.  The ideal case is to keep the old dns entries, and make new resources with the new dns.

glushkov:
Is there a CE retirement procedure in OSG? Or we have to keep in the site configuration all retired resources forever? Can we at least shorten the entry that is left behind to the bare minimum?

bbockelm:
@glushkov - the preferred mechanism is:

1.  Add the new hostname as a separate CE in the same resource group.  That'll allow you to get accounting information to the WLCG, for example, from the new host.
2. Mark the old hostname as `Active: false`.  This indicates the CE is "retired".
3. After a year with no accounting information, we can remove the old CE from git.

We periodically re-compute the accounting numbers to allow late-arriving (or corrected) data to be reflected in the summaries.  The policy is we only recompute information less than a year old; at that point, things can be completely removed.

brianhlin:
> Is there a CE retirement procedure in OSG?

@glushkov as Brian mentioned, the key to retiring a resource in Topology is setting `Active: false`. For future reference, that's documented here: https://osg-htc.org/docs/common/registration/#retiring-resources

PR #3905, opened 2024-05-20
BNL: Retiring gridgk07,8.racf.bnl.gov and Introducing gridgk07,8.sdcc…

glushkov:
….bnl.gov

@bbockelm , @brianhlin  - thanks for the support in https://github.com/opensciencegrid/topology/pull/3865 Hopefully now everything is correct


brianhlin (approved):
LGTM

PR #3907, opened 2024-05-20
Cache and measure the update of the Topology Git repo separately

matyasselmeci:
instead of having the vo data, projects data, mappings data, and resourcegroup data all pulling from the git repo when they get updated.

While I'm at it, use time.monotonic() instead of time.time() for the cache time (which avoids issues if we run, say, ntpdate).


src/webapp/common.py
```diff
@@ -394,6 +395,14 @@ def wrapped():
     return wrapped
 
 
+def get_timestamp():
+    """Return a monotonic timestamp if available, otherwise a wall-clock timestamp."""
+    if hasattr(time, "monotonic"):
```
jthiltges, line 400: Making it backwards compatible is a noble goal, but with time.monotonic() being added since Python 3.3 (which has been EoL since 2017), seems fine to just go ahead and use it without a shim, imho.

jthiltges (changes requested):
LGTM overall, aside from time.monotonic

src/webapp/common.py
matyasselmeci: Sure

jthiltges (approved):
LGTM!

PR #3911, opened 2024-05-21
Rename some project files to match their project names

matyasselmeci:
If present, the Name field in a project yaml takes precedence over the filename. This PR renames two project yamls such that the filename matches the Name.


brianhlin (approved):
LGTM

PR #3928, opened 2024-05-29
Retire /ospool/PROTECTED (INF-1883) and ITB S3 origins

brianhlin:
The old S3 origins needed to be retired and combining these into a single PR allowed us to avoid annoying `base_path`
surgery.


brianhlin:
Dang it

```
     def test_origin_grid_mapfile(self, client: flask.Flask):
        TEST_ORIGIN = "origin-auth2001.chtc.wisc.edu"  # This origin serves protected data
```

matyasselmeci (approved):

PR #3963, opened 2024-06-11
Disable directory listings that are causing redirect failures due to …

bbockelm:
…7.8.4 bug


bbockelm:
To be reverted once 7.8.7 is deployed at the director.

matyasselmeci (approved):

PR #3966, opened 2024-06-12
Unregister the UChicago .well-known directories

matyasselmeci:
/ospool/ap20/.well-known, /ospool/ap21/.well-known, /ospool/ap22/.well-known

These registrations confuse Pelican; fortunately the registrations are unnecessary because the directories are served by an xrootd service that has a manually configured Authfile, and they are not cached.


brianhlin (approved):
LGTM

PR #3980, opened 2024-06-19
ITB: Add "Pelican cache" and "Pelican origin" services (SOFTWARE-5867)

matyasselmeci:
Most cache/origin endpoints work the same for the Pelican cache/origin service types as they do for the XRootD cache/origin services.


matyasselmeci:
Planning on testing this on itb but not deploying to production until Cannon had a chance to look at the map.

PR #3999, opened 2024-07-03
Update LBL_HPCS_downtime.yaml

iraklic:
downtime to upgrade OS to rocky8


iraklic:
I just noticed that something is wrong with this downtime time format.

  StartTime: Oct 04, 2022 17:00 +0000
  EndTime: Oct 05, 2022 17:00 +0000
  CreatedTime: Oct 04, 2022 18:47 +0000

Even CreatedTime which I have not manipulated at all is wrong, which means somewhere between yaml creator and PR times got messed up. I will keep closer eye on this next time. I don't know if someone can have a look and check what yaml creator does.

PR #4006, opened 2024-07-10
register GWU Hosted CE

jdost321:


jdost321:
Ignore the title of this PR it is supposed to be for GWU (George Washington University)

matyasselmeci:
LGTM. I believe GWU is already in the institution database due to having a project (@mwestphall can check) -- if that's true, I think we don't need to ask the project office for approval.

matyasselmeci (approved):

brianhlin:
It's in there but do we want the facility ID to match the institution DB ID (it's ` https://osg-htc.org/iid/67icxo2r0nw7`)?

matyasselmeci:
It doesn't look like we've ever done that before. Do you want to do that piecemeal or would it be better to do a mass switchover, possibly waiting until we have CI to check the results?

PR #4009, opened 2024-07-12
cleaning up some project names

ChristinaLK:
The name in our account management system doesn't match the topology file. Now hopefully they do.


projects/SparseNeuro.yaml
```diff
@@ -0,0 +1,9 @@
+Description: >
```
brianhlin, line 1:
```suggestion
Description: >-
```

brianhlin (approved):
Christina and I spoke about the spinquest rename: users are already submitting with this project and we have other projects that don't follow the project name convention so this is fine

PR #4053, opened 2024-08-02
adding path projects to topology

ChristinaLK:
haven't yet added descriptions for all of them! hopefully soon.


projects/Michigan_2023_Riles.yaml
```diff
@@ -0,0 +1,7 @@
+Department: Physics
+Description: 
+FieldOfScience: Astrophysics
+FieldOfScienceID: '40.0202'
+InstitutionID: https://osg-htc.org/iid/4ocf9kvq30fn
+Organization: University of Michigan?nn Arbor
```
matyasselmeci, line 6: @mwestphall , can you take a look at this? Looks like some unicode got added to the DB for this one.

brianhlin: The DB should support unicode (the UW Madison entry should technically use an en-dash). The question is do we care? If the GRACC is using the `InstitutionID` to look up the organization name, I don't think there are any other consumers of `Organization`.

brianhlin:
Marking this ready for review as this is resulting in some `UNKNOWN` values in the CE dashboard.

@opensciencegrid/project-office how would you like to handle the new NRAO institution?

projects/Michigan_2023_Riles.yaml
matyasselmeci: We'll need to fix either the database entry or the YAML generation since it looks like the encoding of the dash got screwed up _somewhere_.

matyasselmeci, line 6:
```suggestion
Organization: University of Michigan–Ann Arbor
```

brianhlin: It looks ok in the institution DB UI so I think that means that this is an issue with the YAML generation. Does that sound right to you @CannonLock @mwestphall ?

jstathas:
National Radio Astronomy Observatory is approved as Institution Name

jstathas:
OSG ID: https://osg-htc.org/iid/2vy8kkzhno9o

projects/NRAO_2022_Bhatnagar.yaml
```diff
@@ -0,0 +1,7 @@
+Department: 
+Description: 
+FieldOfScience: Astronomy
+FieldOfScienceID: '40.0201'
+InstitutionID: TBD
```
brianhlin, line 5:
```suggestion
InstitutionID: https://osg-htc.org/iid/2vy8kkzhno9o
```

brianhlin (approved):
LGTM. Janet and team will fill out descriptions in subsequent PRs

PR #4080, opened 2024-08-12
Create OregonState_Staff

osu-murphcol:
Requesting access for Oregon State Staff.


osu-murphcol:
Contact information should be colin.murphy@oregonstate.edu, this account may be under an alias murphcol@oregonstate.edu, both route to the same inbox.

brianhlin (approved):
LGTM

PR #4104, opened 2024-08-20
creating /nrp/sdsc

biozit:
@djw8605

### DIFF
--- a/virtual-organizations/NRP.yaml
+++ b/virtual-organizations/NRP.yaml
@@ -123,3 +123,14 @@ DataFederations:
           - SDSC_NRP_OSDF_ORIGIN
         AllowedCaches:
           - ANY
+      - Path: /nrp/sdsc
+        Authorizations:
+          - PUBLIC
+          - SciTokens:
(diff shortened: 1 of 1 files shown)


brianhlin:
Hi @biozit what's this for? Is there a ticket that we can reference?

djw8605:
/nrp/sdsc was an existing namespace used by the NRP.  We are restarting use of it.

bbockelm:
Are we sure `/nrp` is really what we want?

Origins can support any number of namespace prefixes.  My understanding was this was being revived for XenonNT so perhaps `/xenon/sdsc` is more appropriate?

djw8605:
Maybe this is conflating 2 different things.  This requesting is reviving it for the FPGA software for the PNRP.

bbockelm:
> Maybe this is conflating 2 different things

Possibly!  That's why I was asking :)

So, if this is really for NRP itself, PR makes sense.

brianhlin (approved):
Got it, LGTM

PR #4108, opened 2024-08-22
Create Hawaii_Sun.yaml

showmic09:


brianhlin:
@jstathas do you approve the new institution `University of Hawaiʻi at Mānoa`?

brianhlin (approved):
Ah interesting. I think the automerge script is failing on the unicode check. It looks like this institution is already registered. Approving.

PR #4120, opened 2024-08-29
retire sl-um-es5

wuwj6269:


matyasselmeci:
I merged #4119, which already contains these changes, so I will close this.

PR #4171, opened 2024-09-24
adding UW Madison projects, first round

ChristinaLK:


CannonLock:
@ChristinaLK You need two degrees of resolution on the FieldOfScienceID. BMI_Gitter needs to be more specific.

@matyasselmeci Here is the patch to fix the check not catching this

https://github.com/opensciencegrid/topology/pull/4172

brianhlin:
@CannonLock wants to address this in a separate PR. Closing

PR #4172, opened 2024-09-24
Update check_project_fos_precision.yml

CannonLock:


CannonLock:
I set the wrong default branch for PR's and this is not firing on PR's like I intended.

matyasselmeci (approved):

PR #4183, opened 2024-09-27
Create JacksonState_Talchabhadel.yaml

showmic09:


matyasselmeci:
Looks like Jackson State University isn't actually new -- we're setting them up as a resource provider as well -- they just didn't get added to the database. Fixing...

matyasselmeci (approved):

PR #4188, opened 2024-09-27
Add JSU-ap0 resource

aaronmoate:
INF-2178: Configure JSU CM/AP
https://opensciencegrid.atlassian.net/browse/INF-2178

### DIFF
--- a/topology/Jackson State University/JSU RC/JSU-RC.yaml
+++ b/topology/Jackson State University/JSU RC/JSU-RC.yaml
@@ -1,0 +1,41 @@
+Production: true
+SupportCenter: Self Supported
+GroupDescription: This is an OSPool Access Point on our RC Cluster
+
+Resources:
+  JSU-ap0:
+    Active: false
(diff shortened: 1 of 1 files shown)


topology/Jackson State University/JSU RC/JSU-RC.yaml
```diff
@@ -0,0 +1,41 @@
+Production: true
+SupportCenter: Self Supported
+GroupDescription: This is an OSPool Access Point on our RC Cluster
+
+Resources:
+  JSU-ap0:
+    Active: false
```
brianhlin, line 7:
```suggestion
    Active: true
```

Otherwise you won't be able to fetch a token

brianhlin (approved):
This mostly LGTM but @osg-cat and I just had a discussion about organization here: I think the containing dir, `JSU RC`, should probably be renamed to `JSU IT` for https://https://www.jsums.edu/informationtechnology/.

PR #4218, opened 2024-10-07
New proposed institution, UC Merced; FD #76263

osg-cat:
@jstathas, do you approve of this new institution name? https://ror.org/00d9ah105


jstathas (approved):
"University of California, Merced" is approved

mwestphall:
Hi @osg-cat, it looks like Topology's code checks don't like there being a facility with no sites or resources. Do we have preliminary site info for this institution, or is that still forthcoming?

osg-cat:
I was told to request new top-level institutions this way. Should I be doing something differently? And no, we don’t have the rest of the structure yet. I believe @jdost321 or @ashtongraves adds the rest just before or after integration.

PR #4234, opened 2024-10-13
update Security contact and add new CE spce11

dougbenjamin:
Updated the Security Contact information and added new CE spce11.sdcc.bnl.gov


mwestphall:
Hi @dougbenjamin , is resource BNL-SDCC-CE06 being removed and replaced with BNL-SDCC-CE11? If so, we prefer that you update the `Active` field to false in the existing CE06 entry, then create a new CE11 entry.

mwestphall (approved):

PR #4284, opened 2024-11-04
Added missing project: nanoHUB

rynge:


brianhlin:
This looks fine to me but I'm concerned about this message as the following is in the topology institution db:

```
 - New Organization 'Purdue University West Lafayette' requires OSG approval
 ```

@mwestphall any idea why we're getting that message? Have we not updated osgbot to look at the institution db?

brianhlin (approved):

PR #4313, opened 2024-11-19
Use the new pelican-based UChicago project host for writeback and directory listings

matyasselmeci:
This might only affect pre-Pelican clients but I'd like to avoid referencing old endpoints in general.


brianhlin (approved):
LGTM

PR #4320, opened 2024-11-22
Create OSDF_IN_IUCAA_SARATHI.yaml

skjoshi1:
Setting up an OSDF Cache instance at IUCAA, Pune, India for using it with SARATHI Cluster which a part of the IGWN/LIGO Data Grid resources.


mwestphall:
Hi @skjoshi1, would you be able to register yourself as an OSG contact at [this link](https://opensciencegrid.org/docs/common/registration/#registering-contacts)? That will give you an OSG ID that can be populated into the ID field for the administrative/security contacts list.

mwestphall (approved):

PR #4336, opened 2024-12-02
new Swarthmore Firebird Hosted CE

jdost321:


jdost321:
Just a heads up the cluster name here is not a typo. Firebird is a shared cluster between Lafayette and Swarthmore, so this is a different CE from the one registered https://github.com/opensciencegrid/topology/pull/4335 but pointing to the same cluster.

matyasselmeci (approved):

PR #4342, opened 2024-12-03
Update Georgia Tech PACE OSG downtime

zvada:
Add downtime for Georgia Tech PACE OSG due to infrastructure migration


mwestphall:
Hi @zvada , it looks like this PR removes all the old content from the downtime file. Would you be able to update the PR to add the new downtime to the existing list rather than replacing the old entries?

zvada:
> Hi @zvada , it looks like this PR removes all the old content from the downtime file. Would you be able to update the PR to add the new downtime to the existing list rather than replacing the old entries?

oh, sorry, my bad ... I'll fix this asap, do you prefere maybe new PR for this so it's clean? Thanks!

zvada:
> Hi @zvada , it looks like this PR removes all the old content from the downtime file. Would you be able to update the PR to add the new downtime to the existing list rather than replacing the old entries?

Would following [PR](https://github.com/opensciencegrid/topology/pull/4348) replace this one perhaps? If so, we can close this one then and merge [4348](https://github.com/opensciencegrid/topology/pull/4348) it that looks OK?

PR #4354, opened 2024-12-05
update WT2 current CE and SE (for ATLAS)

wyang007:


wyang007:
WT2 was renamed to WT2_CE... Will also upate WT2_downtime.yaml to reflect that change.

PR #4355, opened 2024-12-05
 update WT2 current CE and SE (for ATLAS)

wyang007:
Note: WT2 is renamed to WT2_CE in this pull request. This is reflected in both WT2.yaml and WT2_downtime.yaml

### DIFF
--- a/topology/Stanford University/SLAC/WT2.yaml
+++ b/topology/Stanford University/SLAC/WT2.yaml
@@ -1,4 +1,4 @@
-GroupDescription: (No resource group description)
+GroupDescription: SLAC National Accelerator Laboraotry
 GroupID: 91
 Production: true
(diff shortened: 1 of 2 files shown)


mwestphall:
Hi  @wyang007 , would you be able to update this PR to keep the existing `WT2` around with `Active: false` and create a new entry for `WT2_CE`? This helps us with continuity for record tracking.

mwestphall (changes requested):
We should keep old, inactive entries around for record-keeping purposes, WT2_CE should be added as a new entry under WT2

brianhlin:
There actually haven't been any records attributed to the old CE in the past 2 years so this PR is fine as is. Approved

brianhlin (approved):
LGTM

PR #4362, opened 2024-12-07
Datacenter power upgrades (part 2)

hseverini:


bbockelm:
@matyasselmeci -- is it clear to you what the CI failure is?  The traceback for the HTTP 500 doesn't look obvious.

hseverini:
Hi all,

could it be that the parentheses in the reason -- (part 2) -- is causing
problems, maybe because they're not properly escaped? Just an idea,
I don't know anything about this system. :)

Thanks,

	Horst

mwestphall:
The CI pipeline failure appears to be due to an issue with an external service (which we really should investigate) and not anything specific to this PR, I'll merge it in bypassing the CI checks as a workaround.

mwestphall (approved):

hseverini:
Thanks all!

	Horst

PR #4386, opened 2024-12-19
Add downtime for NMSU-Discovery-CE1 due to scheduled maintenance

nlvw:
Scheduled Biannual Downtime for Maintenance Jan 2025


nlvw:
Not sure if I got the resource right as there were a lot of old/stale entries in [https://topology.opensciencegrid.org/generate_downtime](https://topology.opensciencegrid.org/generate_downtime).  Regardless the cluster itself will be down on the dates specified and a reservation for the specified dates is already registered on our Slurm side of things.  SSH access will be intermittent but job submission will be disabled for the full duration specified via the Slurm reservation.

mwestphall (approved):

PR #4399, opened 2024-12-23
Add downtime for BNL-ATLAS due to BNL F&O Power Intervention

orind:


bbockelm (approved):
Verified this was the correct Ofer out-of-band and will go ahead and merge.

Will strike up an email thread to get the account correctly associated with the resources.

PR #4426, opened 2025-01-07
Update BNL-SDCC.yaml

dougbenjamin:
Add new CE for BNL SDCC shared pool


topology/Brookhaven National Laboratory/BNL-SDCC/BNL-SDCC.yaml
```diff
@@ -129,10 +129,10 @@ Resources:
     Services:
       CE:
         Description: Shared Pool Compute Element
-  BNL-SDCC-CE11:
+  BNL-SDCC-SPOOLCE01:
     Active: true
     Description: This is a compute cluster for SDCC.
-    ID: 
+    ID:
```
brianhlin, line 135:
```suggestion
```

We autogenerate IDs now

brianhlin (commented):
@dougbenjamin we should probably just create a new record unless spce11.sdcc.bnl.gov has long since been retired: we resummarize hours going back 1y so if the CE has been active in the past year, you may lose accounting data in our dashboards and potentially what we send to the WLCG

brianhlin:
It looks like spce11 contributed quite a bit in the [past year](https://gracc.opensciencegrid.org/d/000000079/site-summary?orgId=1&var-site=All&var-type=Batch&var-interval=$__auto_interval_interval&var-Filter=ProbeName%7C%3D%7Chtcondor-ce:spce11.sdcc.bnl.gov&from=now-1y&to=now) so we'll want to deactivate the resource and spin up a new one for the CE instead of the replacement in this PR

PR #4428, opened 2025-01-07
Add CHTC-ITB-PELICANPLATFORM-ORIGIN

matyasselmeci:
This is the origin in that will serve /pelicanplatform in the ITB OSDF. It will live in the osdf-dev namespace on Tiger.

(INF-2538)


matyasselmeci:
Relatedly, how do you feel about moving OSDF-related resources out of CHTC-ITB.yaml and into a new CHTC-ITB-OSDF.yaml (like we have CHTC-ODSF.yaml) so we avoid getting this file too crowded?

brianhlin (approved):
LGTM and sure, the RG organization is there specifically to make our lives easier!

PR #4429, opened 2025-01-08
renaming PDX project due to truncated name in Connect app

ChristinaLK:


brianhlin (approved):
LGTM

@osg-cat approved this solution for the issue we're seeing that resulted in this showing up as an unknown project

osg-cat:
For the record, I don’t think it is my place to approve or disapprove of policy-driven decisions about how to manage researcher accounts and projects.

PR #4430, opened 2025-01-08
Reorganize CHTC-OSDF and CHTC-ITB resources; move OSDF-ITB resources to a separate ResourceGroup

matyasselmeci:


brianhlin (approved):
LGTM

PR #4433, opened 2025-01-08
Refactor two projects

CannonLock:
- These projects were named with the normalized comparator string and not the actual CHTC name. Mapping scripts are case-sensitive so this needed to be updated.


ChristinaLK (approved):
LGTM.

brianhlin (approved):
LGTM

PR #4443, opened 2025-01-09
adding  /sage-backup namespace

biozit:


biozit:
The only reason is to use an external token issuer, but we are trying to avoid it.

brianhlin:
@biozit I assume that this is still necessary?

biozit:
Yes, it is necessary


*--Fábio Andrijauskas*

brianhlin (approved):
LGTM

PR #4448, opened 2025-01-09
removing the capitalization of the URL of CMS's new issuer

ddavila0:
At UCSD we noticed that the mapping of the new CMS issuer was making our CEs to fail the auth of SAM-dev tests. Removing the capitalization in the URL seemed to fix the issue. e.g.

https://CMS-auth.cern.ch/ => https://cms-auth.cern.ch/

More in here:https://ggus.eu/index.php?mode=ticket_info&ticket_id=169391

### DIFF
--- a/virtual-organizations/CMS.yaml
+++ b/virtual-organizations/CMS.yaml
@@ -43,37 +43,37 @@ Credentials:
 
     - Description: CMS IAM development instance
       Subject: bad55f4e-602c-4e8d-a5c5-bd8ffb762113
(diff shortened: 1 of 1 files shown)


brianhlin (approved):
LGTM. Waiting on approval from others in CMS before merging

stlammel:
Thanks Diego! I don't know who requested the capital CMS in the new IAM name. We have it lowercase everywhere else. I assume this was just a mistake and are certainly in favour of changing it to lowercase, especially if a capital CMS isn't working properly.
 - Stephan

PR #4492, opened 2025-01-21
Add CHTC_PELICANPLATFORM_ORIGIN, which will serve the /pelicanplatform namespace in OSDF

matyasselmeci:


brianhlin (approved):
LGTM, though the `AllowedVOs` seems unnecessary

matyasselmeci:
I think the attribute is necessary (that is, Topology will complain if it's missing), but we shouldn't be restricting it to GLOW.

PR #4499, opened 2025-01-23
Create new resource for University of Wisconsin–La Crosse

flyn-org:


matyasselmeci (approved):

matyasselmeci:
A resource requires at least one Service; Tim told me you'll be running Container EPs, so I added an "Execution Endpoint" service (that being an old name for what we now call an EP).

PR #4505, opened 2025-01-27
Build from OSG 24, pulling software-base from Harbor

matyasselmeci:
keeping it on EL8 but preemptively enabling SHA1 for when we upgrade.


brianhlin (approved):
LGTM

PR #4512, opened 2025-01-30
Added Laf-Swat Firebird contacts (many just names); FD #78353

osg-cat:


osg-cat:
Note: I have names for all of their contacts, but getting people to register in COmanage is like pulling teeth. So I put in what we have and want to move on now. If they register more people, it is easy to update the contacts structure.

topology/Lafayette College/Lafayette ITS/Lafayette-Firebird.yaml
```diff
@@ -17,6 +17,39 @@ Resources:
...
+      Local Operational Contact:
+        Primary:
+          Name: Peter Goode
+          ID: OSG1000899
+        # Secondary:
+        #   Name: Jason Simms
+        #   ID: 
+        Tertiary:
+          Name: <FIRSTNAME> <LASTNAME>
+          ID: <EMAIL HASH>
```
brianhlin, line 40:
```suggestion
```

brianhlin (approved):
LGTM

PR #4533, opened 2025-02-05
Added resource group for Colgate University

tsoukalask:


topology/Colgate University/Colgate ITS/Colgate-CCARE.yaml
```diff
@@ -0,0 +1,180 @@
...
+    # If you cannot find the contact above XML, please register the contact:
+    # https://opensciencegrid.org/docs/common/registration/#registering-contacts
+    ContactLists:
+      # Administrative Contacts are persons or groups of people (i.e.,
+      # mailing lists) that are directly responsible for the
+      # maintenance of the resource
+      Administrative Contact:
+        Primary:
+          Name: Kyriakos Tsoukalas
+          ID: 6c83aeaf8c7a03c7b8f82792b1c5253e2b0fe90829d7cb49c95ec0bc09ff62f4
```
mwestphall, line 33:
```suggestion
          ID: OSG1000925
```

topology/Colgate University/Colgate ITS/Colgate-CCARE.yaml
```diff
@@ -0,0 +1,180 @@
...
+        #   Name: <FIRSTNAME> <LASTNAME>
+        #   ID: <EMAIL HASH>
+
+      # Security Contact are persons or groups of people (i.e.,
+      # mailing lists) that are responsible for handling security
+      # issues related to the resource
+      Security Contact:
+        Primary:
+          Name: Kyriakos Tsoukalas
+          ID: 6c83aeaf8c7a03c7b8f82792b1c5253e2b0fe90829d7cb49c95ec0bc09ff62f4
```
mwestphall, line 47:
```suggestion
          ID: OSG1000925
```

mwestphall (commented):
We should use OSGID, found via https://topology.opensciencegrid.org/contacts , when available

osg-cat:
@brianhlin, thanks for asking. I did give Kai the complete set of identifiers to use, but it seems like the Resource name got lost in translation. The Resource Group of “Colgate-CCARE” is good, but the Resource name itself should be “Colgate-CCARE-EP”. Could you update that, please?

I have no idea whether “GPU-Compute” is an appropriate Services identifier.

And it seems like you could take out the “Other services” of Slurm. Not sure that’s relevant at this time, and if it becomes relevant, there will be a separate Resource for it.

matyasselmeci (approved):

PR #4556, opened 2025-02-13
change fqdn of ap22 pelican origin

djordan66:


djordan66:
Closed pull request. Will do something different

PR #4564, opened 2025-02-19
Register osdf-itb-origin2000 (INF-2739)

theng-vang:


matyasselmeci:
Hi Theng, could you move this to CHTC-OSDF-ITB.yaml and also name it CHTC-ITB-ORIGIN2000?

matyasselmeci (approved):

PR #4589, opened 2025-02-24
Add CHTC-wright-ap to CHTC facility CHTC site

aaronmoate:
Done in support of INF-2841: Configure wright-ap4000 to flock to the OSPool
https://opensciencegrid.atlassian.net/browse/INF-2841


brianhlin (approved):
LGTM

PR #4618, opened 2025-03-05
using NEBRASKA_NRP_OSDF_ORIGIN

biozit:


brianhlin (approved):
LGTM but can you leave a comment as to why we're making this change? What's happening with the previous origin?

biozit:
> LGTM but can you leave a comment as to why we're making this change? What's happening with the previous origin?

I used the wrong origin

PR #4621, opened 2025-03-06
Update public gwdata LIGO origin to be marked as Pelican

bbockelm:
Per Josh Willis, this host has been cut over to Pelican to attempt to fix the access to public `/gwdata` namespace.


bbockelm:
Bypassing merge rules as `/gwdata` is otherwise broken.

PR #4622, opened 2025-03-06
Disable non-functioning origins for `/gwdata`

bbockelm:
The `/gwdata` namespace is currently only functioning on `CIT_LIGO_ORIGIN`; momentarily comment out the other two origins until we are able to diagnose the problems with them (esp. as they are non-Pelican).

(Purposely commenting out so this change instead of removing the lines so this is clearly visible.)


bbockelm:
Bypassing merge request to make `/gwdata` work after-hours.

PR #4625, opened 2025-03-06
Migrate CIT_LIGO_ORIGIN_TEST_WRITE from XRootD origin server to Pelican origin server (FD #79132)

matyasselmeci:


virtual-organizations/LIGO.yaml
```diff
@@ -272,8 +272,8 @@ DataFederations:
         AllowedOrigins:
           - CIT_LIGO_ORIGIN_TEST_WRITE
         AllowedCaches: *ligo-allowed-caches
-        Writeback: https://origin-writetest.ligo.caltech.edu:1095
-        DirList: https://origin-writetest.ligo.caltech.edu:1095
+        Writeback: https://origin-writetest.ligo.caltech.edu:8443
+        DirList: https://origin-writetest.ligo.caltech.edu:8443
```
brianhlin, line 276: Since the Topology config here is just for the caches, I don't think we actually need these anymore. I imagine that it was just the clients that needed this bit

brianhlin (approved):
LGTM with one minor comment (that's maybe more of a question). As discussed, let's hold off on this until we can coordinate with Josh + approve in the registry.

virtual-organizations/LIGO.yaml
matyasselmeci: Looks like it's only used by pre-Pelican clients, and also the director for marking pre-Pelican origins as supporting listings and writeback. I'll remove it.

PR #4627, opened 2025-03-07
Using nrp domain dtn-pas.kans.nrp.internet2.edu

biozit:


brianhlin (approved):
LGTM but what do we actually need the DN for these days?

PR #4639, opened 2025-03-11
HW fixed

biozit:


matyasselmeci:
Fabio, the EndTime is before the StartTime. Can you take a look? (did you mean Mar 10 instead of Feb 10?)

PR #4640, opened 2025-03-12
Add back Writeback and DirList to the /igwn/test-write namespace

matyasselmeci:
Looks like the director uses these to mark the namespace as supporting Write and Listing, respectively.

This reverts commit edf19a8f87a1c6d73df54be0a497b38ea65bc862.


brianhlin:
Let's see if this is necessary post namespace registration

matyasselmeci:
Looks like it wasn't: 
image

PR #4653, opened 2025-03-18
Update JLab-Farm.yaml

kurtjstrosahl-jlab:
osg-ce-1 has been retired from service, and can be removed from the topology file

### DIFF
--- a/topology/Thomas Jefferson National Accelerator Facility/JLAB/JLab-Farm.yaml
+++ b/topology/Thomas Jefferson National Accelerator Facility/JLAB/JLab-Farm.yaml
@@ -236,30 +236,6 @@ Resources:
         Description: JLab Pelican origin
     VOOwnership:
       JLab: 100
-
-  JLab-FARM-CE:
-    Active: true
-    ContactLists:
-      Administrative Contact:
-        Primary:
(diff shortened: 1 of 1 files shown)


mwestphall:
Hi Kurt, would you be able to update this PR to keep the contents of `JLab-FARM-CE` present but set `Active: false` rather than fully removing it? Fully deleting resources breaks some topology reporting scripts.

topology/Thomas Jefferson National Accelerator Facility/JLAB/JLab-Farm.yaml
```diff
@@ -236,30 +236,6 @@ Resources:
         Description: JLab Pelican origin
     VOOwnership:
       JLab: 100
-
-  JLab-FARM-CE:
-    Active: true
```
mwestphall, line 241: This field can just be set to `Active: false` rather than fully deleting the resource

mwestphall (changes requested):

topology/Thomas Jefferson National Accelerator Facility/JLAB/JLab-Farm.yaml
kurtjstrosahl-jlab: Done!  Will note that for future work

mwestphall (approved):

PR #4660, opened 2025-03-19
Add downtime for UNL cache and Origin due to NRP issue

biozit:


mwestphall:
Hi @biozit , it looks like we don't support having downtimes for both the origin and the cache in a single downtime file. Would you be able to move the downtime for `NEBRASKA_NRP_OSDF_CACHE` into the `UNLCachingInfrastructure_downtime.yaml` downtimes file instead?

PR #4684, opened 2025-04-02
Register CHTC-INTERACTIVE-OSPOOL-EP

matyasselmeci:
an EP set up on tiger for interactive jobs, to be run by OSG Staff


brianhlin (approved):
LGTM

PR #4697, opened 2025-04-07
Replace LIGO security contacts for origin-readtest.ligo.caltech.edu and origin-writetest.ligo.caltech.edu

matyasselmeci:
See https://support.opensciencegrid.org/a/tickets/79132


brianhlin (approved):
LGTM

PR #4716, opened 2025-04-17
Use generalized image build flow

CannonLock:
Moves the image build to a generalized docker build flow intended for flux automations. I use this action for quite a few other things set up to track images and it supports the current topo dev flow. 

https://github.com/osg-htc/build-push

Basically it will build all the images that track a branch with a appended timestamp that flux can consume and interpret as the latest image. 

I am noting that I will not be pushing to docker anymore, is this a issue?


brianhlin:
Seems reasonable to me. Any objections @matyasselmeci @mwestphall

PR #4726, opened 2025-04-21
UCOPS-128: drop retired and Pelican Topology UC namespaces and drop Authfile and grid-mapfile tests

brianhlin:
All of this is now handled by the Pelican services + OSDF Registry


brianhlin:
@matyasselmeci `test_origin_grid_mapfile` depends on ap20 -- maybe we should just drop this test altogether in the Pelican World Order?

matyasselmeci (approved):

PR #4728, opened 2025-04-22
Beoshock At WSU

nathanrwells-ksu:
Add Beoshock at Wichita State University


osg-cat:
I want to switch to using `BeoShock` everywhere, as this appears to be the representation of that name on the WSU site(s). We can do that internally.

matyasselmeci (approved):

PR #4729, opened 2025-04-22
InstitutionID update and improved verification check

CannonLock:
Fix issues where people omitted or added null InstitutionId's for facilities that have resources registered. 

This is going to fail until we get something assigned to CancerComputer and Cybera but that is good because it will remind us to fix this until it is fixed :)


.github/scripts/check_facility_has_institution_id/main.py
```diff
@@ -51,6 +63,15 @@ def main():
             except Exception as e:
                 errors.append((file.split("/")[-2], e))
 
+    # Check that we have the same amount of facility filhttps://osg-htc.org/iid/52f5piuly2gges as topology facilities
```
matyasselmeci, line 66: Bad copy-paste here?

CannonLock: Exactly that, fixed.

matyasselmeci (approved):

PR #4773, opened 2025-05-06
Update CIT_LIGO resources to reflect migration of origins to Pelican

matyasselmeci:
[FD 79845](https://support.opensciencegrid.org/a/tickets/79845)


brianhlin (approved):
LGTM! As discussed during the OSDF Ops meeting, I think we need to get rid of the namespace entries from VO/LIGO.yaml to get those origins to stop showing up in the director

matyasselmeci:
Should I go do that now, or is there something else that needs to be done first?

PR #4783, opened 2025-05-11
Deactivated old resources; FD #79900

osg-cat:


osg-cat:
@jdost321, see any reason why we should not deactivate these old CEs and APs at Utah? The CEs are the old SLATE ones and should have been deactivated long ago, I think. Neither Brian H. (Utah) nor I could find any evidence that the APs still exist or have been used recently.

matyasselmeci (approved):
Approving but let's wait until @jdost321 's input before merging.

jdost321:
Looks good to me, please merge!

PR #4800, opened 2025-05-20
adding pelicanfacilitation namespace

biozit:


matyasselmeci:
I assume the origin hosting this is using Pelican; in that case, the namespace registration is not necessary, because the registry will be used to handle it instead.

PR #4817, opened 2025-05-28
Adding new I2 AMST cache

biozit:


topology/Internet2/Internet2Amst/I2BoiseInfrastructure.yaml
matyasselmeci, line 1: Can you rename this file? It's not in Boise.

biozit: done

PR #4818, opened 2025-05-28
Retire ComputeCanada cache

biozit:
The current node, installed in 2016, is very old and unstable. We need to retire it.


brianhlin:
@biozit it's been a few months so I'm closing this

PR #4821, opened 2025-05-29
Revert "Add downtime for Stashcache-UofA due to too many open files - redirec…"

biozit:
Reverts opensciencegrid/topology#4820


biozit:
“Too many open files in system” is errno ENFILE.  The value of ENFILE is 23 on Linux.  Libcurl’s error code 23 is “write callback error” which is generated when the origin has an invalid response to a cache.

So, “too many open files in system” is the mistranslation of “invalid origin response”.  Nothing to do with the literal system.

Prepping a PR to improve this - the existing code handles it wrong because the version if libcurl we test with has a different error code than the version on RHEL9 and the unit test gets it wrong.

PR #4847, opened 2025-06-09
BNL: Removing .racf. CEs

glushkov:


mwestphall:
Hi @glushkov , our github actions validation don't work well with fully removing resource definitions. Would you be able to just set `active: false` on those resources instead?

PR #4863, opened 2025-06-16
Create Psychology_Rogers.yaml

aowen-uwmad:

PR #4871, opened 2025-06-19
Mass-rename "University of Wisconsin-Madison" to have an endash instead of a hyphen

matyasselmeci:
This matches Psychology_Rogers.yaml (https://github.com/opensciencegrid/topology/pull/4863)


brianhlin (approved):
LGTM

djw8605:
Grafana seems ok with the change.  I think everywhere that matters properly converts the unicode.  I'm sure we will have problems somewhere... but it seems fine for now.

image

djw8605 (approved):

PR #4910, opened 2025-07-08
Add new scitokens issuer at JLAB to GLUEX mapping

jasoncpatton:
This PR assumes that the mapfile generated from Topology can be many to one, that is both the UConn and JLAB issuer URLs can map to the `gluex` local user. Both issuer URLs contain the same public key data, but the (new) JLAB URL is expected to be the long term home for the issuer. The idea with listing them both is that it gives us some time for uptake of the new mapfile before switching the frontend over to issuing tokens with the JLAB issuer URL.


brianhlin (approved):
LGTM

PR #4933, opened 2025-07-21
Preliminary work for facility renaming (SOFTWARE-6183)

matyasselmeci:
This adds / fixes some data that we need to do before renaming any facilities.

- For facilities that didn't have the ID field, add them. (If missing, they're autogenerated based on names.) The IDs are required to maintain continuity between the old names and the new names

- Fix the InstitutionID of the University of Wyoming (it was actually set to the ROR ID, not the OSG ID)

- Add InstitutionID fields to some facilities that didn't have them. Generally, they were in the Institution DB under their English names.


brianhlin (approved):
LGTM, feel free to rebase and merge

PR #4943, opened 2025-07-29
Add VO to SURFMS4.yaml

lnauta:
The SURF site has upgraded to pelican and the dashboard is all green, we see data flowing and files arriving.

For testing we removed the VO some months ago, now that the new setup works the VO is placed back.


matyasselmeci:
Added a commit to rename the resource in the _downtime.yaml as well (https://github.com/opensciencegrid/topology/commit/0df0768744f0bd3545c779629295b3cb31e3f467) and merged it by hand. You should be all set.

PR #4991, opened 2025-08-19
cleanup of slate squid as it does not exist anymore

ivukotic:


topology/University of Michigan/AGLT2/AGLT2.yaml
```diff
@@ -2,40 +2,6 @@ GroupDescription: (No resource group description)
 GroupID: 1
 Production: true
 Resources:
-  AGLT2-squid-2:
-    Active: true
```
mwestphall, line 6: @ivukotic we prefer to keep de-activated resources around for accounting purposes, would you be able to un-delete `AGLT2-squid-2` and its downtimes and then just set this field to `Active: false` instead?

PR #4998, opened 2025-08-21
Create UAF_PISMCloud.yaml

aowen-uwmad:
Note: slashes in the description were automatically added by the Project YAML Generator..


matyasselmeci (approved):
LGTM. The backslashes don't show up in the final result -- they are used to escape the newlines so the resulting description shows up as a single-line string.

PR #5006, opened 2025-08-26
Added new GT Buzzard resource group

osg-cat:


osg-cat:
@zvada, @jdost321, @brianhlin, @biozit, @rynge — sorry for all the ats, but this PR contains all new Topology registrations for the GA Tech Buzzard cluster. See this ticket: https://support.opensciencegrid.org/a/tickets/80670  You are welcome to add a formal review or just check things over. See the attached spreadsheet from today to see where the various IDs came from, etc.

zvada:
This looks good to me with only one comment:

`GATech-Buzzard-Squid1` defines DNS alias `buzzard-squid.pace.gatech.edu` which points to two real squid hosts running Frontier Squid service in HA mode. Those two hosts are: `buzzard-squid-{1,2}-ext.pace.gatech.edu`. Unless you're registering both underlying hosts (which we agreed in another thread you don't have to), I'd propose define only Squid element as `GATech-Buzzard-Squid` instead of `GATech-Buzzard-Squid1`.

osg-cat:
Marian, the “1” suffix is generally part of our naming conventions.
@jdost321, @brianhlin — any strong preferences for keeping or dropping the “1” suffix for their Squid service?
In any case, can we wrap this up today?

jdost321:
@osg-cat @brianhlin I think we can go with Marian's suggestion and drop the 1 for the squid. The convention makes sense for the general CE case if we ever want to add one for redundancy. But as Marian stated, they will do that behind the DNS. Also we don't didn't do this for the cache either.

osg-cat:
Ok, I made that change. Last call for change requests! Let’s get this done by 4pm CT today!

PR #5420, opened 2026-05-13
Add red-squid.unl.edu

DrDaveD:
Copied from red-squid1 & red-squid2


DrDaveD:
Nevermind, it's only a virtual host

PR #5462, opened 2026-06-08
Remove old CHTC learn submit registration

brianhlin:
We provisioned a new host that reuses the FQDN but now sets the
SiteName to `CHTC-learn-ap`


osg-cat (commented):
What are the consequences of having one of the two Resources for the “learn” FQDN be active and the other inactive? Would it be better to completely remove one entry?

brianhlin:
@osg-cat there shouldn't be any. The one I'm setting inactive in this PR was the resource associated with the now-retired old submit host, which had `SiteName: "CHTC-learn-submit"`

osg-cat (approved):
Ok, sounds good.

PR #5472, opened 2026-06-15
Align UW OSDF cache naming

brianhlin:
We call it the "UW OSDF Cache" / that's what the deployment is called,
so this fixes the close-but-not-quite FQDN.

TimC asked for this specific site name. It's fine by me because even
if it isn't exactly aligned it has both "UW" and "CHTC" in the name so
any sort of grep / search in the director UI will likekly yield the
cache that we want.

Create a new resource + deactivate the old one in case there's
anything that cares about the old FQDN and resource name.


topology/University of Wisconsin–Madison/CHTC/CHTC-OSDF.yaml
```diff
@@ -225,6 +225,40 @@ Resources:
     AllowedVOs:
       - ANY
 
+  UWMadison-CHTC-OSDF-Cache-1:
```
matyasselmeci, line 234: Why the `-1`?

brianhlin: In case we spin up another

matyasselmeci: Can we change the hostname too?

brianhlin: To include the `-1`? Sure.

FWIW, I'm ok with naming that doesn't exactly align between all of them

1. Deployment: `uw-osdf-pelican-cache`
2. FQDN: `uw-osdf-cache.svc.osg-htc.org`
3. Resource name `UWMadison-CHTC-OSDF-Cache-1`

Aligning everything on 3 would be kind of annoying to type out

matyasselmeci: Yeah, I'm less bothered by the prefix mismatch than by the inconsistency with the `-1`. The prefix is essentially for avoiding collisions, but the suffix is our choice, and it would be nice if we were consistent with it.

topology/University of Wisconsin–Madison/CHTC/CHTC-OSDF.yaml
```diff
@@ -231,6 +231,40 @@ Resources:
...
+          ID: OSG1000589
+          Name: William N Swanson
+        Secondary:
+          ID: OSG1000002
+          Name: Matyas Selmeci
+        Tertiary:
+          ID: OSG1000003
+          Name: Brian Hua Lin
+    FQDN: uw-osdf-cache.svc.osg-htc.org
+    DN: /CN=uw-osdf-cache.svc.osg-htc.org
```
matyasselmeci:
```suggestion
    FQDN: uw-osdf-cache-1.svc.osg-htc.org
    DN: /CN=uw-osdf-cache-1.svc.osg-htc.org
```

topology/University of Wisconsin–Madison/CHTC/CHTC-OSDF.yaml
```diff
@@ -198,7 +198,7 @@ Resources:
       - ANY
 
   CHTC_PELICAN_CACHE:
-    Active: true
+    Active: false
```
matyasselmeci, line 201:
```suggestion
    Active: true
```
I'd hold off on deactivating the existing cache until you've actually brought up the new one.

matyasselmeci (approved):
One suggestion; pre-approving.

PR #5492, opened 2026-06-29
Update secondary contact information for OSG AP host in BNL-SDCC.yaml

xzhao87:
add Xin Zhao as secondary security contact for the OSG AP host at BNL (osgsub01), so he can receive notifications when OSG blocks this AP host due to security update requirements.


xzhao87:
I added the missing ID and the checks are now passing.

brianhlin:
Waiting on @dougbenjamin 's feedback on whether or not he should be replaced as a security contact here

xzhao87:
OK thanks Brian for the info.

dougbenjamin:
who is using this AP host from BNL?  I would think that Robert Hancock should be listed as a security contact also. I do not need to be a security contact if DUNE is not using this AP. I don't believe it is used by ATLAS

xzhao87:
Hi Doug -- This AP is used by ePIC PanDA to submit jobs to the OSG pool.  Rob is the primary security contact already, that won't change.

brianhlin:
@dougbenjamin in case you didn't get notified of Xin's response, I've copied it below:

> Hi Doug -- This AP is used by ePIC PanDA to submit jobs to the OSG pool. Rob is the primary security contact already, that won't change.

dougbenjamin:
Since this is for ePIC, I certainly don't need to be listed as a security contact.

brianhlin (approved):
LGTM

PR #5506, opened 2026-07-03
Add API key support

matyasselmeci:
This pull request introduces API key authentication to the application, allowing clients to authenticate using bearer tokens in the `Authorization` header instead of X.509 certs. It also adds a script for creating API keys.

The API key must be a UUID with a "tk-" ("topology key") prefix. The client sends the API Key as a "Bearer" credential in the HTTP Authorization header.  (As-is, no need to base64-encode it, since the format already only uses URL-safe characters.)

The hash of the incoming key is checked against hashes loaded from an API_KEYS_FILE. 

API_KEYS_FILE must be a YAML mapping of ID to FullName and APIKeyHash (similar to contacts.yaml) and the IDs and FullNames are checked against the topology contacts.

There is a `bin/make_api_key.py` script to help generate valid keys and API_KEYS_FILE blocks.

The Docker image includes a stub api_keys.yaml file in `/secrets/api_keys/api_keys.yaml`; the extra directory layer is there so you can mount a Secret over the `/secrets/api_keys` directory and have it get updated as you update the Secret. Topology re-reads the file every 5 minutes.

### DIFF
--- a/bin/make_api_key.py
+++ b/bin/make_api_key.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+
+import argparse
+import hashlib
+import pathlib
+import re
+import sys
+import urllib.request
+import uuid
+import xml.etree.ElementTree as ET
+
+USERS_ENDPOINT = "https://topology.opensciencegrid.org/miscuser/xml"
+VALID_API_KEY_RE = re.compile(
+    r"^tk-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+)
+
+
+class NoMatch(Exception):
+    pass
+
+
+def get_user_info(fullname=None, id_=None, endpoint=USERS_ENDPOINT) -> tuple[str, str]:
+    """
+    Get the miscuser XML from Topology, look up a user by ID or FullName,
+    and return a tuple containing the (FullName, ID) from Topology, or raise
+    NoMatch if no record is found.
+
+    The XML is retrieved from the URL at `endpoint` and looks something like this:
+    <Users>
+        <User>
+            <FullName>Name Of Person</FullName>
+            <ID>(either an OSG ID or a hash)</ID>
+        </User>
+    </Users>
+
+    If neither fullname nor id_ are specified, ValueError will be raised.
+    If both fullname and id_ are specified, looks up the record by ID and
+    makes sure the FullName matches, raising a NoMatch exception if it
+    does not.
+    """
+    if not fullname and not id_:
+        raise ValueError("Either fullname or id_ must be provided")
+
+    # Fetch the XML from the endpoint
+    with urllib.request.urlopen(endpoint) as response:
+        xml_data = response.read()
+
+    # Parse the XML
+    root = ET.fromstring(xml_data)
+
+    # Search through users for a match
+    for user in root.findall('User'):
+        user_fullname = user.findtext('FullName', '').strip()
+        user_id = user.findtext('ID', '').strip()
+
+        if id_ and fullname:
+            # Both specified: match by ID and verify FullName matches
+            if user_id == id_:
+                if user_fullname == fullname:
+                    return (user_fullname, user_id)
+                else:
+                    raise NoMatch(
+                        f"ID {id_} found but FullName does not match. Expected {fullname}, got {user_fullname}"
+                    )
+        elif id_:
+            # Match by ID only
+            if user_id == id_:
+                return (user_fullname, user_id)
+        elif fullname:
+            # Match by FullName only
+            if user_fullname == fullname:
+                return (user_fullname, user_id)
+
+    # No match found
+    if id_ and fullname:
+        raise NoMatch(f"No user found with ID {id_} and FullName {fullname}")
+    elif id_:
+        raise NoMatch(f"No user found with ID {id_}")
+    else:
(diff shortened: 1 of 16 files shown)


src/app.py
```diff
@@ -5,6 +5,7 @@
 import flask
 import flask.logging
 from flask import Flask, Response, make_response, request, render_template, redirect, url_for, session
+import hashlib
 from io import StringIO
```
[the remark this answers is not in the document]
matyasselmeci: Fixed

src/webapp/models.py
```diff
@@ -268,6 +270,20 @@ def get_dns(self) -> Optional[Set]:
                 self.contacts_data.try_again()
         return self.dn_set.data
 
+    def get_api_keys(self) -> Optional[Dict[str, str]]:
+        if self.api_key_set.should_update():
+            contact_db_data = self.get_contact_db_data()
+            try:
+                self.api_key_set.update(
+                    contacts_reader.get_api_keys_data(self.api_keys_file, contact_db_data)
+                )
```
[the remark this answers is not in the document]
matyasselmeci: Fixed.

matyasselmeci:
Tested on topology-itb, with a contacts.yaml contact ID. COManage-only users aren't showing up on topology-itb at all for some reason (broken LDAP credentials maybe?) so I wasn't able to test those.

bin/make_api_key.py
```diff
@@ -0,0 +1,219 @@
...
+import os
+import pathlib
+import re
+import sys
+import urllib.request
+import uuid
+import xml.etree.ElementTree as ET
+
+USERS_ENDPOINT = "https://topology.opensciencegrid.org/miscuser/xml"
+VALID_API_KEY_RE = re.compile(
```
mwestphall:
```suggestion
# Regex for an API key, formatted as `tk-` followed by a UUID
VALID_API_KEY_RE = re.compile(
```

src/README.md
```diff
@@ -606,3 +606,57 @@ The final result looks like
...
+
+Only contacts registered in contacts.yaml or COManage can have API keys.
+API keys not associated with a valid contact are not accepted.
+A contact may only have one API key.
+The API keys file is re-read at the same frequency as contact info
+(configured via CONTACT_CACHE_LIFETIME).
+
+Use the `bin/make_api_key.py` script to make a new API key and also output
+the YAML block that should be copy-pasted into the API keys file.
+You may specify the contact by FullName (`--name`) or ID (`--id`).
```
mwestphall, line 636: Might be worth specifying the environment prerequisites for running the script here? It "just works" on my laptop, I assume because x509 auth is configured?

src/webapp/common.py
```diff
@@ -400,3 +400,35 @@ def wrapped():
...
+    if isinstance(token, bytes):
+        token_s = token.decode("latin-1")
+    else:
+        token_s = token
+    if not API_KEY_RE.fullmatch(token_s):
+        raise ValueError("Token does not match pattern")
+    return "sha256:" + hashlib.sha256(token_s.encode()).hexdigest()
+
+
+def shorten(a_str: str, maxlen: int = 80) -> str:
```
mwestphall: Is this function used? I see it imported but not called from anywhere.

bin/make_api_key.py
```diff
@@ -0,0 +1,219 @@
...
+    if keyfile:
+        api_key = read_api_key_from_file(keyfile)
+    else:
+        api_key = "tk-" + str(uuid.uuid4())
+        assert VALID_API_KEY_RE.fullmatch(api_key)
+    return api_key
+
+
+def make_key_hash(api_key: str) -> str:
+    """
```
mwestphall: Small amount of logic re-implementation between this and token_to_apikeyhash in common.py, might be worth it to refactor into a shared library function.

src/app.py
```diff
@@ -1105,8 +1154,61 @@ def _get_authorized():
...
+    if app and app.logger:
+        log = app.logger
+    else:
+        log = logging.getLogger(__name__)
+
+    auth_header = request.headers.get("Authorization", "")
+    if not auth_header.startswith("Bearer "):
+        return False
+
+    bearer_token = auth_header[7:].strip(" \t\n'\"")
```
mwestphall, line 1202: Will this positional parsing of the auth header cause issues with leading/trailing spaces? eg `Bearer  tk-...`

mwestphall (approved):
Small comments, mostly relating to clarity of documentation, pre-approving

mwestphall:
We will also want to test this once deployed into ITB topology, I only ran the tests and used the CLI tools locally and did not validate that they work against a running docker container.

src/README.md
matyasselmeci: No, FullName and ID are public. There aren't any environment prerequisites other than outbound interent.

src/app.py
matyasselmeci: No, we've already verified that the first 7 characters are `Bearer ` and the `strip()` takes care of any whitespace afterward. (The argument to `strip()` is not a sequence of characters, but is treated as a set, so any combination of its characters are removed.)

src/webapp/common.py
matyasselmeci: Not anymore. I deleted it.

bin/make_api_key.py
matyasselmeci: I decided to use the function from common.py here in make_api_key.py as well.

PR #5580, opened 2026-08-17
Remove legacy squid servers from Vanderbilt University

appeltel:
Squid and Squid2 are obsolete and replaced with Squid4 and Squid5 - remove the old entries so we can decommission them.


brianhlin (approved):
LGTM

PR #5607, opened 2026-09-01
Add MERIT_SCOPE_X.yaml for SCOPE-X project

dmora127:
Added project details for SCOPE-X initiative including description, field of science, and principal investigator.

### DIFF
--- a/projects/MERIT_SCOPE_X.yaml
+++ b/projects/MERIT_SCOPE_X.yaml
@@ -1,0 +1,17 @@
+Department: R&D
+Description: 'SCOPE-X (Scalable Cyberinfrastructure for Observatory Pipeline Enhancement)
+  advances the cyberinfrastructure of ORION (Observatory for Cyber-Risk Insights and
+  Outages of Networks), one of the oldest and largest operational network telescopes
+  in the United States, operated by Merit Network. ORION provides visibility into
+  unsolicited and potentially malicious Internet traffic and supports shared cybersecurity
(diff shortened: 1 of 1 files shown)


dmora127:
@opensciencegrid/project-office can you verify the detail for MERIT for this project? Looks like it might be missing from the organizations in topology.

brianhlin:
@dmora127 nah it's just that our CI is old and bad. @matyasselmeci where does that organization check live so we can rip it out (or update it to look for a...missing `InstitutionId`)?

matyasselmeci:
Probably the webhook? You might want to debug why it didn't find the institution this time - it usually seems to get it right.

brianhlin:
@matyasselmeci I thought that was because the webhook is driven off of https://github.com/opensciencegrid/topology/blob/master/mappings/project_institution.yaml, which is no longer the source of truth.

EDIT: Ah nope, it's even worse. We just look to see if the `Organization` has already been used in an existing project [here](https://github.com/opensciencegrid/topology/blob/master/src/webapp/automerge_check.py#L129-L133) and [here](https://github.com/opensciencegrid/topology/blob/master/src/webapp/automerge_check.py#L223))

matyasselmeci:
Ah, I see. We should probably make the webhook look at the topology institutions api (https://github.com/osg-htc/topology-institutions-api) (bonus points for actually checking that the InstitutionsID matches)...

brianhlin (approved):
@dmora127 sorry we got lost in the sauce here. This LGTM

